PR #1, opened 2022-09-29
PIP-209: Removed Python client wrapper

merlimat:
Removing Python client client which has been moved to https://github.com/apache/pulsar-client-python


BewareMyPower (approved):

BewareMyPower:
I see there is no PR in https://github.com/apache/pulsar-client-python to use the existing C++ code. IMO, we should use pulsar-client-cpp as the submodule in pulsar-client-python.

Are you working in progress? If not, I can help with this issue.

merlimat:
@BewareMyPower I have this: https://github.com/apache/pulsar-client-python/pull/1 

It's still going to be dependent on the 1st release of C++ lib, since it uses unreleased C++ APIs.

PR #11, opened 2022-09-29
Fixed RPM / Deb / APK packages and enabled CI build

merlimat:
### Motivation

Fixed RPM / Deb / APK packages and enabled CI build

### DIFF
--- a/pkg/deb/build-deb.sh
+++ b/pkg/deb/build-deb.sh
@@ -20,28 +20,28 @@
 
 set -e -x
 
-cd /pulsar
+cd /pulsar-client-cpp
 SRC_ROOT_DIR=$(pwd)
-cd $SRC_ROOT_DIR/pulsar-client-cpp/pkg/deb
+cd pkg/deb
 
-POM_VERSION=`$SRC_ROOT_DIR/src/get-project-version.py`
+POM_VERSION=`cat $SRC_ROOT_DIR/version.txt | xargs`
 # Sanitize VERSION by removing `SNAPSHOT` if any since it's not legal in DEB
 VERSION=`echo $POM_VERSION | awk -F-  '{print $1}'`
 
-ROOT_DIR=apache-pulsar-$POM_VERSION-src
-CPP_DIR=$ROOT_DIR/pulsar-client-cpp
+ROOT_DIR=apache-pulsar-client-cpp-$POM_VERSION
+CPP_DIR=$ROOT_DIR
 
 rm -rf BUILD
 mkdir BUILD
 cd BUILD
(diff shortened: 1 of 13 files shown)


pkg/deb/build-deb.sh
```diff
@@ -77,13 +77,8 @@ ls $CPP_DIR/lib/libpulsar*
...
 cp $CPP_DIR/lib/libpulsarwithdeps.a $DEVEL_DEST_DIR/usr/lib
-cp $CPP_DIR/lib/libpulsar.so.$POM_VERSION $DEST_DIR/usr/lib
-cp $CPP_DIR/lib/libpulsarnossl.so.$POM_VERSION $DEST_DIR/usr/lib
-
-pushd $DEST_DIR/usr/lib
-ln -s libpulsar.so.$POM_VERSION libpulsar.so
-ln -s libpulsarnossl.so.$POM_VERSION libpulsarnossl.so
-popd
+cp $CPP_DIR/lib/libpulsar.so $DEST_DIR/usr/lib
+cp $CPP_DIR/lib/libpulsarnossl.so $DEST_DIR/usr/lib
```
BewareMyPower, line 81: Do we need a version suffix?

pkg/deb/docker-build-deb.sh
```diff
@@ -20,15 +20,10 @@
...
-IMAGE_NAME=apachepulsar/pulsar-build:debian-9-2.11
+ROOT_DIR=$(git rev-parse --show-toplevel)
 
-if [[ -z $BUILD_IMAGE ]]; then
-    # pull the image from DockerHub by default
-    docker pull $IMAGE_NAME
-else
-    docker build --platform linux/amd64 -t $IMAGE_NAME $ROOT_DIR/pulsar-client-cpp/pkg/deb
-fi
+IMAGE_NAME=apachepulsar/pulsar-build:debian-9-2.11-x86_64
```
BewareMyPower, line 25: Back to the issue that has been discussed before, what's the process if I want to update the base image? How should I name the image?

merlimat: Ideally, we should use unique image tags. 

At this point the process would be to update the Dockerfile, build the image locally and push it to Dockerhub with the same tag. All the builds after that will sync and fetch the updated image.

pkg/deb/build-deb.sh
merlimat: We don't really need them. It would be anyway unsupported to have different versions of the same libpulsar in the system.

pkg/deb/docker-build-deb.sh
BewareMyPower: Got it. We should include this process when we added documents for the release process.

BewareMyPower (approved):

PR #23, opened 2022-10-07
[feat] Support acknowledging a list of messages

shibd:
### Motivation
https://github.com/apache/pulsar-client-cpp/issues/90

### Modifications
- Add acknowledge message id list API.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### Matching PR in forked repository

https://github.com/shibd/pulsar-client-cpp/pull/3

### DIFF
--- a/lib/MultiTopicsConsumerImpl.cc
+++ b/lib/MultiTopicsConsumerImpl.cc
@@ -647,6 +647,41 @@ void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageId& msgId, ResultCal
     }
 }
 
+void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageIdList& messageIdList, ResultCallback callback) {
+    if (state_ != Ready) {
+        callback(ResultAlreadyClosed);
+        return;
+    }
+
+    std::unordered_map<std::string, MessageIdList> topicToMessageId;
+    for (const MessageId& messageId : messageIdList) {
+        auto topicName = messageId.getTopicName();
+        topicToMessageId[topicName].emplace_back(messageId);
+    }
+
+    auto needCallBack = std::make_shared<std::atomic<int>>(topicToMessageId.size());
+    Result res = ResultOk;
+    auto cb = [callback, needCallBack, &res](Result result) {
+        if (result != ResultOk) {
+            res = result;
+        }
+        needCallBack->fetch_sub(1);
+        if (needCallBack->load() == 0) {
+            callback(res);
+        }
+    };
+    for (const auto& kv : topicToMessageId) {
+        auto optConsumer = consumers_.find(kv.first);
+        if (optConsumer.is_present()) {
+            unAckedMessageTrackerPtr_->remove(kv.second);
+            optConsumer.value()->acknowledgeAsync(kv.second, cb);
+        } else {
+            LOG_ERROR("Message of topic: " << kv.first << " not in unAckedMessageTracker");
+            callback(ResultUnknownError);
+        }
+    }
+}
+
 void MultiTopicsConsumerImpl::acknowledgeCumulativeAsync(const MessageId& msgId, ResultCallback callback) {
     callback(ResultOperationNotSupported);
 }
--- a/lib/MultiTopicsConsumerImpl.cc
+++ b/lib/MultiTopicsConsumerImpl.cc
@@ -660,14 +660,15 @@ void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageIdList& messageIdLis
     }
 
     auto needCallBack = std::make_shared<std::atomic<int>>(topicToMessageId.size());
-    Result res = ResultOk;
-    auto cb = [callback, needCallBack, &res](Result result) {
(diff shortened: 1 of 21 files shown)


shibd:
@BewareMyPower @RobertIndie @Demogorgon314  PTAL.

lib/MultiTopicsConsumerImpl.cc
```diff
@@ -647,6 +647,41 @@ void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageId& msgId, ResultCal
...
+    auto needCallBack = std::make_shared<std::atomic<int>>(topicToMessageId.size());
+    Result res = ResultOk;
+    auto cb = [callback, needCallBack, &res](Result result) {
+        if (result != ResultOk) {
+            res = result;
+        }
+        needCallBack->fetch_sub(1);
+        if (needCallBack->load() == 0) {
+            callback(res);
+        }
```
BewareMyPower, line 671: The combination of two method calls on an atomic variable is not atomic. You can simply write `if (--(*needCallBack) == 0)`.

BewareMyPower (changes requested):

lib/MultiTopicsConsumerImpl.cc
BewareMyPower, line 671: `res` should be defined inside the lambda, not outside. Because it's only used in the lambda.

BTW, please keep it in mind that never capture a reference of a local variable. 
1. The lifetime of the local variable will end after going out of the scope, while the lambda could be invoked in another thread. 
2. Since multiple threads can access the shared local variable, it must be atomic.

That's also why `std::shared_ptr<std::atomic<T>>` is used in other places.
1. Capturing a `std::shared_ptr<U>`  will increase the reference count and extend the lifetime of the `U` instance until all lambdas finished.
2. `U` is `std::atomic<T>`, which is thread safe to access among different threads.

BewareMyPower (changes requested):

lib/MultiTopicsConsumerImpl.cc
BewareMyPower: You can test this example:

```c++
#include <chrono>
#include <iostream>
#include <thread>
using namespace std;

void f() {
  int x = 0;
  int a[1];  // do nothing
  std::thread t([&x] {
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
    cout << std::this_thread::get_id() << " " << &x << " " << x << endl;
  });
  t.detach();
  cout << std::this_thread::get_id() << " " << &x << " " << x << endl;
}

int main() {
  f();
  std::this_thread::sleep_for(std::chrono::milliseconds(100));
}
```

```bash
$ g++ 1.cc -std=c++11 -pthread
$ ./a.out
140118122346304 0x7ffffa28a90c 0
140118122342144 0x7ffffa28a90c 32623
$ ./a.out
139920470566720 0x7fffd23f916c 0
139920470562560 0x7fffd23f916c 32767
```

lib/MultiTopicsConsumerImpl.cc
```diff
@@ -647,6 +647,41 @@ void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageId& msgId, ResultCal
...
+    };
+    for (const auto& kv : topicToMessageId) {
+        auto optConsumer = consumers_.find(kv.first);
+        if (optConsumer.is_present()) {
+            unAckedMessageTrackerPtr_->remove(kv.second);
+            optConsumer.value()->acknowledgeAsync(kv.second, cb);
+        } else {
+            LOG_ERROR("Message of topic: " << kv.first << " not in unAckedMessageTracker");
```
BewareMyPower, line 679: This log is misleading. `consumers_` is not `UnAckedMessageTracker`.

BewareMyPower:
Overall LGTM except the comments I've mentioned.

lib/MultiTopicsConsumerImpl.cc
shibd: Thank your explanation, I remove `res`. PTAL.

lib/MultiTopicsConsumerImpl.cc
```diff
@@ -647,6 +647,42 @@ void MultiTopicsConsumerImpl::acknowledgeAsync(const MessageId& msgId, ResultCal
...
+        topicToMessageId[topicName].emplace_back(messageId);
+    }
+
+    auto needCallBack = std::make_shared<std::atomic<int>>(topicToMessageId.size());
+    auto cb = [callback, needCallBack](Result result) {
+        if (result != ResultOk) {
+            LOG_ERROR("Filed when acknowledge list: " << result);
+            callback(result);
+            // set needCallBack is -1 to avoid repeated callback.
+            needCallBack->store(-1);
```
BewareMyPower, line 668: It's better to call `needCallBack->store(-1)` first  because `callback` is passed from the application side and it could took some time to return. And should we add a `return` here?

BewareMyPower (approved):

PR #31, opened 2022-10-09
Link to OpenSSL statically when LINK_STATIC is ON

BewareMyPower:
### Motivation

See discussions here: https://github.com/apache/pulsar-client-cpp/pull/28#issuecomment-1272588627

The original purpose to not include static OpenSSL library in `libpulsarwithnossl.so` and `libpulsarwithdeps.a` is https://github.com/apache/pulsar/pull/6458. However, the ABI compatibility of OpenSSL is not good. If the Pulsar C++ library links dynamically to OpenSSL and the user only changes the OpenSSL dynamic library, some symbols might not be found.

### Modifications

Use `LINK_STATIC` option to determine whether to link to OpenSSL library statically. After that, there are only 3 libraries generated when `LINK_STATIC` is ON.
- `libpulsar.so`: the dynamic pulsar-client-cpp library.
- `libpulsar.a`: the static pulsar-client-cpp library. When it's used, users must link to all 3rd party dependencies (OpenSSL, curl, etc.)
- `libpulsarwithdeps.a`: the static pulsar-client-cpp library.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,45 +122,28 @@ endif(NOT LOG_CATEGORY_NAME)
 
 add_definitions(-DLOG_CATEGORY_NAME=${LOG_CATEGORY_NAME} -DBUILDING_PULSAR -DBOOST_ALL_NO_LIB -DBOOST_ALLOW_DEPRECATED_HEADERS)
 
-set(OPENSSL_ROOT_DIR ${OPENSSL_ROOT_DIR} /usr/lib64/)
-
-### This part is to find and keep SSL dynamic libs in RECORD_OPENSSL_SSL_LIBRARY and RECORD_OPENSSL_CRYPTO_LIBRARY
-### After find the libs, will unset related cache, and will not affect another same call to find_package.
-if (APPLE)
-    set(OPENSSL_INCLUDE_DIR /usr/local/opt/openssl/include/ /opt/homebrew/opt/openssl/include)
-    set(OPENSSL_ROOT_DIR ${OPENSSL_ROOT_DIR} /usr/local/opt/openssl/ /opt/homebrew/opt/openssl)
-endif ()
-
-set(OPENSSL_USE_STATIC_LIBS FALSE)
-find_package(OpenSSL REQUIRED)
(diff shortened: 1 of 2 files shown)


BewareMyPower:
After this PR, here are examples about how do I build an application with the given three libraries on macOS. Assume they are built with `LINK_STATIC=ON` option. It's a little tricky to build the static library of libcurl, see my comments here: https://github.com/apache/pulsar/issues/4967#issuecomment-1272440827

Given such an application code (`example.cc`):

```c++
#include <pulsar/Client.h>
using namespace pulsar;

int main() {
    client.close();
}
```

Assuming the headers are under `include/` directory and the libraries are under `lib/` directory, which is not a system path.

### libpulsar.so

Build:

```bash
clang++ example.cc -I include/ -L lib/ -lpulsar
```

Run:

```bash
# You have to add the path of `libpulsar.dylib` to a specific env variable so that symbols can be found
DYLD_LIBRARY_PATH=$PWD/lib ./a.out
```

### libpulsar.a

Build:

```bash
# You have to link all 3rd party dependencies.
# BTW, the directory followed by `-L` might be different in your actual env.
clang++ example.cc -I include/ 
lib/libpulsar.a \
-lcurl -lz -lzstd \
-L /usr/local/opt/protobuf/lib/ -lprotobuf \
-L /usr/local/opt/openssl@3/lib/ -lssl -lcrypto \
-L /usr/local/opt/snappy/lib/ -lsnappy
```

Run:

```bash
./a.out
```

### libpulsarwithdeps.a

Build:

```bash
# NOTE: these -framework options are required by curl.a and only needed on macOS 
clang++ example.cc lib/libpulsarwithdeps.a -framework SystemConfiguration -framework CoreFoundation
```

Run:

```bash
./a.out
```

merlimat (approved):

BewareMyPower:
I will fix the rpm / deb scripts soon.

BewareMyPower:
image

I see the "Squash and merge" button is not disabled even if some tests failed. Could you help make some workflows REQUIRED? @merlimat

PR #37, opened 2022-10-11
Added option to set the method name for Basic authentication

merlimat:
### Motivation

Added option to set the method name for Basic authentication

### DIFF
--- a/lib/auth/AuthBasic.cc
+++ b/lib/auth/AuthBasic.cc
@@ -41,6 +41,14 @@ std::string base64_encode(const std::string& s) {
 AuthDataBasic::AuthDataBasic(const std::string& username, const std::string& password) {
     commandAuthToken_ = username + ":" + password;
     httpAuthToken_ = base64_encode(commandAuthToken_);
+    methodName_ = DEFAULT_BASIC_METHOD_NAME;
+}
+
+AuthDataBasic::AuthDataBasic(const std::string& username, const std::string& password,
+                             const std::string& methodName) {
+    commandAuthToken_ = username + ":" + password;
+    httpAuthToken_ = base64_encode(commandAuthToken_);
+    methodName_ = methodName;
 }
 
 AuthDataBasic::~AuthDataBasic() {}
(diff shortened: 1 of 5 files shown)


lib/auth/AuthBasic.cc
```diff
@@ -96,11 +113,17 @@ AuthenticationPtr AuthBasic::create(ParamMap& params) {
     if (passwordIt == params.end()) {
         throw std::runtime_error("No password provided for basic provider");
     }
-
-    return create(usernameIt->second, passwordIt->second);
+    auto methodIt = params.find("method");
```
BewareMyPower, line 116: Could you add a test to cover this part that a param string is accepted? Like https://github.com/apache/pulsar-client-cpp/blob/829618d686c11217922c70b4e5ffba97559c7056/tests/AuthBasicTest.cc#L125

merlimat: 👍  done

lib/auth/AuthBasic.cc
```diff
@@ -41,6 +41,14 @@ std::string base64_encode(const std::string& s) {
 AuthDataBasic::AuthDataBasic(const std::string& username, const std::string& password) {
     commandAuthToken_ = username + ":" + password;
     httpAuthToken_ = base64_encode(commandAuthToken_);
+    methodName_ = DEFAULT_BASIC_METHOD_NAME;
+}
```
BewareMyPower, line 45:
```suggestion
AuthDataBasic::AuthDataBasic(const std::string& username, const std::string& password)
    : AuthDataBasic(username, password, DEFAULT_BASIC_METHOD_NAME) {}
```

Reuse the other constructor.

BewareMyPower (approved):
Overall LGTM. Just left a minor comment.

nodece:
Do we need to extend the `c_Authentication.cc`?

merlimat:
> Do we need to extend the c_Authentication.cc?

Yes, we can add it later. Just want to get this included in the release and python too.

PR #38, opened 2022-10-11
Build on Intel-based macOS with statically linked dependencies

BewareMyPower:
### Motivation

When we build Python or Node.js clients based on the C++ client, it's better to link dependencies statically so that they don't need to link 3rd party dependencies.

There is an existing script to build Python wheel on macOS:

https://github.com/apache/pulsar-client-python/blob/main/build-mac-wheels.sh

However, it's still based on the previous Pulsar main repo. There is no script to build the statically linked C++ client currently.

### Modifications

Add `mac-intel-build.sh` to build C++ client with statically linked dependencies. It should be noted that this script only applies the Intel-based macOS currently.

Then verify this script on a macOS 11 GitHub runner, whose architecture is also x86.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pr-validation.yaml
+++ b/.github/workflows/ci-pr-validation.yaml
@@ -275,7 +275,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build Pulsar C++ client with static linking
-        run: ./mac-intel-build.sh
+        run: ./mac-intel-install.sh
 
   # Job that will be required to complete and depends on all the other jobs
   check-completion:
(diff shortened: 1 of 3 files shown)


BewareMyPower:
@merlimat Could you take a look? Since I don't have a ARM based mac now, I cannot verify the universal build. But I believe it could be simply done based on this patch.

Just change

```bash
export CFLAGS="-fPIC -arch x86_64 -mmacosx-version-min=10.15"
```

to

```bash
export CFLAGS="-fPIC -arch x86_64 -arch arm64 -mmacosx-version-min=10.15"
```

for dependencies other than OpenSSL. For OpenSSL, build another library with `darwin64-arm64-cc` and use `lipo` to create a universal binary.

BewareMyPower:
Close this PR since we cannot use `libpulsarwithdeps.a` to build a universal binary, which is recommended to use on macOS. See https://github.com/apache/pulsar-client-python/pull/11#discussion_r993427825

PR #39, opened 2022-10-11
Use Alpine 3.12 to build APKs

merlimat:
### Motivation

Use Alpine 3.12 to build APKs instead of 3.16 as it could link with symbols that are not available on manylinux images


addisonj (approved):
LGTM

PR #42, opened 2022-10-12
Send notifications to dev@ mailing list

michaeljmarshall:
### Motivation

In our other client repos, we do not have a notification section, and that (apparently) triggers notifications to go to the dev list. I find this particularly helpful when keeping up with our non apache/pulsar repos. If we're concerned about the volume, we can always turn it down later.

https://github.com/apache/pulsar-client-go/blob/master/.asf.yaml

### Modifications

* Remove the `notifications` section of the `.asf.yaml` so that all notifications go to dev@ pulsar

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -63,11 +63,3 @@ github:
 
     # The following branch protections only ensure that force pushes are not allowed
     # branch-3.xxx: {}
-
-
-notifications:
-  commits:      commits@pulsar.apache.org
-  issues:       commits@pulsar.apache.org
-  pullrequests: commits@pulsar.apache.org
-  discussions:  dev@pulsar.apache.org
-  jira_options: link label


merlimat:
This was done on purpose to avoid double/triple notifications since you can get these notifications by "watching" the repo.

michaeljmarshall:
In that case, would it make sense to update the other repos too? I had assumed we had a rule, albeit an implicit one, to send all non apache/pulsar notifications to the dev list. The double notifications are definitely annoying, which is why I don't watch any of our other repos.

merlimat:
> In that case, would it make sense to update the other repos too? 

Yes, I think that was the intention across all the repos.. although several were not actually updated.

>  I had assumed we had a rule, albeit an implicit one, to send all non apache/pulsar notifications to the dev list

I think it was just removed from the main repo (earlier through asking INFRA to do it and later with the yaml config) just because of the sheer volume. 

In general, I don't think people who are on the dev@ list and are not interested in c++/python/go/.. clients should be spammed with all the commits/issues :) 
With the "watch" they have the option to control the notifications, if we put on dev@, there would be no options to avoid it.

michaeljmarshall:
I agree. I opened this PR to make sure we could get to a solution quickly. I'll open up PRs in the missing repos and send a note to the mailing list. Thanks for your reply!

PR #45, opened 2022-10-12
Stick clang-format version to 11 and add back docker-format.sh

BewareMyPower:
### Motivation

Currently the code is formatted by `clang-format` 11. If it's formatted
by other versions of `clang-format`, the style might be a little
different.

### Modifications

- Add back `docker-format.sh` and the associated `Dockerfile.format`.
  This script build the image with `python3` and `clang-format-11`
  installed and the entrypoint is formatting the code.
- Explain the `clang-format` version in README.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`


shibd (approved):

merlimat:
When installing clang-format on current macos & ubuntu, the default is now 11.0. We don't need to worry about the `docker-format.sh` which is already removed. I think we should stick with that and we should set the format in the `CMakeFiles.txt` too.

BewareMyPower:
@merlimat Yeah, removing these scripts looks good to me too.

> I think we should stick with that and we should set the format in the CMakeFiles.txt too.

So this PR would aim at sticking the clang-format to 11.0.0.

BewareMyPower:
I have updated this PR, PTAL again @merlimat

RobertIndie (commented):
> Yeah, removing these scripts looks good to me too.

According to the discussion, seems we will not worry about the `docker-format.sh`. But this file still be added in this PR. Is there some context I have missed?

BewareMyPower:
@RobertIndie The previous `docker-build.sh` and `docker-test.sh` scripts were removed because now we build and run tests on a Ubuntu based GitHub runner. `docker-format.sh` is only used to format code. However, it also uses a manually maintained pre-built Docker image that contains all dependencies, like other `docker-xxx` scripts. It's too heavy.

Though this PR adds `docker-format.sh` back, it's just convenient for non-Ubuntu users. It's lightweight because it only installs a `clang-format-11` based on a official `ubuntu:20.04` image and doesn't run `cmake`.

README.md
```diff
@@ -222,6 +222,8 @@ cd tests
 
 ## Requirements for Contributors
 
-It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` 6.0+ to format files.  `make format` automatically formats the files.
+It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` **11** to format files.  `make format` automatically formats the files.
+
+For Ubuntu users, you can install `clang-format-11` via `apt install clang-format-11`. For other users, run `./docker-format.sh` if you have Docker installed.
```
BewareMyPower, line 227: @RobertIndie See here. Even if I am a Ubuntu user now, using `./docker-format.sh` could be faster now because I don't need to run `cmake`.

Dockerfile.format
```diff
@@ -0,0 +1,35 @@
+#
```
merlimat, line 1: Maybe we could move this into `build-support`. That would also help to not recreate the image each time any cpp file is changed.

BewareMyPower: Moved. But I don't get why will the image be recreated each time any cpp file is changed.

merlimat (approved):

RobertIndie (approved):

README.md
```diff
@@ -222,6 +222,8 @@ cd tests
 
 ## Requirements for Contributors
 
-It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` 6.0+ to format files.  `make format` automatically formats the files.
+It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` **11** to format files.  `make format` automatically formats the files.
+
+For Ubuntu users, you can install `clang-format-11` via `apt install clang-format-11`. For other users, run `./build-support/docker-format.sh` if you have Docker installed.
 
 We welcome contributions from the open source community, kindly make sure your changes are backward compatible with GCC 4.8 and Boost 1.53.
```
momo-jun, line 229: Do we have requirements on the version number of Boost? If yes, it needs to be shown in the [Requirements](https://github.com/apache/pulsar-client-cpp#requirements) section.

README.md
```diff
@@ -222,6 +222,8 @@ cd tests
 
 ## Requirements for Contributors
 
-It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` 6.0+ to format files.  `make format` automatically formats the files.
+It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ client use `clang-format` **11** to format files.  `make format` automatically formats the files.
```
momo-jun, line 225:
```suggestion
It's required to install [LLVM](https://llvm.org/builds/) for `clang-tidy` and `clang-format`. Pulsar C++ clients use `clang-format` **11** to format files.  `make format` automatically formats the files.
```

BewareMyPower: Thanks for your suggestions, I will address them later after since the CI is very unstable now.

README.md
BewareMyPower: The dependency compatibility is hard to test. line 229 is added in a very early time.

PR #54, opened 2022-10-18
[fix] Fix memory leak caused by incorrect close and destruction 

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/55

### Motivation

1. When a producer or consumer is closed, the reference is still stored
   in `ClientImpl`. If a client kept creating producers or consumers,
   the memory usage would not reduce.
2. When the `HandlerBase::connection_` field is modified, the
   `removeProducer` or `removeConsumer` method is not called. Then these
   producers and consumers will be cached in the connection until the
   connection is closed.
3. The `PartitionedProducerImpl` and `MultiTopicsConsumerImpl` have
   cyclic references, when a `Producer` or `Consumer` instance goes out
   of the scope, the destructors are not called. When I used GDB to
   debug them, I found the reference counts were both greater than 1.

### Modifications

Let's use "handlers" to represent "producers and consumers".

1. In `ClientImpl`, use `SynchronizedHashMap` to store references of
   handlers, as well as the `cleanupXXX` methods to remove a handler.
2. Add `HandlerBase::beforeConnectionChange` method, which is called
   before `connection_` is modified. Disallow the access to
  `connection_` from derived classes.
3. Avoid `shared_from_this()` being passed into callbacks in ASIO
   executors for `PartitionedProducerImpl` and
   `MultiTopicsConsumerImpl`.

This PR also unifies the `shutdown` implementations for handlers and
call `shutdown` in the destructors.
1. Cancel the timers
2. Unregister itself from `ClientImpl` and `ClientConnection`
3. Set the create future with `ResultAlreadyClosed`
4. Set the state to `Closed`

It's called when:
- the destructor is called
- `closeAsync` is completed
- `unsubscribeAsync` is completed with ResultOk

### Verifications

`ShutdownTest` is added to verify the following cases:
- a single topic
- a partitioned topic (multiple topics)
- a partitioned topic with regex subscription

`testClose` verifies `shutdown` when `closeAsync` and `unsubscribeAsync`
are called. `testDestructor` verifies `shutdown` when handlers go out of
the scope and the destructors are called.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


BewareMyPower:
Mark it as draft for now because I found a serious problem and will fix it soon.

BewareMyPower:
Now all tests passed, PTAL @merlimat @Demogorgon314 @RobertIndie @shibd

lib/ConsumerImpl.cc
```diff
@@ -990,20 +1002,25 @@ void ConsumerImpl::negativeAcknowledge(const MessageId& messageId) {
...
+    resetCnx();
     scheduleReconnection(get_shared_this_ptr());
 }
 
-void ConsumerImpl::closeAsync(ResultCallback callback) {
-    // Keep a reference to ensure object is kept alive
-    ConsumerImplPtr ptr = get_shared_this_ptr();
+void ConsumerImpl::closeAsync(ResultCallback originalCallback) {
+    auto callback = [this, originalCallback](Result result) {
+        shutdown();
```
shibd, line 1011: When closing the consumer, we also need to remove the client connection. [cnx->removeConsumer(consumerId_);](https://github.com/apache/pulsar-client-cpp/blob/4a864f20530e230ecee7e4ab09430bd65fb9789c/lib/ConsumerImpl.cc#L1057-L1059)

Also, when invoking `unsubscribe`, don't need to remove the `Connection`, right?

shibd: I want to determine whether the processing behavior of unsubscribe and close is consistent.

BewareMyPower: > I want to determine whether the processing behavior of unsubscribe and close is consistent.

Good question. I'm also confused when I read Java client. It's very similar, the only difference is, when you failed to unsubscribe a client, you should keep the state as Ready, i.e. no cleanup. https://github.com/apache/pulsar/blob/cbf5cf58196b05a35b31385c6a42098c598c9b33/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L422

shibd: Yes, I think that after invoking the unsubscribe method, users can still resubscribe. But this is not the case.

BewareMyPower: > When closing the consumer, we also need to remove the client connection

There is an important point that when `HandlerBase::connection_` is modified, we should unregister itself from the connection. Otherwise, the producer or consumer will always be cached in `ClientConnection`. So this PR addes a `beforeConnectionChange` method, which is always called when `setCnx` or `resetCnx` is called.

The `cnx->removeConsumer` is called in `ConsumerImpl::beforeConnectionChange`. Since `shutdown` calls `resetCnx()`, this method is implicitly called in `shutdown`.

Without this design, users have to remove producer or consumer each time the connection changed, like:

```c++
auto cnx = getCnx().lock();
if (cnx) {
    cnx->removeConsumer(consumerId_);
}
connection_ = newConnection;
```

```c++
auto cnx = getCnx().lock();
if (cnx) {
    cnx->removeConsumer(consumerId_);
}
connection_.reset();
```

BewareMyPower: It's the main cause of #55 because the previous code forgot to remove the producer before `connection_.reset()`. When I debugged, I found `cnx` is always empty in the following code block.

https://github.com/apache/pulsar-client-cpp/blob/7f7653b694996f38cf422d8bef24aa1fdfb21c59/lib/ProducerImpl.cc#L720-L723

I didn't look deeper, but it must be caused by the modification of `connection_` field somewhere else.

lib/ConsumerImpl.cc
```diff
@@ -37,15 +37,14 @@ namespace pulsar {
 
 DECLARE_LOG_OBJECT()
 
-ConsumerImpl::ConsumerImpl(const ClientImplPtr client, const std::string& topic,
-                           const std::string& subscriptionName, const ConsumerConfiguration& conf,
-                           bool isPersistent,
+ConsumerImpl::ConsumerImpl(ClientImpl& client, const std::string& topic, const std::string& subscriptionName,
```
merlimat, line 40: Why do we need to change from storing a `weak_ptr` to a reference?

The problem with this is that if you keep using a producer/consumer after the client is already destroyed, you would have a segfault. With the `weak_ptr` we are force to always validating it before using it.

merlimat (commented):
It seems this PR is included more changes than just the part of using a map to keep track of producers/consumers disposal. 

In particular, I'm not sure why we need to change from weak_ptr to references.

lib/ConsumerImpl.cc
BewareMyPower: > The problem with this is that if you keep using a producer/consumer after the client is already destroyed, you would have a segfault.

Actually not. See `TEST_P(ClientCloseTest, testShutdown)`.

> It seems this PR is included more changes than just the part of using a map to keep track of producers/consumers disposal.

And yes. I think it needs another PR for this refactor. For this PR, I will revert these unnecessary changes soon.

merlimat: > Actually not. See TEST_P(ClientCloseTest, testShutdown).

I think it's probably that most method have already state checks at the beginning. 

(Also I'm seeing many places where are currently doing `client_->lock()` and not checking if we got a valid shared_ptr or an empty one...)

> And yes. I think it needs another PR for this refactor. For this PR, I will revert these unnecessary changes soon.

👍 👍

lib/ConsumerImpl.cc
shibd: I see. Thanks for your explanation.

lib/ConsumerImpl.cc
BewareMyPower: > I think it's probably that most method have already state checks at the beginning.

Yes. If the associated `Client` is expired, the producer or consumer should not be able to do anything except a `ResultAlreadyClosed`.

lib/MultiTopicsConsumerImpl.cc
```diff
@@ -886,3 +906,8 @@ void MultiTopicsConsumerImpl::messageProcessed(Message& msg) {
 std::shared_ptr<MultiTopicsConsumerImpl> MultiTopicsConsumerImpl::get_shared_this_ptr() {
     return std::dynamic_pointer_cast<MultiTopicsConsumerImpl>(shared_from_this());
 }
+
+void MultiTopicsConsumerImpl::beforeConnectionChange(ClientConnection& cnx) {
+    consumers_.forEachValue(
+        [&cnx](const ConsumerImplPtr& consumer) { consumer->beforeConnectionChange(cnx); });
```
shibd, line 912: For `MultiTopicConsumerImpl`, It shouldn't be connected, right? 

Should it be more reasonable to change this implementation to throw an exception or add an empty implementation to `ConsumerImplBase`?

BewareMyPower: Good suggestion, adopted.

lib/ConsumerImpl.cc
```diff
@@ -242,10 +242,11 @@ void ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result r
         consumerCreatedPromise_.setValue(get_shared_this_ptr());
     } else {
         if (result == ResultTimeout) {
+            client_.cleanupConsumer(this);
```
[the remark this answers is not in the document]
BewareMyPower: You are right, in my latest code (WIP), it's removed.

tests/ClientTest.cc
```diff
@@ -295,3 +292,90 @@ TEST(ClientTest, testMultiBrokerUrl) {
...
+    Consumer consumer;
+    {
+        Client client(lookupUrl);
+        ASSERT_EQ(ResultOk, client.createProducer(topic_, producer));
+        ASSERT_EQ(ResultOk, subscribe(client, consumer));
+        ASSERT_EQ(PulsarFriend::getProducers(client).size(), 1);
+        ASSERT_EQ(PulsarFriend::getConsumers(client).size(), 1);
+    }  // client is destructed here and `shutdown()` method will be called
+
+    EXPECT_EQ(ResultAlreadyClosed, producer.send(MessageBuilder().setContent("msg").build()));
```
shibd, line 371: It feels like this test is wrong. The user does not call any `close` or `shutdown` methods and has references to `producer` and `consumer`. However, it does not work properly.

Maybe we still need `weak_ptr` to prevent the client from being destructed in this scenario.

BewareMyPower: Let's use another PR to discuss about this topic. I'm currently removing code that is related to `weak_ptr` to pure reference. And will push a new change soon.

lib/ConsumerImpl.cc
RobertIndie: I think the key point of the unsubscribe operation is to delete the subscription in the broker.

lib/PartitionedProducerImpl.h
```diff
@@ -17,16 +17,16 @@
...
 #include <mutex>
 #include <pulsar/MessageRoutingPolicy.h>
 #include <pulsar/TopicMetadata.h>
 #include <lib/TopicName.h>
 
 namespace pulsar {
 
+class ClientImpl;
+
 class PartitionedProducerImpl : public ProducerImplBase,
```
RobertIndie, line 30: Seems we need to implement the `beforeConnectionChange` for the PartitionedProducerImpl.

lib/ConsumerImpl.cc
RobertIndie: I think the key point of the unsubscribe operation is to delete the subscription in the broker.

lib/AckGroupingTrackerDisabled.cc
```diff
@@ -22,6 +22,7 @@
 #include "HandlerBase.h"
 #include "PulsarApi.pb.h"
 #include <pulsar/MessageId.h>
+#include "LogUtils.h"
```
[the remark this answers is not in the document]
BewareMyPower: No. It's just because I changed `std::weak_ptr<ClientImpl>` to `ClientImpl&` in this PR and used forward declaration in the header, so the source file doesn't include "LogUtils.h" (it was included explicitly)

Since these unrelated changes increased the difficulty to review, I'm working on removing them now.

lib/PartitionedProducerImpl.h
BewareMyPower: No. Because it doesn't implement `HandlerBase` and has no `connection_` field. The `ClientConnection` instance doesn't hold any `PartitionedProducerImpl` reference

lib/ConsumerImpl.cc
```diff
@@ -1018,66 +1035,31 @@ void ConsumerImpl::closeAsync(ResultCallback callback) {
...
-        future.addListener(std::bind(&ConsumerImpl::handleClose, get_shared_this_ptr(), std::placeholders::_1,
-                                     callback, ptr));
-    }
-
-    // fail pendingReceive callback
-    failPendingReceiveCallback();
-    failPendingBatchReceiveCallback();
+    auto self = get_shared_this_ptr();
+    cnx->sendRequestWithId(Commands::newCloseConsumer(consumerId_, requestId), requestId)
+        .addListener([this, self, callback](Result result, const ResponseData&) { callback(result); });
```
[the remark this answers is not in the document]
BewareMyPower: Yes

BewareMyPower:
When I add another tests, I found cyclic references for `PartitionedProducerImpl` and `MultiTopicsConsumerImpl` that their destructors won't be called even if `Producer` and `Consumer` go out of scope. I will fix them as well.

lib/ConsumerImpl.cc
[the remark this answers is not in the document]
RobertIndie: Not sure why this caused a compilation error on my computer, but it's fine in the CI.
```
ConsumerImpl.cc:1047:23: error: lambda capture 'this' is not used [-Werror,-Wunused-lambda-capture]
        .addListener([this, self, callback](Result result, const ResponseData&) { callback(result); });

```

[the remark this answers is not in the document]
BewareMyPower: Maybe because your compiler is Clang, whose checking rules is more strict than GCC. (You can see the `-Wunused-lambda-capture` option is added)

lib/ConsumerImpl.cc
BewareMyPower: @RobertIndie Yes, this PR just follows what Java client does.

https://github.com/apache/pulsar/blob/cbf5cf58196b05a35b31385c6a42098c598c9b33/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L413-L422

BewareMyPower:
@merlimat @shibd @RobertIndie Thanks for your reviews. Now I have removed the (`ClientImplWeakPtr` -> `ClientImpl&`) changes and fixed the newly discovered cyclic references (`testDestructor` will fail without the fix), PTAL again.

I changed the milestone to 3.0.0, could you include this PR in the next candidate? @merlimat

shibd (approved):

RobertIndie (approved):

Demogorgon314 (approved):

lib/HandlerBase.h
```diff
@@ -44,11 +44,9 @@ class HandlerBase {
 
     void start();
 
-    /*
-     * get method for derived class to access weak ptr to connection so that they
-     * have to check if they can get a shared_ptr out of it or not
-     */
-    ClientConnectionWeakPtr getCnx() const { return connection_; }
+    ClientConnectionWeakPtr getCnx() const;
```
merlimat, line 47: nit: for later, we could return the `shared_ptr` here, already locked inside the method

merlimat (approved):
👍

lib/HandlerBase.h
BewareMyPower: Good suggestion. I will do it later

Issue #60, opened 2022-10-21
Improvements on building the project


BewareMyPower:
Currently the build time is long and the binary size is large. For example, building the C++ client with the default CMake options in my laptop with 8 threads is:

```bash
$ cmake -B build
$ time cmake --build build -j8
real    4m36.426s
user    31m50.682s
sys     1m24.831s
$ ls -lhS build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/*.o | head -n 10
-rw-r--r-- 1 xyz xyz   25M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/ClientConnection.cc.o
-rw-r--r-- 1 xyz xyz   19M Oct 21 14:08 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/ClientImpl.cc.o
-rw-r--r-- 1 xyz xyz   13M Oct 21 14:08 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/ConsumerImpl.cc.o
-rw-r--r-- 1 xyz xyz   12M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/PatternMultiTopicsConsumerImpl.cc.o
-rw-r--r-- 1 xyz xyz   11M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/MultiTopicsConsumerImpl.cc.o
-rw-r--r-- 1 xyz xyz   11M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/ProducerImpl.cc.o
-rw-r--r-- 1 xyz xyz  7.4M Oct 21 14:08 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/HTTPLookupService.cc.o
-rw-r--r-- 1 xyz xyz  6.6M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/Url.cc.o
-rw-r--r-- 1 xyz xyz  6.5M Oct 21 14:08 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/BinaryProtoLookupService.cc.o
-rw-r--r-- 1 xyz xyz  6.0M Oct 21 14:09 build/lib/CMakeFiles/PULSAR_OBJECT_LIB.dir/PartitionedProducerImpl.cc.o
```

[...]

PR #64, opened 2022-10-23
[refactor] Apply forward declaration as much as possible

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/60

### Motivation

The includes in pulsar-client-cpp is very casual. There are a lot of implicit includes and the forward declaration is not used much. For example, if `lib/ClientConnection.h` was modified, 27 files would be recompiled.

The other problem is the `SortIncludes` attribute in `.clang-format` file is false. It might be okay in early days. However, as the project grows, including many headers without ordering brings a very bad experience. Combining with the very few usages of the forward declaration, it's hard to determine whether a header is still required after a change.

### Modifications

Apply forward declarations as much as possible and change the `SortIncludes` to true in `.clang-format` file.

It's special for the `PulsarApi.pb.h` file because the size of this header is over 1 MiB. For classes we can use forward declaration, but for enumerations we have to include this header.

To solve this problem, `ProtoApiEnums.h` is added to define some constant integers that can be cast implicitly from the enumerations. If we want to use enumerations from `PulsarApi.pb.h`, we can include `ProtoApiEnums.h` instead.

Finally, to unify the include rules, in `lib/*.h`, if we want to include a header (e.g. `xxx.h`) from the same directory, just use `#include "xxx.h"`. Don't use `#include "lib/xxx.h"` or `#include <lib/xxx.h>` in headers of `lib/` directory.

### Improvements

Since forward declaration is applied everywhere now, take `lib/ClientConnection.h` for example. After this patch, only 11 files needs to be recompiled, while 27 files would be recompiled before.

This patch also reduces the binary size and speeds up the compilation time.

Binary size:
- `libpulsar.a`: 319234696 (305 MiB) -> 286716564 (274 MiB)
- `libpulsar.so`: 110162496 (106 MiB) -> 102428456 (98 MiB)

Compilation time with the following commands:

```bash
cmake -B build -DBUILD_TESTS=OFF
cmake --build build -j8
```

From 2m35.640s to 1m44.213s.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


.clang-format
```diff
@@ -19,7 +19,7 @@
 BasedOnStyle: Google
 IndentWidth: 4
 ColumnLimit: 110
-SortIncludes: false
+SortIncludes: true
```
Demogorgon314, line 22: The value should be `CaseInsensitive`  or `CaseSensitive `, see https://clang.llvm.org/docs/ClangFormatStyleOptions.html

BewareMyPower: Currently the `clang-format` binds to 11.0.0, this option is a boolean value, see https://releases.llvm.org/11.0.1/tools/clang/docs/ClangFormatStyleOptions.html

Demogorgon314: Interesting. Looks like the `clang-format version 11.1.0` changed the option to boolean, and the `clang-format version 14.0.6` changed back. 

```
➜   clang-format -version
clang-format version 14.0.6
➜  clang-format -style=llvm -dump-config | grep SortIncludes
SortIncludes:    CaseSensitive

➜  clang-format-11 -version
clang-format version 11.1.0
➜  clang-format-11 -style=llvm -dump-config | grep SortIncludes
SortIncludes:    true
```

BewareMyPower: No. This option is a boolean since the `clang-format` 3.9. It's a newer version of clang that changed it.

BewareMyPower: https://releases.llvm.org/3.9.0/tools/clang/docs/ClangFormatStyleOptions.html

BewareMyPower:
@merlimat @shibd @RobertIndie @Demogorgon314  Please take a look at this PR in prior to other PRs since it brings many changes. Other PRs should be rebased to it.

Though it's a huge PR, most changes are only related to the header includes, which is definitely good for code quality. The code changes are mostly related to the `ProtoApiEnums.h`. Without it, there would still be many files that have to include `PulsarApi.pb.h`.

merlimat (approved):

PR #66, opened 2022-10-24
Add new interfaces to improve data consuming performance. (#65)

### LINKED ISSUE #65: [Improve] Add interfaces in `class Message` to avoid unnecessary memory copy.
## Motivation

Currently `class Message` provides two groups of functions to access the message's payload:

1. `Message::getData` and `Message::getLength`.
2. `getDataAsString`: this is a wrapper of the two above.

The second one performs `std::string`'s construction, which contains expensive memory copy. If `class Message` can provide a new function that gives payload's ownership to the caller like this:

```cpp
// In SharedBuffer.h add new function:
bool SharedBuffer::pop(std::string &target) {
    // Swap *data_ into target, then reset ptr_, readIdx_ etc.
}

// Message.cc
void Message::moveDataIntoString(std::string &data) {
    // Move payload into data via SharedBuffer::pop.
}
```

In high-throughput scenario, these new functions could bring significant performance improvement at data consuming.

tongsucn:
Fixes #65

### Motivation

Improve data consuming performance by adding new interfaces on `class Message` and `class SharedBuffer`.

### Modifications

Adding two functions:

`lib/SharedBuffer.h`

```cpp
void SharedBuffer::pop(std::string &target)
```

`pulsar/Message.h`, `lib/Message.cc`

```cpp
std::string Message::release()
```

### Verifying this change

Run client's unit test, it's added in `MessageTest.cc`.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [x] `doc-complete`

### DIFF
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -101,6 +101,13 @@ class PULSAR_PUBLIC Message {
     KeyValue getKeyValueData() const;
 
     /**
+     * Release and return the message's payload.
+     *
+     * @return the released payload.
+     */
+    std::string releaseData();
+
+    /**
      * Move payload into the given string.
      *
      * @param data the string that receiving the payload's ownership.
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -101,6 +101,13 @@ class PULSAR_PUBLIC Message {
     KeyValue getKeyValueData() const;
 
     /**
+     * Move payload into the given string.
+     *
+     * @param data the string that receiving the payload's ownership.
+     */
+    void moveDataIntoString(std::string& data);
+
+    /**
      * Get the unique message ID associated with this message.
      *
      * The message id can be used to univocally refer to a message without having to keep the entire payload
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -108,13 +108,6 @@ class PULSAR_PUBLIC Message {
     std::string releaseData();
 
     /**
-     * Move payload into the given string.
-     *
-     * @param data the string that receiving the payload's ownership.
-     */
-    void moveDataIntoString(std::string& data);
-
-    /**
      * Get the unique message ID associated with this message.
      *
      * The message id can be used to univocally refer to a message without having to keep the entire payload
--- a/lib/SharedBuffer.h
+++ b/lib/SharedBuffer.h
@@ -25,6 +25,7 @@
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/detail/socket_ops.hpp>
 #include <memory>
+#include <stdexcept>
 #include <string>
 #include <utility>
 
@@ -205,11 +206,12 @@ class SharedBuffer {
         writeIdx_ = 0;
     }
 
-    bool pop(std::string& target) {
-        if (data_ == nullptr) return false;
+    void pop(std::string& target) {
+        if (data_ == nullptr) {
+            throw std::runtime_error("Popped on SharedBuffer object that does NOT own the data.");
+        }
         target.swap(*data_);
         *this = SharedBuffer();
-        return true;
     }
 
    private:
--- a/lib/SharedBuffer.h
+++ b/lib/SharedBuffer.h
@@ -205,6 +205,13 @@ class SharedBuffer {
         writeIdx_ = 0;
     }
 
+    bool pop(std::string& target) {
+        if (data_ == nullptr) return false;
+        target.swap(*data_);
+        *this = SharedBuffer();
+        return true;
+    }
+
    private:
     std::shared_ptr<std::string> data_;
     char* ptr_;
--- a/tests/MessageTest.cc
+++ b/tests/MessageTest.cc
@@ -61,7 +61,7 @@ TEST(MessageTest, testAllocatedContents) {
         throwed = true;
     } catch (...) {
     }
-    ASSERT_FALSE(throwed);
+    ASSERT_TRUE(throwed);
     delete[] content;
 }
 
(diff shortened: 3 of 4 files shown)


BewareMyPower (commented):
How about returning a rvalue reference? e.g.

```c++
std::string&& releaseData();
```

Though the output parameter is widely used in Pulsar C++ client, it's because the returned value is `Result`. In this case, using an output parameter looks ugly IMO.

lib/SharedBuffer.h
```diff
@@ -203,6 +203,13 @@ class SharedBuffer {
         writeIdx_ = 0;
     }
 
+    bool pop(std::string& target) {
+        if (data_ == nullptr) return false;
```
BewareMyPower, line 207: `data_` could never be a `nullptr`

BewareMyPower (changes requested):

lib/SharedBuffer.h
BewareMyPower, line 207: The returned value is always ignored. I think we should throw an exception to disable `moveDataIntoString` on an invalid `Message` object, i.e. built from a `MessageBuilder` whose content was set by `setAllocatedContent` .

BewareMyPower (changes requested):

RobertIndie (changes requested):
Please add some unit tests.

lib/SharedBuffer.h
tongsucn: Good point. Let me add a flag in `SharedBuffer` to denote if it's a read-only one.

tongsucn: Good point.

tongsucn: Done

shibd:
@tongsucn Please rebase the main branch.

tongsucn:
> Please add some unit tests.

Done

include/pulsar/Message.h
```diff
@@ -100,6 +100,20 @@ class PULSAR_PUBLIC Message {
...
+     * @return the released payload.
+     */
+    std::string releaseData();
+
+    /**
+     * Move payload into the given string.
+     *
+     * @param data the string that receiving the payload's ownership.
+     */
+    void moveDataIntoString(std::string& data);
```
shibd, line 115: useless interface.

tongsucn: Done

shibd (approved):

tests/MessageTest.cc
```diff
@@ -53,6 +53,15 @@ TEST(MessageTest, testAllocatedContents) {
     Message msg = msgBuilder.build();
     ASSERT_FALSE(strncmp("content", (char*)msg.getData(), msg.getLength()));
     ASSERT_EQ(content, (char*)msg.getData());
+
+    bool throwed = false;
+    try {
+        msg.releaseData();
+    } catch (std::runtime_error& ex) {
```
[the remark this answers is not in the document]
BewareMyPower: +1

tests/MessageTest.cc
```diff
@@ -53,6 +53,8 @@ TEST(MessageTest, testAllocatedContents) {
     Message msg = msgBuilder.build();
     ASSERT_FALSE(strncmp("content", (char*)msg.getData(), msg.getLength()));
     ASSERT_EQ(content, (char*)msg.getData());
+
+    ASSERT_THROW(msg.releaseData(), std::runtime_error);
```
RobertIndie, line 57: Please add a test case for releasing the data successfully and verify the data is actually swapped.

RobertIndie (changes requested):

tongsucn:
@BewareMyPower @shibd @RobertIndie I found a problem on the new interface:

When batching is enabled on producer side, `Consumer::receive` actually creates a shallow copy of the whole message batch and sets reading offset (i.e. `SharedBuffer::readIdx_`) via `SharedBuffer::slice` in `Commands::deSerializeSingleMessageInBatch`.

It means that the messages from a batch cannot be **_moved_** into another `std::string` without exposing the reading offset. Therefore I think the current implementation of `Message::releaseData` or `Message::releaseData` itself is not a good design. If you have any better ideas, please tell me, or just close the issue and PR.

BewareMyPower:
@tongsucn Good point. I think it's determined by how do you use that.

If you only want a view of the Message's value, you can create an instance `std::string_view` (or a similar class). But you should be careful that the lifetime is the same as `Message`, i.e. you must keep the `Message` objects in memory to make `std::string_view` valid.

If you want some extra in place operations, I think you can expose the non-const `data()` override.

```c++
const void* getData() const;
void* getData();  // new interface
```

If your process function in application just accepts a `std::string`, I'm afraid you have to change it to `std::string_view`.

BewareMyPower:
I just found the `batchReceive` method added in #21. Maybe if the producer enables batching, the consumer must combine `batchReceive` and the `releaseData` APIs. In this case, I prefer not throwing an exception when `data_` is `nullptr`. Instead, fallback to copying the data rather than moving.

tongsucn:
> I just found the `batchReceive` method added in #21. Maybe if the producer enables batching, the consumer must combine `batchReceive` and the `releaseData` APIs. In this case, I prefer not throwing an exception when `data_` is `nullptr`. Instead, fallback to copying the data rather than moving.

Well, I think the result of `batchReceive` may contain data from multiple message batches, which make the situation more complicated. Falling back without telling caller may confuse the callers.

  Under current design, `pulsar::Message` is actually an *agent* of the message data rather than a *container*. An *agent* should NOT give the ownership of the data to others, because it actually **DOES NOT OWN** the data. Therefore, I think interfaces like `releaseData` should be avoided.

PR #71, opened 2022-10-25
[feat] Support expiration for chunked messages

RobertIndie:
### Motivation

Add support for checking expiration for incomplete chunked messages.

### Modifications

* Add configuration `expireTimeOfIncompleteChunkedMessageMs` to the consumer.
* Add timer to check the expiration incomplete chunked messages

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,27 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
     }
 }
 
+void ConsumerImpl::triggerCheckExpiredChunkedTimer() {
+    checkExpiredChunkedTimer_ = executor_->createDeadlineTimer();
```
shibd, line 324: If `checkExpiredChunkedTimer_`  already existed, not need to recreate it.

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,27 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+    checkExpiredChunkedTimer_->expires_from_now(
+        boost::posix_time::milliseconds(expireTimeOfIncompleteChunkedMessageMs_));
+    checkExpiredChunkedTimer_->async_wait([this](const boost::system::error_code& ec) -> void {
```
shibd, line 327: Need to capture `smart pointers` to prevent segmentation faults. Weak pointers are recommended to prevent circular references.
https://github.com/apache/pulsar/pull/17481

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,27 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+        if (ec) {
+            LOG_DEBUG(getName() << " Check expired chunked messages was failed or cancelled, code[" << ec
+                                << "].");
+            return;
+        }
+        Lock lock(chunkProcessMutex_);
+        long currentTimeMs = TimeUtils::currentTimeMillis();
+        chunkedMessageCache_.removeOldestValuesIf(
+            [this, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
```
shibd, line 336: Just capture `expireTimeOfIncompleteChunkedMessageMs_ ` on here.

shibd (changes requested):
Left some small comments.

lib/ConsumerImpl.cc
RobertIndie: Are there any concerns here?

shibd: Oh, You still caught `this` above.

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,31 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
     }
 }
 
+void ConsumerImpl::triggerCheckExpiredChunkedTimer() {
+    checkExpiredChunkedTimer_->expires_from_now(
+        boost::posix_time::milliseconds(expireTimeOfIncompleteChunkedMessageMs_));
+    std::weak_ptr<ConsumerImplBase> weakSelf{shared_from_this()};
+    checkExpiredChunkedTimer_->async_wait([this, weakSelf](const boost::system::error_code& ec) -> void {
```
shibd, line 327: Here still caught `this`, You can refer:

``` c++
    std::weak_ptr<ConsumerImpl> weakSelf{get_shared_this_ptr()};
    checkExpiredChunkedTimer_->async_wait([weakSelf](const boost::system::error_code& ec) -> void {
        auto self = weakSelf.lock();
        if (!self || ec) {
            return;
        }
        Lock lock(self->chunkProcessMutex_);
        long currentTimeMs = TimeUtils::currentTimeMillis();
        self->chunkedMessageCache_.removeOldestValuesIf(
            [self, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
                return currentTimeMs > ctx.getReceivedTimeMs() + self->expireTimeOfIncompleteChunkedMessageMs_;
            });
        self->triggerCheckExpiredChunkedTimer();
        return;
    });

```

shibd (changes requested):

lib/ConsumerImpl.cc
RobertIndie: Are there any concerns here? I still don't understand what the benefit of this approach. I think may be the effect is the same?

lib/ConsumerImpl.cc
RobertIndie: I think it's okay for the current approach.

shibd: Make sense.

lib/ConsumerImpl.cc
shibd: Yes, no problem here. I mainly want to express that this is captured here: https://github.com/apache/pulsar-client-cpp/pull/71#discussion_r1007591353

Now that you've explained it, you can resolve it.

shibd (approved):
Please rebase main branch.

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,31 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+        if (ec) {
+            LOG_DEBUG(getName() << " Check expired chunked messages was failed or cancelled, code[" << ec
+                                << "].");
+            return;
+        }
+        Lock lock(chunkProcessMutex_);
+        long currentTimeMs = TimeUtils::currentTimeMillis();
+        chunkedMessageCache_.removeOldestValuesIf(
+            [this, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
+                return currentTimeMs > ctx.getReceivedTimeMs() + expireTimeOfIncompleteChunkedMessageMs_;
```
Demogorgon314, line 341: Do we need to do ACK for the removed messageIds?

Demogorgon314: I think we should do ACK here. Otherwise, when the client restart, the consumer will receive the removed message again.

RobertIndie: Good point! The user should be able to choose the behavior(ack or just discard it).

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,48 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+        long currentTimeMs = TimeUtils::currentTimeMillis();
+        chunkedMessageCache_.removeOldestValuesIf(
+            [this, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
+                bool expired =
+                    currentTimeMs > ctx.getReceivedTimeMs() + expireTimeOfIncompleteChunkedMessageMs_;
+                if (!expired) {
+                    return false;
+                }
+                for (const MessageId& msgId : ctx.getChunkedMessageIds()) {
+                    if (autoAckOldestChunkedMessageOnQueueFull_) {
```
Demogorgon314, line 347: I just checked the Java client. It always auto-ack when the message expires. Do we need to follow the Java client behavior?

Also, this config is a little confusing here. It should be auto ack expires chunked message, right?

RobertIndie: Oh Thanks. Make sense. If the user set `expireTimeOfIncompleteChunkedMessageMs_ ` >0, then it means the user would not care about the message loss. So we can just ack them.

Demogorgon314 (approved):

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,45 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+        chunkedMessageCache_.removeOldestValuesIf(
+            [this, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
+                bool expired =
+                    currentTimeMs > ctx.getReceivedTimeMs() + expireTimeOfIncompleteChunkedMessageMs_;
+                if (!expired) {
+                    return false;
+                }
+                for (const MessageId& msgId : ctx.getChunkedMessageIds()) {
+                    LOG_INFO("Removing expired chunk messages: uuid: " << uuid << ", messageId: " << msgId);
+                    doAcknowledgeIndividual(msgId, [uuid, msgId](Result result) {
```
BewareMyPower, line 348: Why did you capture `uuid` and `msgId` here?

lib/ConsumerImpl.cc
```diff
@@ -331,6 +371,13 @@ Optional<SharedBuffer> ConsumerImpl::processMessageChunk(const SharedBuffer& pay
                                                  << payload.readableBytes() << " bytes");
 
     Lock lock(chunkProcessMutex_);
+
+    // Lazy task scheduling to expire incomplete chunk message
+    if (!checkExpiredChunkedTimer_) {
+        checkExpiredChunkedTimer_ = executor_->createDeadlineTimer();
+        triggerCheckExpiredChunkedTimer();
+    }
```
BewareMyPower, line 379: Just create the timer in the constructor. `createDeadlineTimer` won't start any timer, it only creates the object. Otherwise, we should always check it's not a nullptr when accessing it.

BewareMyPower (changes requested):

lib/ConsumerImpl.cc
BewareMyPower: Sorry I see it now.

lib/ConsumerImpl.cc
RobertIndie: The `checkExpiredChunkedTimer_ ` should always be not null in `triggerCheckExpiredChunkedTimer `; outside this method, we have always checked the null for it.
If we go this way, we should always get the state of checkExpiredChunkedTimer_ here to check if the timer has been started to implement lazy scheduling. I haven't found a more elegant way to get the state. I think the current approach should be better.

lib/ConsumerImpl.cc
```diff
@@ -319,6 +320,45 @@ void ConsumerImpl::unsubscribeAsync(ResultCallback originalCallback) {
...
+        }
+        if (ec) {
+            LOG_DEBUG(getName() << " Check expired chunked messages was failed or cancelled, code[" << ec
+                                << "].");
+            return;
+        }
+        Lock lock(chunkProcessMutex_);
+        long currentTimeMs = TimeUtils::currentTimeMillis();
+        chunkedMessageCache_.removeOldestValuesIf(
+            [this, &currentTimeMs](const std::string& uuid, const ChunkedMessageCtx& ctx) -> bool {
```
BewareMyPower, line 340: It's better capture `currentTimeMs` by value. Though the lambda is called before `removeOldestValuesIf` finished, capturing a reference (the 8 bytes underlying pointer) is not better than capturing a long (8 bytes).

RobertIndie: +1

Demogorgon314: Question: if we use x86 architectures 32-bit, it will use a 32-bit(4 bytes) pointer here, right?

BewareMyPower: IIRC, long is also 4 bytes in 32-bit arch. BTW, 4 bytes vs 8 bytes is not so much different.

lib/ConsumerImpl.cc
BewareMyPower: We should add a boolean variable `expireChunkMessageTaskScheduled` like Java client does. You can also use something like `Optional` but since `deadline_timer` is not copyable you have to use the move constructor.

BTW, I am going to remove unnecessary `std::shared_ptr` usages in C++ client. It should not be used everywhere. When writing C++, the best practice is to make the lifetime clear like writing Rust.

BewareMyPower: https://godbolt.org/z/eezdxMGnc @Demogorgon314

lib/ConsumerImpl.cc
BewareMyPower: https://godbolt.org/z/eezdxMGnc @Demogorgon314

lib/ConsumerImpl.cc
RobertIndie: You're right. Agree to make the lifetime clearer.

lib/ConsumerImpl.h
```diff
@@ -302,6 +302,7 @@ class ConsumerImpl : public ConsumerImplBase {
 
     const long expireTimeOfIncompleteChunkedMessageMs_;
     DeadlineTimerPtr checkExpiredChunkedTimer_;
+    bool expireChunkMessageTaskScheduled_ = false;
```
BewareMyPower, line 305: You should use an `atomic_bool` here and use CAS to set it true.

I see Java client doesn't use `AtomicBoolean`, but `processMessageChunk` could be called in different threads.

RobertIndie: Good point.

BewareMyPower (approved):

lib/ConsumerImpl.h
```diff
@@ -295,6 +300,12 @@ class ConsumerImpl : public ConsumerImplBase {
     MapCache<std::string, ChunkedMessageCtx> chunkedMessageCache_;
     mutable std::mutex chunkProcessMutex_;
 
+    const long expireTimeOfIncompleteChunkedMessageMs_;
+    DeadlineTimerPtr checkExpiredChunkedTimer_;
+    std::atomic_bool expireChunkMessageTaskScheduled_ = {false};
```
BewareMyPower, line 305: You can simply use the brace initialization for those have no copy constructors.

```suggestion
    std::atomic_bool expireChunkMessageTaskScheduled_{false};
```

PR #72, opened 2022-10-25
Add windows release artifacts

yaalsn:
### Motivation

Add windows release artifacts and create a release with artifacts after pushing tag.

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


BewareMyPower (approved):

merlimat:
@yaalsn @BewareMyPower This has changed from uploading the binaries to GH actions artifacts into a release. It is breaking the release workflow. 
First, we still must upload as artifacts, but also we may not have the credentials to upload to GH releases.

yaalsn:
@merlimat The uploaded artifacts are release artifacts which need by users, and they don't need to build by themself. The token is GITHUB_TOKEN which is a builtin token in GitHub action so that we can use it directly. 

About GITHUB_TOKEN:

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

shibd:
> @merlimat The uploaded artifacts are release artifacts needed by users, and they don't need to build themself. The token is GITHUB_TOKEN which is a builtin token in GitHub action so that we can use it directly.
> 
> About GITHUB_TOKEN:
> 
> https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I think we need to keep uploading the binaries to GH action artifacts. This is required for the current release process. You can refer to the following:

- https://github.com/apache/pulsar-client-cpp/wiki/Committer-Release-Guide#sign-and-stage-the-artifacts
- https://github.com/apache/pulsar-client-cpp/blob/main/build-support/stage-release.sh

yaalsn:
Thanks @shibd, I didn't notice the release guide before.

merlimat:
@yaalsn we could keep the upload to GH release as an optional step, enabled only if some flag is set. For Apache releases, we need to anyway follow the process of staging the binaries.

PR #73, opened 2022-10-26
Support linking static dependencies when building with MSVC

BewareMyPower:
### Motivation

Currently it's impossible to build Pulsar C++ client on Windows with `LINK_STATIC=ON`. It means users have to package all 3rd-party DLLs as well as `pulsar.dll`, which harms the experience.

### Modifications

Enable `LINK_STATIC` when the Vcpkg triplet is `xxx-static`. In this case, find the 3rd party libraries with correct names on Windows. And replace `Threads::Threads` with `CMAKE_THREAD_LIB_INIT`.

The most important change is replacing the `/MD` compile option with `/MT`. It should have been done by setting the
[`MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html) property, but it seems not work. So this PR just modifies the `CMAKE_CXX_FLAGS_<CONFIG>` variables.

For `pulsarWithAllDeps.lib`, add the actual library (`*.lib`) to `COMMON_LIBS` instead of the target name (`dlfcn-win32::dl`).

Some warnings on Windows caused by incorrect compile options are fixed as well.

A workflow is added to verify the static build for x64 and x86 Windows. And a simple example is added as `win-examples` to show the `pulsarWithAllDeps.lib` can be linked without any other dependency to run an executable.

Change the existing release workflow to release two `*.zip` files:
- pulsar-client-cpp-x64-windows-static.zip
- pulsar-client-cpp-x86-windows-static.zip

Each zip file consists of:

```
bin/pulsar.dll         - The dynamic library that links statically to dependencies
include/pulsar/        - Headers
lib/
  pulsar.lib           - The import library of pulsar.dll
  pulsar-static.lib    - The static library
  pulsarWithDeps.lib   - The static library with all dependnecies included
dependencies.txt       - The vcpkg outputs, which contains the dependency versions
```


yaalsn:
Do we need to support LINK_STATIC when releasing windows artifacts?

https://github.com/apache/pulsar-client-cpp/blob/main/.github/workflows/ci-build-binary-artifacts.yaml

shibd:
> Do we need to support LINK_STATIC when releasing windows artifacts?
> 
> https://github.com/apache/pulsar-client-cpp/blob/main/.github/workflows/ci-build-binary-artifacts.yaml

I think need it.

BewareMyPower:
@yaalsn Yes. I think we'd better provide a package whose structure is like:

```
include/pulsar/*.h
lib/pulsar/
  pulsar.dll
  pulsarWithDeps.lib
```

BewareMyPower:
FYI, I'm also working on the Windows packaging CI based on this PR.

https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/3328649258

shibd:
> @yaalsn Yes. I think we'd better provide a package whose structure is like:
> 
> ```
> include/pulsar/*.h
> lib/pulsar/
>   pulsar.dll
>   pulsarWithDeps.lib
> ```

Would you add this to this PR?

BewareMyPower:
@shibd I'm debugging the packaging workflow. There is still a problem. I think both including it in this PR por opening another PR are okay.

shibd:
@BewareMyPower @merlimat  Can this PR go into 3.0? The packaging of the Windows Node js client depends on this.

BewareMyPower:
Mark it as draft for adding support for debug build. Waiting for https://github.com/apache/pulsar-client-cpp/pull/7 being merged first.

BewareMyPower:
Hi @merlimat @yaalsn @Demogorgon314 @RobertIndie , I made some changes just now and this PR is ready to build debug version libraries. So I think #75 can be closed by this PR. Could you take a look at this PR? In addition, this PR also reverts changes of #72, please also take a look.

BewareMyPower:
ping @merlimat @Demogorgon314 @RobertIndie

BewareMyPower:
There is something wrong with the binary upload, I will fix it soon.

.github/workflows/ci-build-binary-artifacts.yaml
```diff
@@ -154,36 +153,45 @@ jobs:
...
-        uses: svenstaro/upload-release-action@v2
+      - name: Upload artifacts (Debug)
+        uses: actions/upload-artifact@v3
         with:
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          file: Windows-${{ matrix.triplet }}.zip
-          asset_name: Windows-${{ matrix.triplet }}.zip
-          tag: ${{ github.ref }}
-          overwrite: true
+          name: ${{ matrix.triplet }}-Debug
```
shibd, line 196: Change `package-linux` job by the way?

shibd (approved):

.github/workflows/ci-build-binary-artifacts.yaml
BewareMyPower: This PR won't modify the unrelated logic

BewareMyPower: Oh yeah, I got it. I will change it back as well.

BewareMyPower:
The example WIndows artifacts built by the new workflow can be found here: https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/3342437635

BewareMyPower:

merlimat (approved):

Issue #51, opened 2022-10-14
Producer#flush doesn't respect non-batched messages


equanz:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

main( https://github.com/apache/pulsar-client-cpp/commit/ba1a7e1c28ab186ed1d51b5cab47334cca1b2aea )

### Minimal reproduce step

1. Create a producer with `batchingEnabled=false`
2. Send some messages asynchronously
3. Call `Producer#flush` or `Producer#flushAsync`
4. (Close a producer)
  - to get AlreadyClosed exception
     https://github.com/apache/pulsar-client-cpp/blob/ba1a7e1c28ab186ed1d51b5cab47334cca1b2aea/lib/ProducerImpl.cc#L665-L666

### What did you expect to see?

When `Producer#flush` or `Producer#flushAsync` is called, wait for the completion of sending messages, even if non-batched messages.

### What did you see instead?

`Producer#flush` and `Producer#flushAsync` don't guarantee the completion of sending non-batched messages.

### Anything else?

Currently, `ProducerImpl#flushAsync` do nothing if `batchMessageContainer_` is not initialized.
https://github.com/apache/pulsar-client-cpp/blob/ba1a7e1c28ab186ed1d51b5cab47334cca1b2aea/lib/ProducerImpl.cc#L331-L333

In Java client, wait last sent messages. We should implement C++ client like it.
https://github.com/apache/pulsar/pull/2103

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

shibd:
I reproduced it and I'll start fixing it.

equanz:
I'll close this issue because https://github.com/apache/pulsar-client-cpp/pull/98 was merged.
Thank you @shibd

PR #98, opened 2022-11-01
[fix] Flush no batch message when call producer.flush

shibd:
### Motivation

#51

### Modifications

- when calling `producer.flush`, wait for no batch message send suceese. 

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/OpSendMsg.h
+++ b/lib/OpSendMsg.h
@@ -39,6 +39,7 @@ struct OpSendMsg {
     boost::posix_time::ptime timeout_;
     uint32_t messagesCount_;
     uint64_t messagesSize_;
+    std::vector<std::function<void(Result)>> trackerCallbacks_;
 
     OpSendMsg() = default;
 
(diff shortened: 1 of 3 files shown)


lib/OpSendMsg.h
```diff
@@ -59,6 +60,13 @@ struct OpSendMsg {
         if (sendCallback_) {
             sendCallback_(result, messageId);
         }
+        for (const auto& trackerCallback : trackerCallbacks_) {
+            trackerCallback(result);
+        }
+    }
+
+    void addTrackerCallback(std::function<void(Result)> trackerCallback) {
```
Demogorgon314, line 68: nit
```suggestion
    void addTrackerCallback(const std::function<void(Result)> trackerCallback) {
```

Demogorgon314 (approved):
LGTM.

lib/OpSendMsg.h
BewareMyPower: It's common to pass the `std::function` by value. Regarding the value argument, adding `const` does not harm but it also does not bring any benefit.

Take https://en.cppreference.com/w/cpp/string/basic_string/basic_string for example, the constructors that accept integral types don't add `const` qualifiers.

```c++
basic_string( size_type count, CharT ch,
              const Allocator& alloc = Allocator() );
```

BewareMyPower (approved):

PR #105, opened 2022-11-07
Add MessageId::batchSize() and the MessageIdBuilder

BewareMyPower:
Master issue: https://github.com/apache/pulsar-client-cpp/issues/87

### Motivation

To support batch index acknowledgment, we must provide a method to get the batch size of a batched message ID.

### Modifications

Instead of adding another overload constructor to `MessageId`, this PR adds a `MessageIdBuilder` class to construct the `MessageId` in a more elegant way.

The original constructor is counterintuitive because the partition index is the 1st argument.

https://github.com/apache/pulsar-client-cpp/blob/74ef1a01f5c7a4604d251de6d040c433f9bbf56b/include/pulsar/MessageId.h#L47

Therefore, this PR marks it as deprecated and replace all invocations of it with the `MessageIdBuilder` usages.

To verify the `MessageId::batchSize()`, the following tests are modified:
- `BatchMessageTest.testBatchSizeInBytes`: the batch size is always 2 because of the `batchingMaxAllowedSizeInBytes` config.
- `MessageChunkingTest.testEndToEnd`: the batch size field is not set
  (default: 0) because batching is disabled.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/MessageBatch.cc
+++ b/lib/MessageBatch.cc
@@ -47,7 +47,7 @@ MessageBatch& MessageBatch::parseFrom(const SharedBuffer& payload, uint32_t batc
     batch_.clear();
 
     for (int i = 0; i < batchSize; ++i) {
-        batch_.push_back(Commands::deSerializeSingleMessageInBatch(batchMessage_, i));
+        batch_.push_back(Commands::deSerializeSingleMessageInBatch(batchMessage_, i, batchSize));
     }
     return *this;
 }
--- a/include/pulsar/MessageId.h
+++ b/include/pulsar/MessageId.h
@@ -37,7 +37,11 @@ class PULSAR_PUBLIC MessageId {
     MessageId();
 
     /**
+     * @deprecated
+     *
      * Construct the MessageId
+     *
+     * NOTE: This API still exists for backward compatibility, use MessageIdBuilder instead.
      *
      * @param partition the partition number of a topic
      * @param ledgerId the ledger id
@@ -88,6 +92,7 @@ class PULSAR_PUBLIC MessageId {
     int64_t entryId() const;
     int32_t batchIndex() const;
     int32_t partition() const;
+    int32_t batchSize() const;
 
    private:
     friend class ConsumerImpl;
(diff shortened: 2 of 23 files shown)


BewareMyPower:

BewareMyPower:
@merlimat @RobertIndie @Demogorgon314 PTAL. Though I opened a discussion [here](https://lists.apache.org/thread/rdkqnkohbmkjjs61hvoqplhhngr0b0sd) and reached a concensus that Java's MessageId is an opaque, the MessageId in C++ is still constructable from application side.

This PR mainly changes the construction from

```c++
// partition: 1, ledger: 2, entry: 3, batch index: 4
MessageId id(1, 2, 3, 4);
```

to

```c++
auto id = MessageIdBuilder().ledgerId(2).entryId(3).partition(1).batchIndex(4).build();
```

And support `batch_size` field.

lib/MessageBatch.cc
```diff
@@ -47,7 +47,7 @@ MessageBatch& MessageBatch::parseFrom(const SharedBuffer& payload, uint32_t batc
     batch_.clear();
 
     for (int i = 0; i < batchSize; ++i) {
-        batch_.push_back(Commands::deSerializeSingleMessageInBatch(batchMessage_, i));
+        batch_.push_back(Commands::deSerializeSingleMessageInBatch(batchMessage_, i, batchSize));
```
shibd, line 50: Have a question: What is the role of `MessageBatch`, Doesn't seem to find a place to use it.

shibd (approved):

lib/MessageBatch.cc
BewareMyPower: https://github.com/apache/pulsar/pull/4992

Though it's not a good design IMO.

shibd: Get it, Thanks.

Demogorgon314 (approved):

RobertIndie (approved):

Issue #107, opened 2022-11-08
[Bug] Message::getDataAsString() will crash with Windows debug library


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

It works for master branch, but you can also use the release here [x64-windows-static-Debug](https://github.com/streamnative/pulsar-client-cpp/suites/9150203394/artifacts/425759462) in [here](https://github.com/streamnative/pulsar-client-cpp/actions/runs/3403392007).

Uncompress it to a directory, e.g. `D:\pulsar-cpp-debug`.

```
D:\pulsar-cpp-debug
    bin/pulsar.dll
    lib/pulsar.lib
    include/
```

### Minimal reproduce step

CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.15)

project("PulsarApp" CXX)

set(PULSAR_ROOT "D:\\pulsar-cpp-debug")
add_executable(PulsarApp main.cc)
include_directories("${PULSAR_ROOT}/include")
target_link_libraries(PulsarApp PRIVATE "${PULSAR_ROOT}/lib/pulsar.lib")
set_property(TARGET PulsarApp PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
```

main.cc

```c++
#include <pulsar/Client.h>
using namespace pulsar;

int main() {
  auto msg = MessageBuilder().setContent("hello").build();
  std::cout << msg.getDataAsString() << std::endl;
  return 0;
}
```

Then run the following commands in PowerShell

```PowerShell
cmake -B build
cmake --build build --config Debug
cp D:\pulsar-cpp-debug\bin\pulsar.dll .
.\build\Debug\PulsarApp.exe
```

### What did you expect to see?

"hello" is printed.

### What did you see instead?

![image](https://user-images.githubusercontent.com/18204803/200484264-c143061c-58af-4322-8546-5886049a3467.png)

### Anything else?

It looks like to be caused by the combination of static linking of 3rd party dependencies (`LINK_STATIC=ON`) and the `std::string` API compatibility.

For example, the following example works well.

```c++
#include <assert.h>
#include <pulsar/Client.h>
using namespace pulsar;

int main() {
  auto topic = "topic";
  Client client("pulsar://172.24.101.226:6650");
  Producer producer;
[...]

PR #108, opened 2022-11-08
Fix MessageId::getDataAsString() crashed with MSVC debug config

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/107

### Motivation

The `MessageId::getDataAsString()` API returns a `std::string` to the application side. In most cases it's not an issue. However, when building Windows DLLs with `LINK_STATIC=ON`, the library will be built with `/MTd` or `/MT` option to link 3rd party dependencies statically. In this case, the DLL and the application have different C runtime libraries that allocate or deallocate memory. The returned `std::string` object is allocated inside the DLL, while it will be destroyed in the application for debug build because RVO is not applied. The destruction could crash because the application C runtime cannot find the heap address from the C runtime in DLL.

### Modifications

For MSVC debug build, change the API to return a const reference to `std::string`. Then the original `std::string` object will be deallocated inside the DLL.

### DIFF
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -90,8 +90,15 @@ class PULSAR_PUBLIC Message {
      * Get string representation of the message
      *
      * @return the string representation of the message payload
+     *
+     * NOTE: For MSVC with debug mode, return a thread local std::string object to avoid memory allocation
+     * across DLLs and applications, which could lead to a crash.
      */
+#if defined(_MSC_VER) && !defined(NDEBUG)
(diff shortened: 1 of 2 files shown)


BewareMyPower:
Though it changes the public API, but it's only for Windows debug artifacts, which should only be used for tests, not in production. For users of Windows debug libraries, returning a const reference won't affect the application code.

BewareMyPower:

codelipenghui (approved):

PR #117, opened 2022-11-11
Fix getLastMessageId method dead recursion.

shibd:
### Motivation

There is a problem with the `getLastMessageId` method currently implemented.

It is recursively invoking itself.
https://github.com/apache/pulsar-client-cpp/blob/ad79becf0814a992ad493bc625ba0a489900451f/lib/Consumer.cc#L293-L301

### Modifications

- Let it call the actual implementation.

**Note**:  `MultiTopicConsumer.getLastMessageId()` is not supported for the time being. I will support it later. Supporting `MultiTopicConsumer` requires the implementation of [MultiMessageId](https://github.com/apache/pulsar/blob/master/pulsar-client/src/main/java/org/apache/pulsar/client/impl/MultiMessageIdImpl.java), which requires more thought and design.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/ConsumerTest.cc
+++ b/tests/ConsumerTest.cc
@@ -841,6 +841,30 @@ TEST(ConsumerTest, testPartitionsWithCloseUnblock) {
     thread.join();
 }
 
+TEST(ConsumerTest, testGetLastMessageId) {
+    Client client(lookupUrl);
+    const std::string topic = "testGetLastMessageId-" + std::to_string(time(nullptr));
+
+    Consumer consumer;
+    ASSERT_EQ(ResultOk, client.subscribe(topic, "test-sub", consumer));
+
+    MessageId msgId;
+    ASSERT_EQ(ResultOk, consumer.getLastMessageId(msgId));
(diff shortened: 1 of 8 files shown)


tests/ConsumerTest.cc
```diff
@@ -841,6 +841,30 @@ TEST(ConsumerTest, testPartitionsWithCloseUnblock) {
...
+
+    Producer producer;
+    ASSERT_EQ(ResultOk, client.createProducer(topic, producer));
+    Message msg = MessageBuilder().setContent("message").build();
+    ASSERT_EQ(ResultOk, producer.send(msg));
+
+    ASSERT_EQ(ResultOk, consumer.getLastMessageId(msgId));
+    ASSERT_TRUE(msgId != MessageId(-1, -1, -1, -1));
+
+    std::cout << msgId << std::endl;
```
BewareMyPower, line 863:
```suggestion
```

Don't use `std::cout` in tests

BewareMyPower (approved):

PR #122, opened 2022-11-23
Add handle the authorization error

nodece:
### Motivation

When receiving the authorization error, we need to close the connection between the broker and the client.

### Modifications

Add check the `ServerError.AuthorizationError`. 

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -1130,6 +1130,7 @@ void ClientConnection::handleIncomingCommand(BaseCommand& incomingCmd) {
                             }
                         }
                     }
+                    checkServerError(error.error());
                     break;
                 }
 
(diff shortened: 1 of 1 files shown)


lib/ClientConnection.cc
```diff
@@ -1709,6 +1710,9 @@ void ClientConnection::checkServerError(ServerError error) {
             // https://github.com/apache/pulsar/pull/274
             closeSocket();
             break;
+        case proto::ServerError::AuthorizationError:
```
shibd, line 1713: Can you explain why do this? Because I see that the Java client has not yet dealt with it.

nodece: When receiving the AuthenticationError or AuthorizationError, I think we should disconnect the connection. 

Right now, we only handle the AuthenticationError in Java client. See https://github.com/apache/pulsar/blob/94a6d3642800ac48b779168cbe17bb505b7c1f96/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L757

shibd: Yes, I see. I mean, the java client also needs to be disconnected when receive AuthenticationError.

PR #124, opened 2022-11-24
[improve] Exclude debug artifact in the release and tar the windows artifacts

RobertIndie:
### Motivation

The debug artifact for the windows system is too large. They are only used for debugging and don't need to include in the release. And it's better to tar the `windows-static` artifact to make it easier to download.

The final result looks like this: https://dist.apache.org/repos/dist/dev/pulsar/pulsar-client-cpp/pulsar-client-cpp-3.1.0-candidate-1/

### Modifications

* Exclude the windows debug artifact in the release
* Tar `windows-static` artifacts in the release.

### Verifying this change

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/build-support/download-release-artifacts.py
+++ b/build-support/download-release-artifacts.py
@@ -45,6 +45,9 @@ with urllib.request.urlopen(request) as response:
     data = json.loads(response.read().decode("utf-8"))
     for artifact in data['artifacts']:
         name = artifact['name']
+        # Skip debug artifact
+        if name.endswith("-Debug"):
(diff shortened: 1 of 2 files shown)


BewareMyPower (commented):
Could you add a release process so that it would be more clear that when would these scripts be used?

BewareMyPower (approved):

PR #125, opened 2022-11-25
[fix] Fix PartitionedProducerImpl::closeAsync to close sub-producers properly

erobot:
### Motivation

PartitionedProducerImpl do not close sub-producers properly when any sub-producer creation fails. Continuing to retry creating producer will eventually reach the maximum producer limit. It seems a regression caused by #54. 

When sub-producer creation fails, state_ is set to Failed. PartitionedProducerImpl::closeAsync only do cleanup when state_==Ready and sub-producers do not close when state_==Failed.

https://github.com/apache/pulsar-client-cpp/blob/f0268ecd29a6d0030b7d07379ec609884b4c14ff/lib/PartitionedProducerImpl.cc#L273-L276

### Modifications

Close sub-producers when state != Closed.

### Verifying this change

test code:

    Producer producer;
    while (true) {
        Result result = client.createProducer("persistent://public/default/test", producer);
        if (result != ResultOk) {
            cerr << "Error creating producer: " << result << endl;
            std::this_thread::sleep_for(std::chrono::seconds(1));
            continue;
        }
        break;
    }

test cmd:

    pulsar-admin namespaces set-max-producers-per-topic -p 10 public/default
    pulsar-admin topics create-partitioned-topic -n 2 public/default/test
    pulsar-perf produce -r 1 -n 10 public/default/test-partition-0

test procedure:
1. set namespace max producer limit to 10
2. create partitoned topic public/default/test with 2 partitions
3. use pulsar-perf to create 10 prodcuers on test-partition-0
4. run test code to create producer for public/default/test, will fail because test-partition-0 already have 10 producer
5. 
    * without fix, test-partition-1 will have 10 producers because of sub-producer not close
    * with fix, test-partition-1 will have not more than 1 producer

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
bug-fix only

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/PartitionedProducerImpl.cc
+++ b/lib/PartitionedProducerImpl.cc
@@ -266,13 +266,10 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
             originalCallback(result);
         }
     };
-    if (state_ == Closed) {
-        closeCallback(ResultAlreadyClosed);
-        return;
-    }
-    State expectedState = Ready;
-    if (!state_.compare_exchange_strong(expectedState, Closing)) {
-        return;
+
+    State previous_state = state_.exchange(Closed);
+    if (previous_state == Closed) {
+      return closeCallback(ResultAlreadyClosed);
     }
 
     cancelTimers();
--- a/lib/PartitionedProducerImpl.cc
+++ b/lib/PartitionedProducerImpl.cc
@@ -269,7 +269,7 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
 
     State previous_state = state_.exchange(Closed);
     if (previous_state == Closed) {
-      return closeCallback(ResultAlreadyClosed);
+        return closeCallback(ResultAlreadyClosed);
     }
 
     cancelTimers();
--- a/lib/PartitionedProducerImpl.cc
+++ b/lib/PartitionedProducerImpl.cc
@@ -267,8 +267,7 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
         }
     };
 
-    State previous_state = state_.exchange(Closed);
-    if (previous_state == Closed) {
+    if (state_.exchange(Closed) == Closed) {
         return closeCallback(ResultAlreadyClosed);
     }
 
--- a/lib/PartitionedProducerImpl.cc
+++ b/lib/PartitionedProducerImpl.cc
@@ -267,8 +267,9 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
         }
     };
 
-    if (state_.exchange(Closed) == Closed) {
-        return closeCallback(ResultAlreadyClosed);
+    if (state_ == Closed || state_.exchange(Closing) == Closing) {
+        closeCallback(ResultAlreadyClosed);
+        return;
     }
 
     cancelTimers();
--- a/tests/ProducerTest.cc
+++ b/tests/ProducerTest.cc
@@ -421,7 +421,7 @@ TEST_P(ProducerTest, testCloseSubProducerWhenFail) {
 
     std::this_thread::sleep_for(std::chrono::seconds(1));
 
-    // create producer for partition-1, should succ
+    // create producer for partition-1, should succeed
     Producer producer;
     ASSERT_EQ(ResultOk, client.createProducer(topicName + "-partition-1", producerConfiguration, producer));
     producers.push_back(producer);
--- a/tests/ProducerTest.cc
+++ b/tests/ProducerTest.cc
@@ -406,7 +406,7 @@ TEST_P(ProducerTest, testCloseSubProducerWhenFail) {
         producers.push_back(producer);
     }
 
-    // create partitoined producer, should fail because partition-0 already reach max producer limit
+    // create partitioned producer, should fail because partition-0 already reach max producer limit
     for (int i = 0; i < maxProducersPerTopic; ++i) {
         auto partitionedProducer = std::make_shared<PartitionedProducerImpl>(
             PulsarFriend::getClientImplPtr(client), TopicName::get(topicName), partitionNum,
--- a/tests/ProducerTest.cc
+++ b/tests/ProducerTest.cc
@@ -408,15 +408,8 @@ TEST_P(ProducerTest, testCloseSubProducerWhenFail) {
 
     // create partitioned producer, should fail because partition-0 already reach max producer limit
     for (int i = 0; i < maxProducersPerTopic; ++i) {
-        auto partitionedProducer = std::make_shared<PartitionedProducerImpl>(
-            PulsarFriend::getClientImplPtr(client), TopicName::get(topicName), partitionNum,
-            producerConfiguration);
-        partitionedProducer->start();
-        Result result;
-        ProducerImplBaseWeakPtr ptr;
-        ASSERT_TRUE(
-            partitionedProducer->getProducerCreatedFuture().get(result, ptr, std::chrono::seconds(1)));
-        ASSERT_EQ(result, ResultProducerBusy);  // reach max producer limit
+        Producer producer;
+        ASSERT_EQ(ResultProducerBusy, client.createProducer(topicName, producer));
     }
 
     std::this_thread::sleep_for(std::chrono::seconds(1));
--- a/tests/ProducerTest.cc
+++ b/tests/ProducerTest.cc
@@ -374,7 +374,7 @@ TEST_P(ProducerTest, testFlushNoBatch) {
     client.close();
 }
 
-TEST_P(ProducerTest, testCloseSubProducerWhenFail) {
+TEST(ProducerTest, testCloseSubProducerWhenFail) {
     Client client(serviceUrl);
 
     std::string ns = "test-close-sub-producer-when-fail";
--- a/tests/ProducerTest.cc
+++ b/tests/ProducerTest.cc
@@ -22,6 +22,7 @@
 #include <thread>
 
 #include "HttpHelper.h"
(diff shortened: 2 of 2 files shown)


shibd:
@erobot Hi, Can you add unit test based on the steps`verifying this change` chapter?

lib/PartitionedProducerImpl.cc
```diff
@@ -266,13 +266,10 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
...
-        return;
-    }
-    State expectedState = Ready;
-    if (!state_.compare_exchange_strong(expectedState, Closing)) {
-        return;
+
+    State previous_state = state_.exchange(Closed);
+    if (previous_state == Closed) {
+        return closeCallback(ResultAlreadyClosed);
     }
```
BewareMyPower, line 273:
```suggestion
    State previousState = state_.exchange(Closed);
    if (previousState == Closed) {
        return closeCallback(ResultAlreadyClosed);
    }
```

Use camel case instead of snake case to make code style consistent. Or just simplify it to `if (state_.exchange(Closed) == Closed)`.

BewareMyPower (changes requested):
Could you also add the unit test to avoid the regression?

BewareMyPower:
> without fix, test-partition-1 will have 10 producers because of sub-producer not close
> with fix, test-partition-1 will have not more than 1 producer

The difference is not clear from the output.

[with-fix.log](https://github.com/apache/pulsar-client-cpp/files/10091081/1.log)

[without-fix.log](https://github.com/apache/pulsar-client-cpp/files/10091083/2.log)

Maybe a unit test would be better to tell the difference.

lib/PartitionedProducerImpl.cc
BewareMyPower: In addition, we should follow the behavior of the Java client to set the state to `Closing`. Only after all sub-producers are closed should we set the state to `Closed`.

```java
        if (getState() == State.Closing || getState() == State.Closed) {
            return CompletableFuture.completedFuture(null);
        }
        setState(State.Closing); // here
```

lib/PartitionedProducerImpl.cc
```diff
@@ -266,13 +266,10 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
...
-        closeCallback(ResultAlreadyClosed);
-        return;
-    }
-    State expectedState = Ready;
-    if (!state_.compare_exchange_strong(expectedState, Closing)) {
-        return;
+
+    State previous_state = state_.exchange(Closed);
+    if (previous_state == Closed) {
+        return closeCallback(ResultAlreadyClosed);
```
RobertIndie, line 272: We should exchange it with the `Closing` state. And if the producer is already in `Closing`, this code still be passed, which results in wrong behavior.

RobertIndie (changes requested):

BewareMyPower:
Just digged deeper into this issue. I think there is no problem with the current master.

If one internal producer failed to create, the partitioned producer would not be added into the `ClientImpl` or `ClientConnection`, see https://github.com/apache/pulsar-client-cpp/blob/96d870527d2a078645c4866e9d75b3f3b97193ba/lib/ClientImpl.cc#L202-204

Then the `PartitionedProducerImpl` will be destroyed and `shutdown()` will be called for the cleanup work (cancel the timer) in the destructor. Regarding the `ProducerImpl` objects, since they also failed to create, there is no need to call `closeAsync` on them because **closing a failed producer is redundant.**

BewareMyPower:
I got it. It's caused when some sub-producers can be created successfully while other could fail.

lib/PartitionedProducerImpl.cc
```diff
@@ -266,13 +266,9 @@ void PartitionedProducerImpl::closeAsync(CloseCallback originalCallback) {
...
-        closeCallback(ResultAlreadyClosed);
-        return;
-    }
-    State expectedState = Ready;
-    if (!state_.compare_exchange_strong(expectedState, Closing)) {
-        return;
+
+    if (state_.exchange(Closed) == Closed) {
+        return closeCallback(ResultAlreadyClosed);
     }
```
BewareMyPower, line 272: You should set the state to `Closing`, see comment [here](https://github.com/apache/pulsar-client-cpp/pull/125#discussion_r1033067388)

tests/ProducerTest.cc
```diff
@@ -373,4 +374,62 @@ TEST_P(ProducerTest, testFlushNoBatch) {
...
+    // create producers for partition-0 up to max producer limit
+    std::vector<Producer> producers;
+    for (int i = 0; i < maxProducersPerTopic; ++i) {
+        Producer producer;
+        ASSERT_EQ(ResultOk,
+                  client.createProducer(topicName + "-partition-0", producerConfiguration, producer));
+        producers.push_back(producer);
+    }
+
+    // create partitoined producer, should fail because partition-0 already reach max producer limit
```
BewareMyPower, line 409:
```suggestion
    // create partitioned producer, should fail because partition-0 already reach max producer limit
```

tests/ProducerTest.cc
```diff
@@ -373,4 +374,62 @@ TEST_P(ProducerTest, testFlushNoBatch) {
...
+        Result result;
+        ProducerImplBaseWeakPtr ptr;
+        ASSERT_TRUE(
+            partitionedProducer->getProducerCreatedFuture().get(result, ptr, std::chrono::seconds(1)));
+        ASSERT_EQ(result, ResultProducerBusy);  // reach max producer limit
+    }
+
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+
+    // create producer for partition-1, should succ
```
BewareMyPower, line 424:
```suggestion
    // create producer for partition-1, should succeed
```

BewareMyPower (commented):
typo

tests/ProducerTest.cc
```diff
@@ -373,4 +374,62 @@ TEST_P(ProducerTest, testFlushNoBatch) {
...
+    for (int i = 0; i < maxProducersPerTopic; ++i) {
+        auto partitionedProducer = std::make_shared<PartitionedProducerImpl>(
+            PulsarFriend::getClientImplPtr(client), TopicName::get(topicName), partitionNum,
+            producerConfiguration);
+        partitionedProducer->start();
+        Result result;
+        ProducerImplBaseWeakPtr ptr;
+        ASSERT_TRUE(
+            partitionedProducer->getProducerCreatedFuture().get(result, ptr, std::chrono::seconds(1)));
+        ASSERT_EQ(result, ResultProducerBusy);  // reach max producer limit
```
BewareMyPower, line 419:
```suggestion
        Producer producer;
        ASSERT_EQ(ResultProducerBusy, client.createProducer(topicName, producer));
```

You'd better not use the internal class directly. After that, `#include "PulsarFriend.h"` can be removed as well.

lib/PartitionedProducerImpl.cc
BewareMyPower: Maybe the following implementation could be better?

```c++
    if (state_ == Closed || state_.exchange(Closing) == Closing) {
        closeCallback(ResultAlreadyClosed);
        return;
    }
```

tests/ProducerTest.cc
erobot: You’re right, I'll fix it. At first I wanted to test the internal prodcuers state, but it's not needed now.

tests/ProducerTest.cc
```diff
@@ -373,4 +374,62 @@ TEST_P(ProducerTest, testFlushNoBatch) {
     client.close();
 }
 
+TEST_P(ProducerTest, testCloseSubProducerWhenFail) {
```
BewareMyPower, line 377:
```suggestion
TEST(ProducerTest, testCloseSubProducerWhenFail) {
```

No need to use `TEST_P` here.

BewareMyPower (changes requested):

lib/PartitionedProducerImpl.cc
erobot: I'll change to your suggested codes as you’re perfered.

Some thoughts：
* c++ client state is more than java client and it's more complicated to maintain.
* Closing and Closed seems not much different in c++ client.

BewareMyPower: > c++ client state is more than java client and it's more complicated to maintain.

No, the fact is opposite. There are 5 states in C++ client:

```c++
    enum State
    {
        Pending,
        Ready,
        Closing,
        Closed,
        Failed
```

while Java client has 9 states (see `HandlerState.java`):

```java
    enum State {
        Uninitialized, // Not initialized
        Connecting, // Client connecting to broker
        Ready, // Handler is being used
        Closing, // Close cmd has been sent to broker
        Closed, // Broker acked the close
        Terminated, // Topic associated with this handler
                    // has been terminated
        Failed, // Handler is failed
        RegisteringSchema, // Handler is registering schema
        ProducerFenced, // The producer has been fenced by the broker
    }
```

The `Pending` state in C++ client is the `Uninitialized` state in Java client.

> Closing and Closed seems not much different in c++ client.

See the comments in the Java client. `Closing` means the `CloseProducer` command has been sent to the broker and the response is not received yet. In `Closing` state, we should not send any `CloseProducer` command again. It could also be helpful if `producer.close()` stuck. We can debug the application and see the state.

BewareMyPower (approved):

RobertIndie (approved):

lib/PartitionedProducerImpl.cc
erobot: Thanks for your explanation, I found read wrong code. I'm understand the problem better now.

PR #127, opened 2022-11-25
[fix] Fix NamedEntity::checkName regression

erobot:
### Motivation

Namspace name with underscore is valid before but now invalid. This regression caused by #52.

Broker validates NamedEntity using regex `"^[-=:.\\w]*$"`, which actually includes the underscore. Although the comment saying `alphanumeric (a-za-z_0-9) and these special characters -=:` is a bit confusing. But we should not break the actual validation rules used a long time ago.

https://github.com/apache/pulsar/blob/355ba6002fc33b2716f59ca5d3a4fac667ecf3e1/pulsar-common/src/main/java/org/apache/pulsar/common/naming/NamedEntity.java#L30-L33

### Modifications

Make underscore valid in NamedEntity.

### Verifying this change

This change can covered by existing tests with some modifications, such as NamespaceNameTest.testNamespaceNameLegalCharacters and TopicNameTest.testLegalNonAlphaCharacters.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
fix regression only

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/NamedEntity.cc
+++ b/lib/NamedEntity.cc
@@ -22,7 +22,7 @@
 
 /**
  * Allowed characters for property, namespace, cluster and topic names are
- * alphanumeric (a-zA-Z_0-9) and these special chars -=:.
+ * alphanumeric (a-zA-Z0-9) and these special chars _-=:.
  * @param name
  * @return
  */
@@ -33,6 +33,7 @@ bool NamedEntity::checkName(const std::string& name) {
         }
 
         switch (c) {
+            case '_':
             case '-':
             case '=':
             case ':':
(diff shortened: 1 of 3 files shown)


BewareMyPower (approved):
LGTM.

BewareMyPower:
`_` should not be treated as a "special character". Even in C++ regex, "\\w" also includes the understore. Since the 3.1.0 is during the release process, I hope we can include this fix. /cc @RobertIndie @merlimat

RobertIndie (approved):

shibd (approved):

PR #128, opened 2022-11-29
[fix] Fix acknowledge MessageId list does not work when ackGroupingTimeMs is 0

BewareMyPower:
### Motivation

When `ackGroupingTimeMs` is 0, `AckGroupingTrackerDisabled` will be created to acknowledge messages immediately. However, since the `addAcknowledgeList` method is not implemented in #23, it doesn't work.

### Modifications

Implement `AckGroupingTrackerDisabled::addAcknowledgeList`. Move the protocol version validation from `AckGroupingTrackerEnabled` to the base class `AckGroupingTracker` so that the `doImmediateAck` overload that accepts `std::set<MessageId>` can be reused.

Add testing params to `testAckMsgList` and
`testAckMsgListWithMultiConsumer` so that various `ackGroupingTimeMs` can be verified.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/BasicEndToEndTest.cc
+++ b/tests/BasicEndToEndTest.cc
@@ -20,6 +20,7 @@
 #include <pulsar/Client.h>
 
 #include <algorithm>
+#include <atomic>
 #include <chrono>
 #include <cstring>
 #include <functional>
@@ -58,7 +59,7 @@ static int globalCount = 0;
 static long globalResendMessageCount = 0;
 std::string lookupUrl = "pulsar://localhost:6650";
 static std::string adminUrl = "http://localhost:8080/";
-static int uniqueCounter = 0;
+static std::atomic_int uniqueCounter{0};
 
 std::string unique_str() {
     long nanos = std::chrono::duration_cast<std::chrono::milliseconds>(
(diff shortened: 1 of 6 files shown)


BewareMyPower:
@RobertIndie Please also include this PR in 3.1.0 since the VOTE of candidate-2 is not present yet.

shibd (approved):

RobertIndie (approved):

tests/BasicEndToEndTest.cc
```diff
@@ -4277,7 +4277,9 @@ TEST(BasicEndToEndTest, testBatchReceiveClose) { testBatchReceiveClose(false); }
 
 TEST(BasicEndToEndTest, testBatchReceiveCloseWithMultiConsumer) { testBatchReceiveClose(true); }
 
-TEST(BasicEndToEndTest, testAckMsgList) {
+class AcknowledgeTest : public testing::TestWithParam<int> {};
```
RobertIndie, line 4280: How about moving this test class to a separate file.

BewareMyPower: It make sense to me.

BTW, the `BasicEndToEndTest` is too huge, adding a test into this file requires much more compile time than others. The original purpose is to use another PR to split tests in this file. For this PR, I will split the test class to another file soon.

PR #129, opened 2022-11-29
[improve] Add configuration to limit times of client's lookup redirection.

tongsucn:
Master Issue: apache/pulsar#7096

### Motivation

C++ client's `BinaryProtoLookupService` does not limit the times of client's lookup redirection. It may result in broker overloading as described in apache/pulsar#7096.

### Modifications

1. Added max. look redirection limit in `BinaryProtoLookupService::findBroker`.
2. Added configuration to control max times of lookup redirection in both `BinaryProtoLookupService` and `HTTPLookupService`.
3. Added configuration to control initial- and max-backoff interval in both `ProducerImpl` and `ConsumerImpl`.

### Verifying this change

### Test cases are on the way

*(example:)*

### Documentation

- [ ] `doc-required`

- [ ] `doc-not-needed`

- [ ] `doc` 

- [x] `doc-complete`

### DIFF
--- a/include/pulsar/ClientConfiguration.h
+++ b/include/pulsar/ClientConfiguration.h
@@ -64,12 +64,12 @@ class PULSAR_PUBLIC ClientConfiguration {
      *
      * @param timeout the timeout after which the operation will be considered as failed
      */
-    ClientConfiguration& setOperationTimeoutSeconds(int timeout);
+    ClientConfiguration& setOperationTimeoutSeconds(int32_t timeout);
 
     /**
      * @return the client operations timeout in seconds
      */
-    int getOperationTimeoutSeconds() const;
+    int32_t getOperationTimeoutSeconds() const;
 
     /**
      * Set the number of IO threads to be used by the Pulsar client. Default is 1
@@ -77,12 +77,12 @@ class PULSAR_PUBLIC ClientConfiguration {
      *
      * @param threads number of threads
      */
-    ClientConfiguration& setIOThreads(int threads);
+    ClientConfiguration& setIOThreads(int32_t threads);
 
     /**
      * @return the number of IO threads to use
      */
-    int getIOThreads() const;
+    int32_t getIOThreads() const;
 
     /**
      * Set the number of threads to be used by the Pulsar client when delivering messages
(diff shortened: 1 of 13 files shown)


BewareMyPower:
Please rebase to master to avoid broken CI

tongsucn:
## Update

I'm still working on it. I found another potential bug in C++ client. I described it in apache/pulsar#13669: _when unloading bundles with non-persistent topics, the unloaded broker gets stuck until it's killed_.

At that time, I thought it was a bug in broker, and the problem still exists in later versions of brokers. However, a few days ago, I tried writing data to non-persistent topics via Flink and the problem cannot be reproduced. Therefore I think the problem is caused by the C++ client.

For now, the problem can be reproduced by sending data with `pulsar::BinaryProtoLookupService` (i.e. using cluster address starting with `pulsar://`) to lookup topics, and `pulsar::HTTPLookupService` (`http://`) works well.

After adding the lookup-redirection-limit logic, the problem persists, so I'm still locating the problem. If you have any ideas or suggestions on this problem, please let me know.

tongsucn:
> ## Update
> I'm still working on it. I found another potential bug in C++ client. ...

FYI the problem described here is solved by my colleague in #146 , I'll rebase to `main` and start updating test cases.

BewareMyPower:
I think it's okay to change it to a protected function. Users should never use this function.

tongsucn:
FYI related tests has been added.

lib/HandlerBase.h
```diff
@@ -113,13 +113,14 @@ class HandlerBase {
...
-        Closed,
-        Failed,
-        Producer_Fenced
+        NotStarted,       // Not initialized, in Java client: HandlerState.State.Uninitialized
+        Pending,          // Client connecting to broker, in Java client: HandlerState.State.Connecting
+        Ready,            // Handler is being used, in Java client: HandlerState.State.Ready
+        Closing,          // Close cmd has been sent to broker, in Java client: HandlerState.State.Closing
+        Closed,           // Broker acked the close, in Java client: HandlerState.State.Closed
+        Failed,           // Handler is failed, in Java client: HandlerState.State.Failed
+        Producer_Fenced,  // The producer has been fenced by the broker
```
BewareMyPower, line 122: Why not use a name `ProducerFenced`? It's better to keep the variable naming style consistent.

BewareMyPower: I see it's added in a previous PR, just ignore my comment.

include/pulsar/ClientConfiguration.h
```diff
@@ -64,25 +64,25 @@ class PULSAR_PUBLIC ClientConfiguration {
      *
      * @param timeout the timeout after which the operation will be considered as failed
      */
-    ClientConfiguration& setOperationTimeoutSeconds(int timeout);
+    ClientConfiguration& setOperationTimeoutSeconds(int32_t timeout);
```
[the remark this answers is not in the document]
tongsucn: I'll change them later this week

tongsucn:
@BewareMyPower Hi, the integer type in `*Configuration.[h|cpp]` has be fixed.

BewareMyPower (approved):

PR #131, opened 2022-11-30
[fix] Close broker producer created after producer close

erobot:
### Motivation

Close broker producer created after producer close to prevent producers from keeping alive in broker, and fix a race condition here.

Race condition sequence:
1. prodcuer.start() is called
2. handleCreateProducer() is called, and runs to right before setCnx()
3. closeAsync() is called and returns directly after getCnx() returns nullptr, so no CloseProducer cmd is sent
4. handleCreateProducer() continues
5. Result: Producer is closed and no CloseProducer cmd is sent to broker

### Modifications

* Close broker producer created after producer close
* Use mutex to prevent race condition. handleCreateProducer() and closeAsync() should be low frequency operations and using mutex here should not affect performance.

### Verifying this change

* Added unit test ProducerTest.testCloseProducerBeforeCreated

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
bug fix only

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ProducerImpl.cc
+++ b/lib/ProducerImpl.cc
@@ -172,6 +172,8 @@ void ProducerImpl::connectionFailed(Result result) {
 
 void ProducerImpl::handleCreateProducer(const ClientConnectionPtr& cnx, Result result,
                                         const ResponseData& responseData) {
+    Lock lock(mutex_);
+
     LOG_DEBUG(getName() << "ProducerImpl::handleCreateProducer res: " << strResult(result));
 
     // make sure we're still in the Pending/Ready state, closeAsync could have been invoked
@@ -180,11 +182,18 @@ void ProducerImpl::handleCreateProducer(const ClientConnectionPtr& cnx, Result r
     if (state != Ready && state != Pending) {
         LOG_DEBUG("Producer created response received but producer already closed");
         failPendingMessages(ResultAlreadyClosed, false);
+        if (result == ResultOk || result == ResultTimeout) {
+            int requestId = client_.lock()->newRequestId();
+            cnx->sendRequestWithId(Commands::newCloseProducer(producerId_, requestId), requestId);
+        }
+        if (!producerCreatedPromise_.isComplete()) {
+            lock.unlock();
+            producerCreatedPromise_.setFailed(ResultAlreadyClosed);
+        }
         return;
     }
 
     if (result == ResultOk) {
-        Lock lock(mutex_);
         // We are now reconnected to broker and clear to send messages. Re-send all pending messages and
         // set the cnx pointer so that new messages will be sent immediately
         LOG_INFO(getName() << "Created producer on broker " << cnx->cnxString());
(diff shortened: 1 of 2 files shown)


RobertIndie (approved):

lib/ProducerImpl.cc
```diff
@@ -172,6 +172,8 @@ void ProducerImpl::connectionFailed(Result result) {
 
 void ProducerImpl::handleCreateProducer(const ClientConnectionPtr& cnx, Result result,
                                         const ResponseData& responseData) {
+    Lock lock(mutex_);
```
BewareMyPower, line 175: Could you explain more about the race condition you mentioned in this PR? I agree it does not affect performance significantly, but it should be reasonable to use a lock here.

erobot: I have updated the comment to describe the race condition about cnx.

BewareMyPower: Thanks for your explanation. I missed the fact that you moved the `mutex` in line 187 here.

lib/ProducerImpl.cc
```diff
@@ -180,11 +182,18 @@ void ProducerImpl::handleCreateProducer(const ClientConnectionPtr& cnx, Result r
     if (state != Ready && state != Pending) {
         LOG_DEBUG("Producer created response received but producer already closed");
         failPendingMessages(ResultAlreadyClosed, false);
+        if (result == ResultOk || result == ResultTimeout) {
+            int requestId = client_.lock()->newRequestId();
```
BewareMyPower, line 186: Is it guaranteed that `client_.lock()` doesn't return a nullptr here?

lib/ProducerImpl.cc
```diff
@@ -694,6 +705,8 @@ void ProducerImpl::closeAsync(CloseCallback originalCallback) {
         }
     };
 
+    Lock lock(mutex_);
```
BewareMyPower, line 708: Could you use another mutex to synchronize `closeAsync` and `handleCreatedProducer`? The `mutex_` is already used in other places and it's not clear why the mutex is required here from the code.

BewareMyPower: For example, if someone wanted to remove the unnecessary mutex uses in future, the mutex here would be likely to be removed because the state is atomic, and the following code like cancelling the timers and failing the pending messages doesn't require a lock.

lib/ProducerImpl.cc
erobot: May not be guaranteed. I copy the codes below and will add a check for both.

lib/ProducerImpl.cc
erobot: My understanding is that the race condition problem here is about state assumption. `closeAsync` checks state_ then have a assumption about cnx. `handleCreatedProducer` checks state_ then have a assumption that don't need to close producer.

mutex_ here is like a global producer mutex. When performance is not critical, lock on a global mutex is simpler and easy to keep correction. Add new dedicated mutex should be possible, but would add complexity. Should we keep it simple for now, and refactor latter if necessary?

BewareMyPower: It makes sense to me.

lib/ProducerImpl.cc
erobot: Added check for client

BewareMyPower:
Please rebase to master to avoid broken CI

BewareMyPower (approved):

shibd (approved):

PR #133, opened 2022-12-01
Fix the broken master by the upgrade of GTest

BewareMyPower:
### Motivation

The `ubuntu:latest` image became `ubuntu:22.04` after https://github.com/actions/runner-images/issues/6399. The GTest is no longer compatible with the current code so the CI failed.

### Modifications

Pass `int` to `TestWithParam` and stick the image version to `ubuntu-22.04` instead of `ubuntu-latest`.

### DIFF
--- a/.github/workflows/ci-pr-validation.yaml
+++ b/.github/workflows/ci-pr-validation.yaml
@@ -31,7 +31,7 @@ jobs:
 
   unit-tests:
     name: Run unit tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     timeout-minutes: 120
 
     steps:
--- a/tests/KeyValueSchemaTest.cc
+++ b/tests/KeyValueSchemaTest.cc
@@ -25,6 +25,12 @@ using namespace pulsar;
 
 static const std::string lookupUrl = "pulsar://localhost:6650";
 
+// NOTE: the default operator<< for KeyValueEncodingType (value, not const reference) might not work on some
+// GTest implementations. We need to implement the operator<< for const reference.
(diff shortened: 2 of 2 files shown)


BewareMyPower:
@erobot It seems to be a bug of gtest. I fixed it in another way (add an inline overload for const reference). Did you have any thought?

BewareMyPower:
Not sure if it's a bug for GCC 11 or GTest itself. Let's continue this fix because it doesn't change the public API.

tisonkun (approved):
LGTM.

In another perspective, I'd prefer a fixed version of `run-os` instead of `latest` so that we explicit do the upgration on demand.

BewareMyPower:
> I'd prefer a fixed version of run-os instead of latest so that we explicit do the upgration on demand.

Agreed. I will modify it in this PR.

BewareMyPower:
I opened an issue here: https://github.com/google/googletest/issues/4079.

BewareMyPower:
Now the tests work for Ubuntu 22.04 (GCC 11) but it failed for Ubuntu 20.04 (GCC 9) in my local env.

```
/usr/include/gtest/gtest-printers.h:287:7: error: ambiguous overload for ‘operator<<’ (operand types are ‘std::ostream’ {aka ‘std::basic_ostream<char>’} and ‘const pulsar::KeyValueEncodingType’)
  287 |   *os << value;
      |   ~~~~^~~~~~~~
```

I decided to use another way to fix to avoid API changes.

Demogorgon314 (approved):

RobertIndie (approved):

tisonkun (approved):

PR #138, opened 2022-12-06
Boost optional

### LINKED ISSUE #134: [Enhancement]  Do not use custom written Optional<T> class
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

repo/master

### Minimal reproduce step

template <> class Optional;

Use optional class from boost or std

### What did you see instead?

Used custom written Optional class

### Anything else?

At this moment pulsar-client-cpp heavily used boost and std library. I suggest to use boost::optional instead of Optional class.
Boost::optional well-tested and has similar API with std::optional (at most). 
In this issue I do not suggest to increase C++ standart to C++17 (from C++11) where std::optional is available, but in the future it can be done easily.

- [X] I'm willing to submit a PR!

fregate:
Fixes #134

Master Issue: #<xyz>

### Motivation

Switch to use boost::optional instead self-written Optional class

### Modifications

Remove class Optional in Utils.h and change all usage to boost::optional

### Verifying this change

*(or)*

### Documentation

- [ * ] `doc-not-needed` 
This is internal structure, not a public API

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -512,7 +512,7 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
         m.impl_->convertPayloadToKeyValue(config_.getSchema());
 
         const auto startMessageId = startMessageId_.get();
-        if (isPersistent_ && startMessageId.has_value() &&
+        if (isPersistent_ && startMessageId &&
             m.getMessageId().ledgerId() == startMessageId.value().ledgerId() &&
             m.getMessageId().entryId() == startMessageId.value().entryId() &&
             isPriorEntryIndex(m.getMessageId().entryId())) {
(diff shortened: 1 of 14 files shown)


lib/ConsumerImpl.cc
```diff
@@ -512,7 +512,7 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
         m.impl_->convertPayloadToKeyValue(config_.getSchema());
 
         const auto startMessageId = startMessageId_.get();
-        if (isPersistent_ && startMessageId.is_present() &&
+        if (isPersistent_ && startMessageId.has_value() &&
```
shibd, line 515: Uniform style:
```suggestion
        if (isPersistent_ && startMessageId &&
```

shibd (approved):

BewareMyPower (approved):

Issue #114, opened 2019-11-15
[C++ client] Deadletter queue functionality


frejonb:
Is anyone working on porting DLQ from the Java client to Cpp https://github.com/apache/pulsar/pull/2508? I'd like to help/contribute.

I'm thinking on the following API
```c++
Client client("pulsar://localhost:6650");

Consumer consumer;
ConsumerConfiguration config;
config.setMaxRedeliverCount(10);
config.setDeadLetterTopic("my-dlq-topic");

Result result = client.subscribe("persistent://prop/r1/ns1/my-topic", "consumer-1", config, consumer);
```

shibd:
Related: https://github.com/apache/pulsar-client-cpp/issues/77

I will start working on this. @tisonkun Please assign it to me.

tisonkun:
@shibd assigned. Go ahead!

Issue #77, opened 2022-08-17
[feature request] C++ and Python client support Dead Letter Topic


shibd:

BewareMyPower:
Close it by #139

PR #139, opened 2022-12-06
[feat] Support Dead Letter Topic.

shibd:
Master Issue: #114 

### Motivation
#77, #114 

### Modifications
-  Add DeadLetterPolicy. When the message `redeliveryCount > DeadLetterPolicy.maxRedeliveryCount`, the message is sent to the DLQ topic and ack this msg.

### Verifying this change

1. Add DeadLetterQueueTest to verify Send to DLQ, according to the following scenarios(Cartesian product of the following conditions): 
- Producer batch [enabled, disable]
- Topic is [single, multi]
- Consumer sub type [Shared, Key_Shard]
- Trigger by [unack, ack_timeout]

2. Add `DeadLetterQueueTest.testInitSubscription` to cover init DLQ topic subscription: https://github.com/apache/pulsar/pull/13355

3. Add `DeadLetterQueueTest.testAutoSetDLQTopicName` to cover default DLQ topic name.

4. Add `DeadLetterQueueTest.testDLQWithSchema`  to cover producer and consumer with the schema scenario.

5. If a topic has a schema, and the consumer uses `AUTO_CONSUME` schema, and DLQTopic also has the same schema. In this case, the DLQ producer fails to create(IncompatibleSchema).  Refer to java client same issue: https://github.com/apache/pulsar/issues/9935.
**Solution**: Depending on #157, After this, DLQ producers use the `autoDownloadSchema` param when consumers use `AUTO_CONSUM`. 

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/ConsumerConfiguration.h
+++ b/include/pulsar/ConsumerConfiguration.h
@@ -35,11 +35,13 @@
 #include <memory>
 
 #include "BatchReceivePolicy.h"
+#include "DeadLetterPolicy.h"
 
 namespace pulsar {
 
 class Consumer;
 class PulsarWrapper;
+class PulsarFriend;
 
 /// Callback definition for non-data operation
 typedef std::vector<Message> Messages;
@@ -409,6 +411,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
     const BatchReceivePolicy& getBatchReceivePolicy() const;
 
     /**
+     * Set dead letter policy for consumer
+     *
+     * By default, some messages are redelivered many times, even to the extent that they can never be
+     * stopped. By using the dead letter mechanism, messages have the max redelivery count, when they
+     * exceeding the maximum number of redeliveries. Messages are sent to dead letter topics and acknowledged
+     * automatically.
+     *
+     * You can enable the dead letter mechanism by setting the dead letter policy.
+     * Example:
+     *
+     * <pre>
+     * * DeadLetterPolicy dlqPolicy = DeadLetterPolicyBuilder()
+     *                       .maxRedeliverCount(10)
+     *                       .build();
+     * </pre>
+     * Default dead letter topic name is {TopicName}-{Subscription}-DLQ.
+     * To set a custom dead letter topic name
+     * <pre>
+     * DeadLetterPolicy dlqPolicy = DeadLetterPolicyBuilder()
+     *                       .deadLetterTopic("dlq-topic")
+     *                       .maxRedeliverCount(10)
+     *                       .initialSubscriptionName("init-sub-name")
+     *                       .build();
+     * </pre>
+     * @param deadLetterPolicy Default value is empty
+     */
+    void setDeadLetterPolicy(const DeadLetterPolicy& deadLetterPolicy);
+
+    /**
+     * Get dead letter policy.
+     *
+     * @return dead letter policy
+     */
+    const DeadLetterPolicy& getDeadLetterPolicy() const;
+
+    /**
      * Set whether the subscription status should be replicated.
      * The default value is `false`.
      *
@@ -581,6 +619,7 @@ class PULSAR_PUBLIC ConsumerConfiguration {
     bool isBatchIndexAckEnabled() const;
 
     friend class PulsarWrapper;
+    friend class PulsarFriend;
 
    private:
     std::shared_ptr<ConsumerConfigurationImpl> impl_;
--- a/include/pulsar/DeadLetterPolicy.h
+++ b/include/pulsar/DeadLetterPolicy.h
@@ -1,0 +1,71 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef DEAD_LETTER_POLICY_HPP_
+#define DEAD_LETTER_POLICY_HPP_
+
+#include <pulsar/defines.h>
+
+#include <memory>
+#include <string>
+
+namespace pulsar {
+
+struct DeadLetterPolicyImpl;
+
+/**
+ * Configuration for the "dead letter queue" feature in consumer.
+ *
+ * see @DeadLetterPolicyBuilder
+ */
+class PULSAR_PUBLIC DeadLetterPolicy {
+   public:
+    DeadLetterPolicy();
+
+    /**
+     * Get dead letter topic
+     *
+     * @return
+     */
+    const std::string& getDeadLetterTopic() const;
+
+    /**
+     * Get max redeliver count
+     *
+     * @return
+     */
+    int getMaxRedeliverCount() const;
+
+    /**
+     * Get initial subscription name
+     *
+     * @return
+     */
+    const std::string& getInitialSubscriptionName() const;
+
+   private:
+    friend class DeadLetterPolicyBuilder;
+
+    typedef std::shared_ptr<DeadLetterPolicyImpl> DeadLetterPolicyImplPtr;
+    DeadLetterPolicyImplPtr impl_;
+
+    explicit DeadLetterPolicy(const DeadLetterPolicyImplPtr& impl);
+};
+}  // namespace pulsar
+
+#endif /* DEAD_LETTER_POLICY_HPP_ */
--- a/include/pulsar/DeadLetterPolicyBuilder.h
+++ b/include/pulsar/DeadLetterPolicyBuilder.h
@@ -1,0 +1,94 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#ifndef DEAD_LETTER_POLICY_BUILD_HPP_
+#define DEAD_LETTER_POLICY_BUILD_HPP_
+
+#include <pulsar/DeadLetterPolicy.h>
+#include <pulsar/defines.h>
+
+#include <memory>
+
+namespace pulsar {
+
+struct DeadLetterPolicyImpl;
+
+/**
+ * The builder to build a DeadLetterPolicyBuilder
+ *
+ * Example of building DeadLetterPolicy:
+ *
+ * ```c++
+ * DeadLetterPolicy dlqPolicy = DeadLetterPolicyBuilder()
+ *                       .deadLetterTopic("dlq-topic")
+ *                       .maxRedeliverCount(10)
+ *                       .initialSubscriptionName("init-sub-name")
+ *                       .build();
+ * ```
+ */
+class PULSAR_PUBLIC DeadLetterPolicyBuilder {
+   public:
+    DeadLetterPolicyBuilder();
+
+    /**
+     * Set dead letter topic.
+     *
+     * @param deadLetterTopic Name of the dead topic where the failing messages are sent.
+     * The default value is: sourceTopicName + "-" + subscriptionName + "-DLQ"
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& deadLetterTopic(const std::string& deadLetterTopic);
+
+    /**
+     * Set max redeliver count
+     *
+     * @param maxRedeliverCount Maximum number of times that a message is redelivered before being sent
+     * to the dead letter queue.
+     * - The maxRedeliverCount must be greater than 0.
+     * - The default value is {INT_MAX} (DLQ is not enabled)
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& maxRedeliverCount(int maxRedeliverCount);
+
+    /**
+     * Set initial subscription name
+     *
+     * @param initialSubscriptionName  Name of the initial subscription name of the dead letter topic.
+     * If this field is not set, the initial subscription for the dead letter topic is not created.
+     * If this field is set but the broker's `allowAutoSubscriptionCreation` is disabled, the DLQ producer
+     * fails to be created.
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& initialSubscriptionName(const std::string& initialSubscriptionName);
+
+    /**
+     * Build DeadLetterPolicy.
+     *
+     * @return
+     */
+    DeadLetterPolicy build();
+
+   private:
+    std::shared_ptr<DeadLetterPolicyImpl> impl_;
+};
+}  // namespace pulsar
+
+#endif /* DEAD_LETTER_POLICY_BUILD_HPP_ */
--- a/include/pulsar/ProducerConfiguration.h
+++ b/include/pulsar/ProducerConfiguration.h
@@ -532,11 +532,12 @@ class PULSAR_PUBLIC ProducerConfiguration {
      */
     ProducerAccessMode getAccessMode() const;
 
-    friend class PulsarWrapper;
-
    private:
-    struct Impl;
     std::shared_ptr<ProducerConfigurationImpl> impl_;
+
+    friend class PulsarWrapper;
+    friend class ConsumerImpl;
+    friend class ProducerImpl;
 };
 }  // namespace pulsar
 #endif /* PULSAR_PRODUCERCONFIGURATION_H_ */
--- a/lib/Commands.h
+++ b/lib/Commands.h
@@ -111,7 +111,8 @@ class Commands {
                                     const std::map<std::string, std::string>& metadata,
                                     const SchemaInfo& schemaInfo, uint64_t epoch,
                                     bool userProvidedProducerName, bool encrypted,
-                                    ProducerAccessMode accessMode, boost::optional<uint64_t> topicEpoch);
+                                    ProducerAccessMode accessMode, boost::optional<uint64_t> topicEpoch,
+                                    const std::string& initialSubscriptionName);
 
     static SharedBuffer newAck(uint64_t consumerId, int64_t ledgerId, int64_t entryId, const BitSet& ackSet,
                                CommandAck_AckType ackType, CommandAck_ValidationError validationError);
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -18,6 +18,7 @@
  */
 #include "ConsumerImpl.h"
 
+#include <pulsar/DeadLetterPolicyBuilder.h>
 #include <pulsar/MessageIdBuilder.h>
 
 #include <algorithm>
@@ -39,6 +40,7 @@
 #include "MessageIdUtil.h"
 #include "MessageImpl.h"
 #include "MessagesImpl.h"
+#include "ProducerConfigurationImpl.h"
 #include "PulsarApi.pb.h"
 #include "TimeUtils.h"
 #include "TopicName.h"
@@ -115,6 +117,21 @@ ConsumerImpl::ConsumerImpl(const ClientImplPtr client, const std::string& topic,
     // Create msgCrypto
     if (conf.isEncryptionEnabled()) {
         msgCrypto_ = std::make_shared<MessageCrypto>(consumerStr_, false);
+    }
+
+    // Config dlq
+    auto deadLetterPolicy = conf.getDeadLetterPolicy();
+    if (deadLetterPolicy.getMaxRedeliverCount() > 0) {
+        auto deadLetterPolicyBuilder =
+            DeadLetterPolicyBuilder()
+                .maxRedeliverCount(deadLetterPolicy.getMaxRedeliverCount())
+                .initialSubscriptionName(deadLetterPolicy.getInitialSubscriptionName());
+        if (deadLetterPolicy.getDeadLetterTopic().empty()) {
+            deadLetterPolicyBuilder.deadLetterTopic(topic + "-" + subscriptionName + DLQ_GROUP_TOPIC_SUFFIX);
+        } else {
+            deadLetterPolicyBuilder.deadLetterTopic(deadLetterPolicy.getDeadLetterTopic());
+        }
+        deadLetterPolicy_ = deadLetterPolicyBuilder.build();
     }
 
     checkExpiredChunkedTimer_ = executor_->createDeadlineTimer();
@@ -242,6 +259,7 @@ void ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result r
             Lock lock(mutex_);
             setCnx(cnx);
             incomingMessages_.clear();
+            possibleSendToDeadLetterTopicMessages_.clear();
             state_ = Ready;
             backoff_.reset();
             // Complicated logic since we don't have a isLocked() function for mutex
@@ -467,6 +485,7 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
         return;
     }
 
+    auto redeliveryCount = msg.redelivery_count();
     const bool isMessageUndecryptable =
         metadata.encryption_keys_size() > 0 && !config_.getCryptoKeyReader().get() &&
         config_.getCryptoFailureAction() == ConsumerCryptoFailureAction::CONSUME;
@@ -532,6 +551,14 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
             LOG_DEBUG(getName() << " Ignoring message from before the startMessageId: "
                                 << startMessageId.value());
             return;
+        }
+        if (redeliveryCount >= deadLetterPolicy_.getMaxRedeliverCount()) {
+            possibleSendToDeadLetterTopicMessages_.emplace(m.getMessageId(), std::vector<Message>{m});
+            if (redeliveryCount > deadLetterPolicy_.getMaxRedeliverCount()) {
+                redeliverUnacknowledgedMessages({m.getMessageId()});
+                increaseAvailablePermits(cnx);
+                return;
+            }
         }
         executeNotifyCallback(m);
     }
@@ -644,12 +671,21 @@ uint32_t ConsumerImpl::receiveIndividualMessagesFromBatch(const ClientConnection
     int skippedMessages = 0;
 
     auto acker = BatchMessageAckerImpl::create(batchSize);
+    std::vector<Message> possibleToDeadLetter;
     for (int i = 0; i < batchSize; i++) {
         // This is a cheap copy since message contains only one shared pointer (impl_)
         Message msg = Commands::deSerializeSingleMessageInBatch(batchedMessage, i, batchSize, acker);
         msg.impl_->setRedeliveryCount(redeliveryCount);
         msg.impl_->setTopicName(batchedMessage.getTopicName());
         msg.impl_->convertPayloadToKeyValue(config_.getSchema());
+
+        if (redeliveryCount >= deadLetterPolicy_.getMaxRedeliverCount()) {
+            possibleToDeadLetter.emplace_back(msg);
+            if (redeliveryCount > deadLetterPolicy_.getMaxRedeliverCount()) {
+                skippedMessages++;
+                continue;
+            }
+        }
 
         if (startMessageId) {
             const MessageId& msgId = msg.getMessageId();
@@ -674,6 +710,13 @@ uint32_t ConsumerImpl::receiveIndividualMessagesFromBatch(const ClientConnection
         }
 
         executeNotifyCallback(msg);
+    }
+
+    if (!possibleToDeadLetter.empty()) {
+        possibleSendToDeadLetterTopicMessages_.emplace(batchedMessage.getMessageId(), possibleToDeadLetter);
+        if (redeliveryCount > deadLetterPolicy_.getMaxRedeliverCount()) {
+            redeliverUnacknowledgedMessages({batchedMessage.getMessageId()});
+        }
     }
 
     if (skippedMessages > 0) {
@@ -1080,6 +1123,7 @@ std::pair<MessageId, bool> ConsumerImpl::prepareIndividualAck(const MessageId& m
         consumerStatsBasePtr_->messageAcknowledged(ResultOk, CommandAck_AckType_Individual,
                                                    (batchSize > 0) ? batchSize : 1);
         unAckedMessageTrackerPtr_->remove(messageId);
+        possibleSendToDeadLetterTopicMessages_.remove(messageId);
         return std::make_pair(discardBatch(messageId), true);
     } else if (config_.isBatchIndexAckEnabled()) {
         return std::make_pair(messageId, true);
@@ -1169,6 +1213,7 @@ const std::string& ConsumerImpl::getName() const { return consumerStr_; }
 
 void ConsumerImpl::shutdown() {
     incomingMessages_.clear();
+    possibleSendToDeadLetterTopicMessages_.clear();
     resetCnx();
     auto client = client_.lock();
     if (client) {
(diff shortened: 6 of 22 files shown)


lib/DeadLetterPolicyImpl.h
```diff
@@ -0,0 +1,32 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#pragma once
+
+#include "climits"
+#include "string"
```
BewareMyPower, line 22:
```suggestion
#include <climits>
#include <string>
```

We usually use angle brackets for headers from system libraries.

include/pulsar/DeadLetterPolicyBuilder.h
```diff
@@ -0,0 +1,81 @@
...
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& deadLetterTopic(const std::string& deadLetterTopic);
+
+    /**
+     * Set max redeliver count
+     *
+     * @return
+     */
```
BewareMyPower, line 59: When you add the API docs, please add the default value and explain the behavior. For example, from the code in `ConsumerImpl.cc` I see the dead letter policy only works when `maxRedeliverCount <= 0`. And what if users didn't set the dead letter topic? Please make API docs readable and meaningful.

BewareMyPower (changes requested):

lib/ConsumerImpl.cc
```diff
@@ -460,6 +476,7 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
         return;
     }
 
+    auto redeliveryCount = msg.redelivery_count();
```
BewareMyPower, line 479: It seems that we need to pass the `redeliveryCount` to the `decryptMessageIfNeeded` method before? When the decryption failed, `unAckedMessageTracker.add(m, redeliveryCount);` should be called. (Though I didn't find if there is a related test)

tests/ConsumerConfigurationTest.cc
```diff
@@ -20,12 +20,15 @@
...
 
+#include <climits>
+
 #include "NoOpsCryptoKeyReader.h"
 
 DECLARE_LOG_OBJECT()
 
 #include "../lib/Future.h"
 #include "../lib/Utils.h"
+#include "pulsar/DeadLetterPolicyBuilder.h"
```
BewareMyPower, line 31:
```suggestion
#include <pulsar/DeadLetterPolicyBuilder.h>
```

Make include style consistent

test-conf/standalone-ssl.conf
```diff
@@ -50,6 +50,9 @@ brokerShutdownTimeoutMs=3000
 # Enable backlog quota check. Enforces action on topic when the quota is reached
 backlogQuotaCheckEnabled=true
 
+# Disable schema validation: If a producer doesn’t carry a schema, the producer is allowed to connect to the topic and produce data.
+isSchemaValidationEnforced=true
+
```
BewareMyPower, line 55: IMO, currently we don't need to handle this case. The default value of `isSchemaValidationEnforced` is false.

I think you can continue the work of https://github.com/apache/pulsar-client-cpp/pull/142, but the goal is not supporting auto produce schema. Because C++ client only interacts with `SchemaInfo`, while `AutoProduceBytesSchema` is a `Schema<byte[]>`.

I have an idea that we can add an option to download the latest schema info when creating producers without schema. e.g.

```
// it should only work for the case when isSchemaValidationEnforced is true
producerConf.setAutoDownloadSchema(true);
```

Regarding the message validation, users should still perform the message validation by themselves. From the discussion [here](https://lists.apache.org/thread/2gbftzs7xwvvddythn818bor17rqpd4m), we can see Java client does not validate the message in `sendAsync`. Instead, it validates the message in `TypedMessageBuilder#value`. There is no corresponding concept in the C++ client.

shibd: > IMO, currently we don't need to handle this case. The default value of isSchemaValidationEnforced is false.

I think we need to turn this. We should ensure that the schema still works when `isSchemaValidationEnforced` is true. 

For example, DLQ producers create scenarios that fail (refer: verifying this change[5]), If we set `isSchemaValidationEnforced` to false, then the test passes.


> I have an idea that we can add an option to download the latest schema info when creating producers without schema. e.g.

Good point.  I will change #142 to support auto download schema first.

BWT: I'm not going to expose this interface to users, just use it in DLQ for the time being.

lib/ConsumerImpl.cc
shibd: It seems that this is another problem: we need to give the client that cannot decrypt the message a chance to re-decrypt the message, so we need to put it in `unAckedMessageTracker ` and wait for re-delivery.

https://github.com/apache/pulsar/pull/3097 


As for why `redeliveryCount` is passed into `decryptPayloadIfNeeded` methods in the Java implementation, it is because of the support for [PIP-130](https://github.com/apache/pulsar/pull/13707)


Note: I'll open a new PR to support both issues.

include/pulsar/DeadLetterPolicy.h
```diff
@@ -0,0 +1,71 @@
...
+class PULSAR_PUBLIC DeadLetterPolicy {
+   public:
+    DeadLetterPolicy();
+
+    /**
+     * Get dead letter topic
+     *
+     * @return
+     */
+    std::string getDeadLetterTopic() const;
```
BewareMyPower, line 45: It's okay to return by value here because the copy cost of a topic name is not high. But we should keep the code style consistent, I see `ProducerConfiguration::getInitialSubscriptionName` returns a const reference but `DeadLetterPolicy::getInitialSubscriptionName` returns a value.

include/pulsar/ProducerConfiguration.h
```diff
@@ -532,6 +532,18 @@ class PULSAR_PUBLIC ProducerConfiguration {
...
+     *
+     * If this field is not set, the initial subscription is not created.
+     */
+    ProducerConfiguration& setInitialSubscriptionName(const std::string& initialSubscriptionName);
+
+    /**
+     * Get initial subscription name.
+     */
+    const std::string& getInitialSubscriptionName() const;
+
```
BewareMyPower, line 546: These methods should not be exposed to users, see https://github.com/apache/pulsar/pull/13355/files. We should access `ProducerConfigurationImpl::initialSubscriptionName`  directly.

BewareMyPower (changes requested):

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,390 @@
...
+        for (const auto isMultiTopic : isMultiTopics) {
+            for (const auto subType : subTypes) {
+                values.emplace_back(std::make_tuple(isBatch, isMultiTopic, subType));
+            }
+        }
+    }
+    return values;
+}
+
+INSTANTIATE_TEST_CASE_P(Pulsar, DeadLetterQueueTest, ::testing::ValuesIn(getValues()));
```
BewareMyPower, line 388:
```suggestion
INSTANTIATE_TEST_SUITE_P(Pulsar, DeadLetterQueueTest,
                         testing::Combine(testing::Values(true, false), testing::Values(true, false),
                                          testing::Values(ConsumerType::ConsumerShared,
                                                          ConsumerType::ConsumerKeyShared)),
                         [](const testing::TestParamInfo<DeadLetterQueueTest::ParamType> &info) {
                             return "isBatch_" + std::to_string(std::get<0>(info.param)) + "_isMultiTopics_" +
                                    std::to_string(std::get<1>(info.param)) + "_subType_" +
                                    std::to_string(std::get<2>(info.param));
                         });
```

P.S. The last argument here is optional. Without it, the tests will be:

```
Pulsar/DeadLetterQueueTest.testInitSubscription/0
Pulsar/DeadLetterQueueTest.testInitSubscription/1
...
Pulsar/DeadLetterQueueTest.testInitSubscription/7
```

With it, the tests wil be:

```
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_1_isMultiTopics_1_subType_1
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_1_isMultiTopics_1_subType_3
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_1_isMultiTopics_0_subType_1
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_1_isMultiTopics_0_subType_3
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_0_isMultiTopics_1_subType_1
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_0_isMultiTopics_1_subType_3
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_0_isMultiTopics_0_subType_1
Pulsar/DeadLetterQueueTest.testInitSubscription/isBatch_0_isMultiTopics_0_subType_3
```
```

File: lib/ConsumerImpl.cc [hunk 9]
```diff
@@ -527,6 +544,10 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
                                 << startMessageId.value());
             return;
         }
+        if (redeliveryCount >= deadLetterPolicy_.getMaxRedeliverCount()) {
+            possibleSendToDeadLetterTopicMessages_.emplace(m.getMessageId(), std::vector<Message>{m});
+            increaseAvailablePermits(cnx);
```
BewareMyPower, line 549:
```suggestion
            if (redeliveryCount > deadLetterPolicy_.getMaxRedeliverCount()) {
                increaseAvailablePermits(cnx);
            }
```

See the Java implementation [here](https://github.com/apache/pulsar/pull/17287/files#diff-f6e4c1c4091aa10525f331e48e66b29f22b9f7987755c1b4fb887e24f198bed6).

And should we call `redeliverUnacknowledgedMessages` here as well?

BewareMyPower (changes requested):

lib/ConsumerImpl.cc
BewareMyPower: If the redeliver logic needs extra tests. We can add a TODO here and add the tests in another PR.

lib/ConsumerImpl.cc
```diff
@@ -1209,7 +1239,28 @@ void ConsumerImpl::redeliverUnacknowledgedMessages(const std::set<MessageId>& me
...
     }
-    redeliverMessages(messageIds);
+
+    ClientConnectionPtr cnx = getCnx().lock();
+    if (cnx) {
+        if (cnx->getServerProtocolVersion() >= proto::v2) {
+            auto needRedeliverMsgs = std::make_shared<std::set<MessageId>>();
+            auto needCallBack = std::make_shared<std::atomic<int>>(messageIds.size());
+            auto self = get_shared_this_ptr();
+            for (const auto& msgId : messageIds) {
```
BewareMyPower, line 1249: You can add a TODO to add the `MAX_REDELIVER_UNACKNOWLEDGED` support in future.

lib/ConsumerImpl.cc
```diff
@@ -1526,4 +1577,96 @@ void ConsumerImpl::cancelTimers() noexcept {
...
+            producerConfiguration.setBlockIfQueueFull(false);
+            if (!deadLetterPolicy_.getInitialSubscriptionName().empty()) {
+                producerConfiguration.setInitialSubscriptionName(
+                    deadLetterPolicy_.getInitialSubscriptionName());
+            }
+            ClientImplPtr client = client_.lock();
+            if (client) {
+                client->createProducerAsync(
+                    deadLetterPolicy_.getDeadLetterTopic(), producerConfiguration,
+                    [this](Result res, Producer producer) {
```
BewareMyPower, line 1603: We should catch `shared_from_this()` to avoid `this` pointer is expired.

lib/ConsumerImpl.cc
```diff
@@ -1526,4 +1577,96 @@ void ConsumerImpl::cancelTimers() noexcept {
...
+                            LOG_ERROR("Dead letter producer create exception with topic: "
+                                      << deadLetterPolicy_.getDeadLetterTopic() << " ex: " << res);
+                            deadLetterProducer_.reset();
+                        }
+                    });
+            } else {
+                LOG_WARN(getName() << "Client is destroyed and cannot create dead letter producer.");
+            }
+        }
+        createLock.unlock();
```
BewareMyPower, line 1616:
```suggestion
```

`createLock.unlock()` will be automatically called. And since you don't need to unlock before `createLock` goes out of scope, you can change its type to `std::lock_guard<std::mutex>` instead of the default `Lock` (`std::unique_lock`).

lib/ConsumerImpl.cc
```diff
@@ -1526,4 +1577,96 @@ void ConsumerImpl::cancelTimers() noexcept {
...
+                .setProperties(message.getProperties())
+                .setProperty(PROPERTY_ORIGIN_MESSAGE_ID, originMessageIdStr.str())
+                .setProperty(SYSTEM_PROPERTY_REAL_TOPIC, message.getTopicName());
+            if (message.hasPartitionKey()) {
+                msgBuilder.setPartitionKey(message.getPartitionKey());
+            }
+            if (message.hasOrderingKey()) {
+                msgBuilder.setOrderingKey(message.getOrderingKey());
+            }
+            producer.sendAsync(msgBuilder.build(), [self, originMessageId, cb](Result res,
```
BewareMyPower, line 1636: I prefer catching a `weak_ptr` instead of a `shared_ptr` here because the pending sends could extend the lifetime of the consumer. If the consumer closed before the send is completed, the callback should not be called.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,390 @@
...
+    auto &consumerImpl = PulsarFriend::getConsumerImpl(consumer);
+    consumerImpl.unAckedMessageTrackerPtr_.reset(
+        new UnAckedMessageTrackerEnabled(10, PulsarFriend::getClientImplPtr(client), consumerImpl));
+
+    Producer producer;
+    ASSERT_EQ(ResultOk, client.createProducer(topic, producer));
+    producer.send(MessageBuilder().setContent("msg").build());
+
+    // Wait a while, message should not be send to DLQ
+    sleep(2);
```
BewareMyPower, line 121: Are 2 seconds too long? You can use `std::chrono` to sleep N milliseconds.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,390 @@
...
+
+    Message msg;
+    ASSERT_EQ(ResultOk, consumer.receive(msg));
+    client.close();
```
BewareMyPower, line 125: I didn't get the logic here, I think even without other logics, the consumer could still receive the message sent by the producer.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,390 @@
...
+            unackTimeOut, PulsarFriend::getClientImplPtr(client), *multiConsumer));
+        multiConsumer->consumers_.forEachValue([&client, unackTimeOut](ConsumerImplPtr consumer) {
+            consumer->unAckedMessageTrackerPtr_.reset(new UnAckedMessageTrackerEnabled(
+                unackTimeOut, PulsarFriend::getClientImplPtr(client), *consumer));
+        });
+    } else {
+        auto &consumerImpl = PulsarFriend::getConsumerImpl(consumer);
+        consumerImpl.unAckedMessageTrackerPtr_.reset(new UnAckedMessageTrackerEnabled(
+            unackTimeOut, PulsarFriend::getClientImplPtr(client), consumerImpl));
+    }
```
BewareMyPower, line 208: I think a better way to set the `unackTimeout` is to modify the `impl_` field of `ConsumerConfiguration` directly, so we can pass the check in `ConsumerConfiguration::setUnAckedMessagesTimeoutMs`.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,390 @@
...
+    Message msg;
+    for (int i = 0; i < dlqPolicy.getMaxRedeliverCount() * num + num; ++i) {
+        ASSERT_EQ(ResultOk, consumer.receive(msg));
+        consumer.negativeAcknowledge(msg);
+    }
+
+    // assert dlq msg.
+    for (int i = 0; i < num; i++) {
+        ASSERT_EQ(ResultOk, deadLetterConsumer.receive(msg, 5000));
+        ASSERT_TRUE(!msg.getDataAsString().empty());
```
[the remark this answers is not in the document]
BewareMyPower: Please change other places of `ASSERT_TRUE(!condition)` to `ASSERT_FALSE(condition)` as well.

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
      */
     const BatchReceivePolicy& getBatchReceivePolicy() const;
 
+    /**
+     * Set dead letter policy for consumer
+     *
+     * By default some message will redelivery so many times possible, even to the extent that it can be never
```
Anonymitaet, line 405:
```suggestion
     * By default, some messages are redelivered many times, even to the extent that they can never be
```

Write in the simple present tense as much as possible if you are covering facts that were, are, and forever shall be true. https://docs.google.com/document/d/1lc5j4RtuLIzlEYCBo97AC8-U_3Erzs_lxpkDuseU0n4/edit#bookmark=id.e8uqh1awkcnp

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     * stop. By using dead letter mechanism messages will has the max redelivery count, when message exceeding
```
Anonymitaet, line 406:
```suggestion
     * stopped. By using the dead letter mechanism, messages have the max redelivery count, when they exceeding
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     * the maximum number of redeliveries, message will send to the Dead Letter Topic and acknowledged
```
Anonymitaet, line 407:
```suggestion
     * the maximum number of redeliveries. Messages are sent to dead letter topics and acknowledged
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     * automatic.
```
Anonymitaet, line 408:
```suggestion
     * automatically.
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     *
+     * You can enable the dead letter mechanism by setting dead letter policy.
```
Anonymitaet, line 410:
```suggestion
     * You can enable the dead letter mechanism by setting the dead letter policy.
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     * example:
```
Anonymitaet, line 411:
```suggestion
     * Example:
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     *
+     * <pre>
+     * * DeadLetterPolicy dlqPolicy = DeadLetterPolicyBuilder()
+     *                       .maxRedeliverCount(10)
+     *                       .build();
+     * </pre>
+     * Default dead letter topic name is {TopicName}-{Subscription}-DLQ.
+     * To setting a custom dead letter topic name
```
Anonymitaet, line 419:
```suggestion
     * To set a custom dead letter topic name
```

include/pulsar/ConsumerConfiguration.h
```diff
@@ -398,6 +399,42 @@ class PULSAR_PUBLIC ConsumerConfiguration {
...
+     * <pre>
+     * DeadLetterPolicy dlqPolicy = DeadLetterPolicyBuilder()
+     *                       .deadLetterTopic("dlq-topic")
+     *                       .maxRedeliverCount(10)
+     *                       .initialSubscriptionName("init-sub-name")
+     *                       .build();
+     * </pre>
+     * @param deadLetterPolicy thd default is empty
```
Anonymitaet, line 427:
```suggestion
     * @param deadLetterPolicy Default value is empty
```

include/pulsar/DeadLetterPolicyBuilder.h
```diff
@@ -0,0 +1,94 @@
...
+ * ```
+ */
+class PULSAR_PUBLIC DeadLetterPolicyBuilder {
+   public:
+    DeadLetterPolicyBuilder();
+
+    /**
+     * Set dead letter topic.
+     *
+     * @param deadLetterTopic Name of the dead topic where the failing messages will be sent.
```
Anonymitaet, line 51:
```suggestion
     * @param deadLetterTopic Name of the dead topic where the failing messages are sent.
```

include/pulsar/DeadLetterPolicyBuilder.h
```diff
@@ -0,0 +1,94 @@
...
+     * The default value is: sourceTopicName + "-" + subscriptionName + "-DLQ"
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& deadLetterTopic(const std::string& deadLetterTopic);
+
+    /**
+     * Set max redeliver count
+     *
+     * @param maxRedeliverCount Maximum number of times that a message will be redelivered before being sent
```
Anonymitaet, line 61:
```suggestion
     * @param maxRedeliverCount Maximum number of times that a message is redelivered before being sent
```

include/pulsar/DeadLetterPolicyBuilder.h
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @return
+     */
+    DeadLetterPolicyBuilder& maxRedeliverCount(int maxRedeliverCount);
+
+    /**
+     * Set initial subscription name
+     *
+     * @param initialSubscriptionName  Name of the initial subscription name of the dead letter topic.
+     * If this field is not set, the initial subscription for the dead letter topic will not be created.
```
Anonymitaet, line 74:
```suggestion
     * If this field is not set, the initial subscription for the dead letter topic is not created.
```

include/pulsar/DeadLetterPolicyBuilder.h
```diff
@@ -0,0 +1,94 @@
...
+     * If this field is set but the broker's `allowAutoSubscriptionCreation` is disabled, the DLQ producer
+     * will fail to be created.
```
Anonymitaet, line 76:
```suggestion
     * fails to be created.
```

lib/ConsumerImpl.cc
shibd: Make sense.

lib/ConsumerImpl.cc
shibd: > And should we call redeliverUnacknowledgedMessages here as well?

Yes, Need to call `redeliverUnacknowledgedMessages(const std::set<MessageId>& messageIds)`, I understand it wrong before.

The design of the DLQ goes like this:

When receiving the message `redeliveryCount == deadletterPolicy_.getMaxRedeliverCount`, We add it to `possibleSendToDeadLetterTopicMessages_` waiting for the user to call `nack method` or `ackTimeOut` trigger send to DLQ.

https://github.com/apache/pulsar-client-cpp/blob/e818ead391dc52ef08bbd59c748a00492dcf0cb5/lib/NegativeAcksTracker.cc#L82-L84

https://github.com/apache/pulsar-client-cpp/blob/1721e0005975bcc9cbd49566d6047760e6621a3b/lib/UnAckedMessageTrackerEnabled.cc#L68-L73

This can reduce once redeliver message from the server to the client.

But there are also unexpected scenarios, case `transaction.abort` : apache/pulsar/pull/17287, It does not go through the `redeliverUnacknowledgedMessages(const std::set<MessageId>& messageIds)` method, so need to directly judge again whether it is greater than: `if (redeliveryCount > deadLetterPolicy_.getMaxRedeliverCount())`

I originally thought the CPP client had no transactions, so it didn't need to handle the scenario. 

But I found that the user can call this method:

https://github.com/apache/pulsar-client-cpp/blob/2980bbe9d3ef47eef4a1d468c3898b5a989b4df0/include/pulsar/Consumer.h#L359

It also triggers the server to deliver messages directly without going through the logic of sending them to DLQ. 

I added back this judgment condition and added the test to cover it: `DeadLetterQueueTest.testSendDLQTriggerByRedeliverUnacknowledgedMessages`

tests/DeadLetterQueueTest.cc
shibd: In order cover to this: https://github.com/apache/pulsar/pull/3079

tests/DeadLetterQueueTest.cc
shibd: In this case we need to add `friend` to the `ConsumerConfiguration` include file. This may expose the test class to the user.

Is there any good way?

shibd:
@BewareMyPower @Anonymitaet Thanks for your reviews. All suggestions have been fixed. PTAL.

Anonymitaet:
Thanks! LGTM from the doc perspective.

tests/DeadLetterQueueTest.cc
BewareMyPower: The class won't be exposed. The `PulsarFriend` is already exposed in these headers.

In addition, the friend declaration does not **expose** anything. It just indicates you can implement a class with the same name. Exposing usually means the methods and fields are exposed to users.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,438 @@
...
+    client.close();
+}
+
+class DeadLetterQueueTest : public ::testing::TestWithParam<std::tuple<bool, bool, ConsumerType>> {
+   public:
+    void SetUp() override {
+        isProducerBatch_ = std::get<0>(GetParam());
+        isMultiConsumer_ = std::get<1>(GetParam());
+        consumerType_ = std::get<2>(GetParam());
+        producerConf_ = ProducerConfiguration().setBatchingEnabled(isProducerBatch_);
```
BewareMyPower, line 154: Could you also initialize the `ConsumerConfiguration` here? (this test file is too huge with 400+ lines)

include/pulsar/ProducerConfiguration.h
```diff
@@ -532,11 +532,15 @@ class PULSAR_PUBLIC ProducerConfiguration {
...
 
-    friend class PulsarWrapper;
-
    private:
-    struct Impl;
-    std::shared_ptr<ProducerConfigurationImpl> impl_;
+    typedef std::shared_ptr<ProducerConfigurationImpl> ProducerConfigurationImplPtr;
+    ProducerConfigurationImplPtr impl_;
+
+    ProducerConfiguration(ProducerConfigurationImplPtr& impl);
```
BewareMyPower, line 539: Pass a `std::shared_ptr` by const reference or value, don't pass it by non-const reference. Actually, I found this constructor is never used. And could you avoid other unrelated changes (`typedef`, moving the position of `friend`) as well?

```diff
diff --git a/include/pulsar/ProducerConfiguration.h b/include/pulsar/ProducerConfiguration.h
index dc708e9..ca11bd0 100644
--- a/include/pulsar/ProducerConfiguration.h
+++ b/include/pulsar/ProducerConfiguration.h
@@ -532,15 +532,12 @@ class PULSAR_PUBLIC ProducerConfiguration {
      */
     ProducerAccessMode getAccessMode() const;

-   private:
-    typedef std::shared_ptr<ProducerConfigurationImpl> ProducerConfigurationImplPtr;
-    ProducerConfigurationImplPtr impl_;
-
-    ProducerConfiguration(ProducerConfigurationImplPtr& impl);
-
     friend class PulsarWrapper;
-    friend class ConsumerImpl;
     friend class ProducerImpl;
+    friend class ConsumerImpl;
+
+   private:
+    std::shared_ptr<ProducerConfigurationImpl> impl_;
 };
 }  // namespace pulsar
 #endif /* PULSAR_PRODUCERCONFIGURATION_H_ */
diff --git a/lib/ProducerConfiguration.cc b/lib/ProducerConfiguration.cc
index 3c1bd24..0ee79f6 100644
--- a/lib/ProducerConfiguration.cc
+++ b/lib/ProducerConfiguration.cc
@@ -32,8 +32,6 @@ ProducerConfiguration::~ProducerConfiguration() {}

 ProducerConfiguration::ProducerConfiguration(const ProducerConfiguration& x) : impl_(x.impl_) {}

-ProducerConfiguration::ProducerConfiguration(ProducerConfigurationImplPtr& impl) : impl_(impl) {}
-
 ProducerConfiguration& ProducerConfiguration::operator=(const ProducerConfiguration& x) {
     impl_ = x.impl_;
     return *this;
```

test-conf/standalone-ssl.conf
BewareMyPower: It conflicts with https://github.com/apache/pulsar-client-cpp/pull/157, since this change is not required in this PR, I think you can revert it.

BewareMyPower (changes requested):

BewareMyPower:
BTW, let's address comments in this PR in priority to #157.

tests/DeadLetterQueueTest.cc
shibd: Yes, I refactored it, PTAL.

include/pulsar/ProducerConfiguration.h
shibd: Yes, this #157 related modifications. I reverted it.

include/pulsar/ConsumerConfiguration.h
```diff
@@ -554,6 +593,8 @@ class PULSAR_PUBLIC ConsumerConfiguration {
     bool isStartMessageIdInclusive() const;
 
     friend class PulsarWrapper;
+    friend class DeadLetterQueueTest;
+    friend class DeadLetterQueueTest_testWithoutConsumerReceiveImmediately_Test;
```
BewareMyPower, line 597: Could you replace them with `PulsarFriend` and add some methods to `PulsarFriend` to access private fields of `ConsumerConfiguration`? Once the test name changed, these friend classes would be meaningless.

BewareMyPower (changes requested):

lib/ProducerConfiguration.cc
```diff
@@ -16,6 +16,8 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+#include "pulsar/ProducerConfiguration.h"
```
BewareMyPower, line 19:
```suggestion
#include <pulsar/ProducerConfiguration.h>
```

We usually use `<>` for headers of the Pulsar client itself.

lib/ProducerConfigurationImpl.h
```diff
@@ -50,6 +50,20 @@ struct ProducerConfigurationImpl {
...
+     * If this field is not set, the initial subscription will not be created.
+     * This method is limited to internal use
+     *
+     * @param initialSubscriptionName Name of the initial subscription of the topic.
+     */
+    void setInitialSubscriptionName(const std::string& initialSubscriptionNameParam) {
+        initialSubscriptionName = initialSubscriptionNameParam;
+    }
+
+    const std::string& getInitialSubscriptionName() const { return initialSubscriptionName; }
```
BewareMyPower, line 66: It's better not to add a setter and getter to a POD class. `initialSubscriptionName` is a public field, we can modify it directly. Adding a getter and setter to a POD class is redundant and adds much more code.

lib/Commands.h
```diff
@@ -110,7 +110,8 @@ class Commands {
                                     const std::map<std::string, std::string>& metadata,
                                     const SchemaInfo& schemaInfo, uint64_t epoch,
                                     bool userProvidedProducerName, bool encrypted,
-                                    ProducerAccessMode accessMode, boost::optional<uint64_t> topicEpoch);
+                                    ProducerAccessMode accessMode, boost::optional<uint64_t> topicEpoch,
+                                    std::string initialSubscriptionName);
```
BewareMyPower, line 114: Pass a `std::string` by const reference (`const std::string&`), not value.

lib/ConsumerImpl.cc
```diff
@@ -1220,7 +1287,7 @@ void ConsumerImpl::redeliverMessages(const std::set<MessageId>& messageIds) {
             LOG_DEBUG("Sending RedeliverUnacknowledgedMessages command for Consumer - " << getConsumerId());
         }
     } else {
-        LOG_DEBUG("Connection not ready for Consumer - " << getConsumerId());
+        LOG_WARN("Connection not ready for Consumer - " << getConsumerId());
```
BewareMyPower, line 1290: Please revert this unrelated change. It should not be a warning log when a `shared_ptr` is expired.

lib/ConsumerImpl.cc
```diff
@@ -1526,4 +1593,108 @@ void ConsumerImpl::cancelTimers() noexcept {
...
+                        if (res == ResultOk) {
+                            self->deadLetterProducer_->setValue(producer);
+                        } else {
+                            LOG_ERROR("Dead letter producer create exception with topic: "
+                                      << self->deadLetterPolicy_.getDeadLetterTopic() << " ex: " << res);
+                            self->deadLetterProducer_.reset();
+                        }
+                    });
+            } else {
+                LOG_WARN(getName() << "Client is destroyed and cannot create dead letter producer.");
```
BewareMyPower, line 1628: In this case, we should skip the code below because `deadLetterProducer_->getFuture()` will never complete.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,399 @@
...
+
+class DeadLetterQueueTest : public ::testing::TestWithParam<std::tuple<bool, bool, ConsumerType>> {
+   public:
+    void SetUp() override {
+        isProducerBatch_ = std::get<0>(GetParam());
+        isMultiConsumer_ = std::get<1>(GetParam());
+        consumerType_ = std::get<2>(GetParam());
+
+        std::string testSuiteName = testing::UnitTest::GetInstance()->current_test_info()->name();
+        replace(testSuiteName.begin(), testSuiteName.end(), '/', '_');
```
BewareMyPower, line 154:
```suggestion
        std::replace(testSuiteName.begin(), testSuiteName.end(), '/', '_');
```

include/pulsar/ConsumerConfiguration.h
shibd: Yes. This way is better.

tests/DeadLetterQueueTest.cc
```diff
@@ -0,0 +1,399 @@
...
+
+    void TearDown() override { client_.close(); }
+
+   protected:
+    Client client_{lookupUrl};
+    ProducerConfiguration producerConf_;
+    ConsumerConfiguration consumerConf_;
+    bool isMultiConsumer_;
+    bool isProducerBatch_;
+    ConsumerType consumerType_;
```
BewareMyPower, line 181: Change them to local variables because they are only used in `SetUp()`.

BewareMyPower:
I have left the last comments. This PR overall LGTM. Please avoid committing such a huge patch next time. If a PR contains many code changes, it would be very hard to review all of them. Some potential issues might not be exposed.

BewareMyPower (approved):

PR #142, opened 2022-12-12
[feat] Support AUTO_PUBLISH schema.

shibd:
Master Issue: apache/pulsar/pull/2685

### Motivation

Implementing AUTO_PUBLISH schema solves the following problems:
1. If a schema exists in a topic, but the producer does not know it. He still wanted to send data to the topic. (The scenario in which the DLQ sends a message: Please refer #139 **[Verifying this change][5]**)

2. For clients based on C++ implementations(Node and Python), these client supports help users serialize and deserialize.
They can use this `AUTO_PUBLISH` schema feature to implement the same semantics as Java clients.

### Modifications
- When creating a producer with `AUTO_PUBLISH` schema, try to get the schema of that topic and use it.
- Support `getSchema` on `LookupService`(HTTP and Binary).

### Verifying this change

  - Add `LookupServiceTest` unit test to cover `getSchema` logic.
  - Add `SchemaTest.testAutoPublicSchema` unit test to cover creating producer success when the topic has a schema.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Result.h
+++ b/include/pulsar/Result.h
@@ -92,6 +92,8 @@ enum Result
     ResultMemoryBufferIsFull,  /// Client-wide memory limit has been reached
 
     ResultInterrupted,  /// Interrupted while waiting to dequeue
+
+    ResultNotFound  /// The generic was not found
 };
 
 // Return string representation of result code
--- a/include/pulsar/Schema.h
+++ b/include/pulsar/Schema.h
@@ -134,6 +134,8 @@ enum SchemaType
 // Return string representation of result code
 PULSAR_PUBLIC const char *strSchemaType(SchemaType schemaType);
 
+PULSAR_PUBLIC SchemaType enumSchemaType(std::string schemaTypeStr);
+
 class SchemaInfoImpl;
 
 typedef std::map<std::string, std::string> StringMap;
@@ -195,7 +197,6 @@ class PULSAR_PUBLIC SchemaInfo {
    private:
     typedef std::shared_ptr<SchemaInfoImpl> SchemaInfoImplPtr;
     SchemaInfoImplPtr impl_;
-    static constexpr uint32_t INVALID_SIZE = 0xFFFFFFFF;
 };
 
 }  // namespace pulsar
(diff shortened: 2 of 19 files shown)


BewareMyPower:
It looks like this PR mainly solves the issue that the producer cannot be created when `isSchemaValidationEnforced` is true. But the validation still requires the new dependencies like Avro. Otherwise, the `AUTO_PUBLISH` schema with no validation could be weird.

We should not introduce the validation logic into the C++ client. Maybe we can think of a way to support passing a customized message builder so that we can perform the validation on a generic type without exposing the internal code.

> If a schema exists in a topic, but the producer does not know it. He still wanted to send data to the topic. (The scenario in which the DLQ sends a message

Could you describe in which case could the DLQ topic have schema? I thought he topic could only be created by the client itself, right?

shibd:
> It looks like this PR mainly solves the issue that the producer cannot be created when isSchemaValidationEnforced is true. But the validation still requires the new dependencies like Avro. Otherwise, the AUTO_PUBLISH schema with no validation could be weird.

Yes, and we also need to support setting the schema when sending messages. Only then is `AUTO_PUBLISH` fully usable.

> We should not introduce the validation logic into the C++ client. Maybe we can think of a way to support passing a customized message builder so that we can perform the validation on a generic type without exposing the internal code.

I think we can open an issue or mail thread to discuss it.

> Could you describe in which case could the DLQ topic have schema? I thought he topic could only be created by the client itself, right?

Yes, When the user creates a dead-letter topic by himself, and the schema of the dead-letter topic is not compatible from the schema of the source topic, the DLQ producer will create failed.

BTW: This scenario is relatively rare, and I don't think it affects the release of DLQ features.


I will close this PR first.

shibd:
> Yes, When the user creates a dead-letter topic by himself, and the schema of the dead-letter topic is not compatible from the schema of the source topic, the DLQ producer will create failed.

If a topic has a schema, and the consumer uses AUTO_CONSUME schema, and DLQTopic also has the same schema. In this case, the DLQ producer fails to create(IncompatibleSchema). Because DLQ producers created in consumers do not have a schema.

PR #146, opened 2022-12-15
[fix] Use authoritative argument correctly in BinaryProtoLookupService::findBroker

erobot:
### Motivation

Use authoritative argument correctly in BinaryProtoLookupService::findBroker.

In the current code, authoritative field of lookup request is always false. When a bundle is not loaded by any broker, a lookup request with authoritative as false will not trigger broker to load the bundle, and lookup requests will loop between leader broker and the chosen broker. In this case, the lookup will not succeed and result in a large number of lookup requests to the broker as current c++ client do not have lookup redirect limit. If broker return failure after a large number of redirects, chains 
 of future will result in a large depth of function calls and may cause the stack to overflow.

https://github.com/apache/pulsar-client-cpp/blob/44f3b5240732a035395ec34b977301435dd0ec71/lib/BinaryProtoLookupService.cc#L55

### Modifications

Use authoritative argument.

### Verifying this change

This change is a trivial fix and not easy to test in standalone (need 2 or more broker).

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
bug fix only

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/BinaryProtoLookupService.cc
+++ b/lib/BinaryProtoLookupService.cc
@@ -38,7 +38,7 @@ auto BinaryProtoLookupService::findBroker(const std::string& address, bool autho
     LOG_DEBUG("find broker from " << address << ", authoritative: " << authoritative << ", topic: " << topic);
     auto promise = std::make_shared<Promise<Result, LookupResult>>();
     // NOTE: we can use move capture for topic since C++14
-    cnxPool_.getConnectionAsync(address).addListener([this, promise, topic, address](
+    cnxPool_.getConnectionAsync(address).addListener([this, promise, topic, address, authoritative](
                                                          Result result,
                                                          const ClientConnectionWeakPtr& weakCnx) {
         if (result != ResultOk) {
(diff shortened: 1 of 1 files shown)


shibd:
> When a bundle is not loaded by any broker, a lookup request with authoritative as false will not trigger broker to load the bundle.

@erobot Hi, Thanks for your pr. I don't quite understand this paragraph. According to the following code, when `authoritative` is false, an attempt is made to acquire new bundle ownership.

https://github.com/apache/pulsar/blob/b669acc2dd341fd6281ee63dffa2554f4918f9a4/pulsar-broker/src/main/java/org/apache/pulsar/broker/namespace/NamespaceService.java#L485-L523

BewareMyPower (approved):

tisonkun (approved):

BewareMyPower:
To avoid the regression, it might be better to add tests based on a multi-brokers cluster. But it could take some time to work. Let's merge this PR first.

PR #148, opened 2022-12-16
[feat] PIP 107: Introduce chunk message ID

### LINKED ISSUE #79: [Feature] Support chunked message id
## Motivation

This is the feature for the implementation of https://github.com/apache/pulsar/issues/12402 at the cpp client side.

RobertIndie:
Fixes #79

### Motivation

This is the C++ implementation for https://github.com/apache/pulsar/issues/12402

### Modifications

* Add ChunkMessageIdImpl
* Return ChunkMessageId when the Producer produces the chunk message or when the consumer consumes the chunk message.
* In cosumer.seek, use the first chunk message-id of the chunk message-id. This will solve the problem caused by seeking chunk messages. This is also the impact of this PIP on the original business logic.

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -188,8 +188,7 @@ class PULSAR_PUBLIC Message {
     MessageImplPtr impl_;
 
     Message(MessageImplPtr& impl);
-    Message(const proto::CommandMessage& msg, proto::MessageMetadata& data, SharedBuffer& payload,
-            int32_t partition);
+    Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload);
     /// Used for Batch Messages
     Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload,
             proto::SingleMessageMetadata& singleMetadata, const std::string& topicName);
--- a/lib/ChunkMessageIdImpl.h
+++ b/lib/ChunkMessageIdImpl.h
@@ -19,12 +19,14 @@
 
 #pragma once
 
+#include <pulsar/MessageId.h>
+
 #include "MessageIdImpl.h"
 
 namespace pulsar {
 class ChunkMessageIdImpl;
 typedef std::shared_ptr<ChunkMessageIdImpl> ChunkMessageIdImplPtr;
-class ChunkMessageIdImpl : public MessageIdImpl {
+class ChunkMessageIdImpl : public MessageIdImpl, public std::enable_shared_from_this<ChunkMessageIdImpl> {
    public:
     ChunkMessageIdImpl() : firstChunkMsgId_(std::make_shared<MessageIdImpl>()) {}
 
@@ -36,9 +38,9 @@ class ChunkMessageIdImpl : public MessageIdImpl {
         this->partition_ = msgId.partition();
     }
 
-    std::shared_ptr<MessageIdImpl> getFirstChunkMessageId() { return firstChunkMsgId_; }
+    std::shared_ptr<const MessageIdImpl> getFirstChunkMessageId() const { return firstChunkMsgId_; }
 
-    static MessageId buildMessageId(const ChunkMessageIdImplPtr& msgIdImpl) { return MessageId{msgIdImpl}; }
+    MessageId build() { return MessageId{std::dynamic_pointer_cast<MessageIdImpl>(shared_from_this())}; }
 
    private:
     std::shared_ptr<MessageIdImpl> firstChunkMsgId_;
--- a/lib/ChunkMessageIdImpl.h
+++ b/lib/ChunkMessageIdImpl.h
@@ -36,9 +36,9 @@ class ChunkMessageIdImpl : public MessageIdImpl {
         this->partition_ = msgId.partition();
     }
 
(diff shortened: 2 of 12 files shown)


include/pulsar/Message.h
```diff
@@ -188,8 +188,7 @@ class PULSAR_PUBLIC Message {
     MessageImplPtr impl_;
 
     Message(MessageImplPtr& impl);
-    Message(const proto::CommandMessage& msg, proto::MessageMetadata& data, SharedBuffer& payload,
-            int32_t partition);
+    Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload);
```
BewareMyPower, line 191: Just have a thought, not a requested change and this PR needs no change about this comment. I think the constructor of `Message` should have only one private constructor.

```c++
Message(const MessageImplPtr& impl);
```

The other constructors should be moved to `MessageImpl`. After that, there is no need to expose the classes (even if as the forward declarations) in the header.

lib/ChunkMessageIdImpl.h
```diff
@@ -0,0 +1,46 @@
...
+
+    void setFirstChunkMessageId(const MessageId& msgId) { *firstChunkMsgId_ = *msgId.impl_; }
+
+    void setLastChunkMessageId(const MessageId& msgId) {
+        this->ledgerId_ = msgId.ledgerId();
+        this->entryId_ = msgId.entryId();
+        this->partition_ = msgId.partition();
+    }
+
+    std::shared_ptr<MessageIdImpl> getFirstChunkMessageId() { return firstChunkMsgId_; }
```
BewareMyPower, line 39:
```suggestion
    std::shared_ptr<MessageIdImpl> getFirstChunkMessageId() const { return firstChunkMsgId_; }
```

lib/ChunkMessageIdImpl.h
```diff
@@ -0,0 +1,46 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include "MessageIdImpl.h"
```
BewareMyPower, line 22:
```suggestion
#include <pulsar/MessageId.h>

#include "MessageIdImpl.h"
```

The `MessageIdImpl.h` only includes the `<cstdint>` and `<string>` headers. Thought it doesn't cause the compilation failure because the files that include `ChunkMessageIdImpl.h` all have other necessary headers (like `<memory>` for `std::shared_ptr`) included, but there is a possibility that a new file that includes this header doesn't include these required headers.

It could also make same static syntax analysis tool fail. See my Vim:

image

lib/ChunkMessageIdImpl.h
```diff
@@ -0,0 +1,46 @@
...
+
+    void setLastChunkMessageId(const MessageId& msgId) {
+        this->ledgerId_ = msgId.ledgerId();
+        this->entryId_ = msgId.entryId();
+        this->partition_ = msgId.partition();
+    }
+
+    std::shared_ptr<MessageIdImpl> getFirstChunkMessageId() { return firstChunkMsgId_; }
+
+    static MessageId buildMessageId(const ChunkMessageIdImplPtr& msgIdImpl) { return MessageId{msgIdImpl}; }
```
BewareMyPower, line 41: Instead of the static method, we can just add a non-static method like:

```c++
    MessageId build() const {
        return MessageId{std::dynamic_pointer_cast<MessageIdImpl>(
            std::const_pointer_cast<ChunkMessageIdImpl>(shared_from_this()))};
    }
```

To have `shared_from_this()` method, you should inherit another class:

```c++
class ChunkMessageIdImpl : public MessageIdImpl, public std::enable_shared_from_this<ChunkMessageIdImpl> {
```

Then we can just use `chunkedMsgId->build()` instead of `ChunkedMessageIdImpl::buildMessage(chunkedMsgId);`

lib/ChunkMessageIdImpl.h
RobertIndie: Thanks. I have also changed the return type to `std::shared_ptr<const MessageIdImpl>`. Otherwise, even if we mark the method as const but the caller could also change the MessageIdImpl that the pointer point to and therefore change the ChunkMessageIdImpl which should be const.

lib/ChunkMessageIdImpl.h
RobertIndie: Make sense to me. Fixed it.

lib/ChunkMessageIdImpl.h
RobertIndie: Good idea. Fixed it. Note that I have removed `const_pointer_cast ` because we don't need it in this approach.

include/pulsar/Message.h
RobertIndie: Agree. We should make the constructor clearer and make this existing implementation out of the constructor.

BewareMyPower (approved):

PR #149, opened 2022-12-16
[feat] Support messages with generic types

BewareMyPower:
### Motivation

Pulsar C++ client doesn't support schema yet. It only supports configuring `SchemaInfo` when creating producers or consumers. The main reason is that C++'s templates are processed at the time of compilation. Templatizing `Producer`, `Consumer` or `Message` could expose all internal code.

Currently, users might write the following code for serialization and deserialization:

```c++
producer.send(MessageBuilder().setContent(encode(value)).build());
```

```c++
Message msg;
consumer.receive(msg);
auto value = decode(msg.getData(), msg.getSize());
```

However, the `encode` and `decode` functions are just possible solutions from users, they might use some other interfaces like a class with two virtual methods. There is no way to provide a common interface for serialization and deserialization.

### Modifications

Add a `TypedMessageBuilder<T>` class template that accepts an encoder function and a validation function. The validation function is used when users want to simulate the Java client's `AUTO_PRODUCE` schema. Define a full specialization for `std::string` template argument to avoid encoding. Since it inherits the `MessageBuilder`, it's compatible with the current code style:

```c++
// It should be noted you have to call `setValue` before methods in base class
auto msg = TypedMessageBuilder<T>(encoder).setValue(value).setPartitionKey(key).build();
```

Add a `TypedMessage<T>` class that only adds a decoder to the `Message` instance and can be converted from the `Message` directly:

```c++
auto typedMsg = TypedMessage<T>(msg, decoder);
std::cout << typedMsg.getValue() << std::endl;  // decode the bytes
std::cout << typedMsg.getMessageId() << std::endl;  // call methods from Message
```

For convenience, the following APIs are added to `Consumer`:

```c++
template <typename T>
Result receive(TypedMessage<T>& msg, typename TypedMessage<T>::Decoder decoder);

template <typename T>
Result receive(TypedMessage<T>& msg, int timeoutMs, typename TypedMessage<T>::Decoder decoder);

template <typename T>
void receiveAsync(std::function<void(Result result, const TypedMessage<T>&)> callback,
                  typename TypedMessage<T>::Decoder decoder)
```

The `ConsumerConfiguration` can configure a listener that accepts a
`TypedMessage<T>` now:

```c++
template <typename T>
ConsumerConfiguration& setTypedMessageListener(
        std::function<void(Consumer&, const TypedMessage<T>&)> listener,
        typename TypedMessage<T>::Decoder decoder);
```

Since it calls the original listener and `Consumer` is only forward declared in `ConsumerConfiguration.h`, the 1st argument is changed from `Consumer` to `Consumer&`. It's an API change but it's backward compatible because `std::function<void(Consumer, ...)>` can be cast to `std::function<Consumer&, ...)>` implicitly.

Based on these API changes, we can write a separated header-only C++ library as the schema extension.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/TypedMessageTest.cc
+++ b/tests/TypedMessageTest.cc
@@ -1,0 +1,136 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include <gtest/gtest.h>
+#include <pulsar/Client.h>
+#include <pulsar/MessageBuilder.h>
+#include <pulsar/TypedMessage.h>
+#include <pulsar/TypedMessageBuilder.h>
+
+#include <mutex>
+#include <vector>
+
+#include "WaitUtils.h"
+#include "lib/Latch.h"
+
+using namespace pulsar;
+
+static std::string lookupUrl = "pulsar://localhost:6650";
+
+extern std::string unique_str();
+
+using IntMessageBuilder = TypedMessageBuilder<int>;
+static auto intEncoder = [](int x) { return std::to_string(x); };
+static auto intDecoder = [](const char* data, size_t size) { return std::stoi(std::string(data, size)); };
+
+TEST(TypedMessageTest, testReceive) {
+    const auto topic = "typed-message-test-receive-" + unique_str();
+    Client client(lookupUrl);
+    Producer producer;
+    ASSERT_EQ(ResultOk, client.createProducer(topic, producer));
+    Consumer consumer;
+    ASSERT_EQ(ResultOk, client.subscribe(topic, "sub", consumer));
+
+    std::vector<MessageId> messageIds;
+    constexpr int numMessages = 100;
+
+    for (int i = 0; i < numMessages; i++) {
+        MessageId messageId;
+        auto msg = TypedMessageBuilder<int>{intEncoder}.setValue(i).build();
+        ASSERT_EQ(ResultOk, producer.send(msg, messageId));
+        messageIds.emplace_back(messageId);
+    }
+    for (int i = 0; i < numMessages; i++) {
+        TypedMessage<int> msg;
+        // ensure the thread safety for `msg`, which could be modified in the callback of `receiveAsync`
+        std::mutex msgMutex;
+        if (i % 3 == 0) {
+            ASSERT_EQ(ResultOk, consumer.receive(msg, intDecoder));
+        } else if (i % 3 == 1) {
+            ASSERT_EQ(ResultOk, consumer.receive(msg, 3000, intDecoder));
+        } else {
+            Latch latch{1};
+            consumer.receiveAsync(
+                [&latch, &msg, &msgMutex](Result result, const TypedMessage<int>& receivedMsg) {
+                    ASSERT_EQ(ResultOk, result);
+                    {
+                        std::lock_guard<std::mutex> lock{msgMutex};
+                        msg = TypedMessage<int>{receivedMsg, intDecoder};
+                    }
+                    latch.countdown();
+                });
+            ASSERT_TRUE(latch.wait(std::chrono::seconds(1)));
+        }
+
+        std::lock_guard<std::mutex> lock{msgMutex};
+        ASSERT_EQ(msg.getValue(), i);
+        ASSERT_EQ(msg.getMessageId(), messageIds[i]);
+    }
+
+    client.close();
+}
+
+TEST(TypedMessageTest, testListener) {
+    const auto topic = "typed-message-test-listener-" + unique_str();
+    Client client(lookupUrl);
+
+    ConsumerConfiguration conf;
+    std::vector<int> values;
+    std::mutex valuesMutex;
+    conf.setTypedMessageListener<int>(
+        [&values, &valuesMutex](Consumer& consumer, const TypedMessage<int>& msg) {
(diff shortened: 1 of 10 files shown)


tests/TypedMessageTest.cc
```diff
@@ -0,0 +1,77 @@
...
+    Consumer consumer;
+    ASSERT_EQ(ResultOk, client.subscribe(topic, "sub", ConsumerConfiguration().setSchema(schema), consumer));
+
+    std::vector<MessageId> messageIds;
+    constexpr int numMessages = 100;
+
+    IntSchema intSchema{schema};
+    for (int i = 0; i < numMessages; i++) {
+        MessageId messageId;
+        ASSERT_EQ(ResultOk, producer.send(intSchema.newMessage(i).build(), messageId));
```
shibd, line 52: Using primitive types doesn't seem to show the benefits of this approach. Can we provide more examples for record types?

tests/IntSchema.h
```diff
@@ -0,0 +1,63 @@
...
+#include <pulsar/TypedMessage.h>
+#include <pulsar/TypedMessageBuilder.h>
+
+#include <functional>
+#include <stdexcept>
+#include <string>
+
+namespace pulsar {
+
+class IntSchema {
```
shibd, line 32: For users, should we try to abstract an interface to the user?

tests/TypedMessageTest.cc
BewareMyPower: It's a test for `encoder` and `decoder`, not an example for complicated cases. For example, the following implementations are also okay.

```c++
    std::function<std::string(int)> encoder() { return ""; }
    std::function<int(const char*, size_t)> decoder() { return 0; }
```

BewareMyPower: I will redesign the tests soon. The consumer's API might also change.

BewareMyPower:
Hi @shibd, I changed the tests and APIs in the latest code, PTAL again.

BewareMyPower:

shibd:
@BewareMyPower Thanks for your PR, This PR can make it easier for users to serialize and deserialize messages, and it makes a lot of sense.

I'm trying to diverge my ideas. Can I use it in combination with `SchemaInfo`? This ensures that the schema and message are consistent.

Take the Producer side, for example:


1. Let `TypedMessageBuilder` provide `SchemaInfo`.

``` c++
template <typename T>
class TypedMessageBuilder : public MessageBuilder {
    // ... Omit existing code

    SchemaInfo getSchemaInfo() {
        // Two methods to generate:
        // 1. Generated according to Template T?
        // 2. The user passes it in by himself?
    }

   private:
    const Encoder encoder_;
    const Validator validator_;
};
```


2. Let `Producer` support template and cache `TypedMessageBuilder`


(1) set TypedMessageBuilder when creating a producer.

``` c++

auto typeMsg = TypedMessageBuilder<int>{intEncoder}
// The SchemaInfo will be obtained and set according to TypedMessageBuilder
client.createProducer(topic, typeMsg, producer);

```

(2) The `Producer` add support send T value method.

``` c++
class Producer {
   Result send(const T& msg) {
       Message msg = typeMessageBuilder.value(msg);
       return  send(msg);
  }
}

```


Complete users-side code

``` c++
auto typeMsg = TypedMessageBuilder<int>{intEncoder}
client.createProducer(topic, typeMsg, producer);
producer.send(100);
```

BewareMyPower:
@shibd See the point I mentioned in this PR:

> Templatizing Producer, Consumer or Message could expose all internal code.

You can try implementing the templated producer yourself. Then you might know better what I meant.

BTW, `std::function` is flexible enough. The `SchemaInfo` can be catched in a `std::function`, for example, assuming here is a `StudentSchema` class to serialize a `Student` class:

```c++
struct Student {
    int age;
    std::string name;
};

class StudentSchema {
   public:
    StudentSchema() {}

    SchemaInfo getSchemaInfo() const { return schemaInfo_; }

    std::string operator()(const Student& student) const {
        // TODO: serialize student using the schemaInfo_ and Avro C++ client
        return R"("age": )" + std::to_string(student.age) + R"(, "name": ")" + student.name + R"(")";
    }

   private:
    SchemaInfo schemaInfo_{SchemaType::AVRO, "<name>", "<avro-string>"};
};
```

Then, users can write a producer like:

```c++
    StudentSchema schema;
    Client client(lookupUrl);
    Producer producer;
    ProducerConfiguration conf;
    conf.setSchema(schema.getSchemaInfo());
    client.createProducer("my-topic", conf, producer);
    // NOTE: schema is a StudentSchema, which could be treated as
    // a std::function<std::string(const Student&)>
    producer.send(TypedMessageBuilder<Student>{schema}.setValue(Student{10, "xyz"}).build());
```

BewareMyPower:
@shibd See the point I mentioned in this PR:

> Templatizing Producer, Consumer or Message could expose all internal code.

You can try implementing the templated producer yourself. Then you might know better what I meant.

BTW, `std::function` is flexible enough. The `SchemaInfo` can be catched in a `std::function`, for example, assuming here is a `StudentSchema` class to serialize a `Student` class:

```c++
struct Student {
    int age;
    std::string name;
};

class StudentSchema {
   public:
    StudentSchema() {}

    SchemaInfo getSchemaInfo() const { return schemaInfo_; }

    std::string operator()(const Student& student) const {
        // TODO: serialize student using the schemaInfo_ and Avro C++ client
        return R"("age": )" + std::to_string(student.age) + R"(, "name": ")" + student.name + R"(")";
    }

   private:
    SchemaInfo schemaInfo_{SchemaType::AVRO, "<name>", "<avro-string>"};
};
```

Then, users can write a producer like:

```c++
    StudentSchema schema;
    Client client(lookupUrl);
    Producer producer;
    ProducerConfiguration conf;
    conf.setSchema(schema.getSchemaInfo());
    client.createProducer("my-topic", conf, producer);
    // NOTE: schema is a StudentSchema, which could be treated as
    // a std::function<std::string(const Student&)>
    producer.send(TypedMessageBuilder<Student>{schema}.setValue(Student{10, "xyz"}).build());
```

BewareMyPower:
The `SchemaInfo` can be passed by users or generated from `T` (if the Avro C++ client supports it). That's another topic.

shibd:
> BTW, std::function is flexible enough. The SchemaInfo can be catched in a std::function, for example, assuming here is a StudentSchema class to serialize a Student class:

This is a good example, it shows how to use it with `SchemaInfo`. Thanks for your explain. We can add more similar examples to examples later.

shibd (approved):

Demogorgon314 (approved):

PR #155, opened 2022-12-23
[fix][build] Hide non-exported symbols from the dependencies

BewareMyPower:
### Motivation

Currently the released libraries don't hide the symbols from the dependencies, it can be verified by the following steps on Ubuntu:

```bash
curl -O -L https://archive.apache.org/dist/pulsar/pulsar-client-cpp-3.1.0/deb-x86_64/apache-pulsar-client.deb
apt install ./apache-pulsar-client.deb
nm -D /usr/lib/libpulsar.so | grep curl
```

You will see lots of symbols from libcurl are included in `libpulsar.so`:

```
0000000000709f50 T curl_easy_cleanup
000000000070a000 T curl_easy_duphandle
...
```

The root cause is that `-Wl,--exclude-libs,ALL` is added as the compile option, but it should work as a link option.

### Modifications

Use `add_link_options` to add `-Wl,--exclude-libs=ALL` as the link option. It seems that `=ALL` should be correct but `,ALL` also works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@
 # under the License.
 #
 
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.13)
 
 project (pulsar-cpp)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake_modules")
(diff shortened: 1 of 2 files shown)


BewareMyPower:
I tried to build the x64 deb package in my local env, it does not include the symbols from libcurl or openssl.

```bash
$ nm -D /usr/lib/libpulsar.so | grep curl
0000000000bab180 B _ZN6pulsar9TopicName15curlHandleMutexE
0000000000bab1a8 B _ZN6pulsar9TopicName4curlE
$ nm -D /usr/lib/libpulsar.so | grep SSL
```

shibd (approved):

tisonkun (approved):

PR #157, opened 2022-12-26
[feat] Support auto download schema when create producer.

shibd:
### Motivation
 If a schema exists in a topic, but the producer does not know it. He still wanted to send data to the topic. (The scenario in which the DLQ sends a message: Please refer #139 **[Verifying this change][5]**)

### Modifications
- When creating a producer with `autoDownloadSchema` param, try to get the schema of that topic and use it.
- Support `getSchema` on `LookupService`(HTTP and Binary).

### Verifying this change

  - Add `LookupServiceTest` unit test to cover `getSchema` logic.
  - Add `SchemaTest.testAutoDownloadSchema` unit test to cover creating producer success when the topic has a schema.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Result.h
+++ b/include/pulsar/Result.h
@@ -94,6 +94,7 @@ enum Result
     ResultInterrupted,  /// Interrupted while waiting to dequeue
 
     ResultDisconnected,  /// Client connection has been disconnected
+    ResultNotFound       /// The generic was not found
 };
 
 // Return string representation of result code
--- a/include/pulsar/Result.h
+++ b/include/pulsar/Result.h
@@ -94,7 +94,6 @@ enum Result
     ResultInterrupted,  /// Interrupted while waiting to dequeue
 
     ResultDisconnected,  /// Client connection has been disconnected
-    ResultNotFound       /// The generic was not found
 };
 
 // Return string representation of result code
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -1308,6 +1308,52 @@ void ClientConnection::handleIncomingCommand(BaseCommand& incomingCmd) {
                     break;
                 }
 
+                case BaseCommand::GET_SCHEMA_RESPONSE: {
+                    const auto& response = incomingCmd.getschemaresponse();
+                    LOG_DEBUG(cnxString_ << "Received GetSchemaResponse from server. req_id: "
+                                         << response.request_id());
+                    Lock lock(mutex_);
+                    PendingGetSchemaMap::iterator it = pendingGetSchemaRequests_.find(response.request_id());
+                    if (it != pendingGetSchemaRequests_.end()) {
+                        Promise<Result, boost::optional<SchemaInfo>> getSchemaPromise = it->second;
+                        pendingGetSchemaRequests_.erase(it);
+                        lock.unlock();
+
+                        if (response.has_error_code()) {
+                            if (response.error_code() == proto::TopicNotFound) {
+                                getSchemaPromise.setValue(boost::none);
+                            } else {
+                                Result result = getResult(response.error_code(), response.error_message());
+                                LOG_WARN(cnxString_ << "Received error GetSchemaResponse from server "
+                                                    << result
+                                                    << (response.has_error_message()
+                                                            ? (" (" + response.error_message() + ")")
+                                                            : "")
+                                                    << " -- req_id: " << response.request_id());
+                                getSchemaPromise.setFailed(result);
+                            }
+                            return;
+                        }
+
+                        auto schema = response.schema();
+                        auto properMap = schema.properties();
+                        StringMap properties;
+                        for (auto kv = properMap.begin(); kv != properMap.end(); ++kv) {
+                            properties[kv->key()] = kv->value();
+                        }
+                        SchemaInfo schemaInfo(static_cast<SchemaType>(schema.type()), "",
+                                              schema.schema_data(), properties);
+                        getSchemaPromise.setValue(schemaInfo);
+                    } else {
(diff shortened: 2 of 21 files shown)


include/pulsar/Result.h
```diff
@@ -94,6 +94,7 @@ enum Result
     ResultInterrupted,  /// Interrupted while waiting to dequeue
 
     ResultDisconnected,  /// Client connection has been disconnected
+    ResultNotFound       /// The generic was not found
```
BewareMyPower, line 97: What does "generic" mean?

BewareMyPower: It seems to be "The schema was not found". If yes, it's better to use `ResultSchemaNotFound` for it.

lib/ClientConnection.cc
```diff
@@ -1308,6 +1308,52 @@ void ClientConnection::handleIncomingCommand(BaseCommand& incomingCmd) {
                     break;
                 }
 
+                case BaseCommand::GET_SCHEMA_RESPONSE: {
+                    const auto& response = incomingCmd.getschemaresponse();
+                    LOG_DEBUG(cnxString_ << "Received GetSchemaResponse from server. req_id: "
+                                         << response.request_id());
+                    Lock lock(mutex_);
+                    PendingGetSchemaMap::iterator it = pendingGetSchemaRequests_.find(response.request_id());
```
BewareMyPower, line 1316:
```suggestion
                    auto it = pendingGetSchemaRequests_.find(response.request_id());
```

lib/ClientConnection.cc
```diff
@@ -1308,6 +1308,52 @@ void ClientConnection::handleIncomingCommand(BaseCommand& incomingCmd) {
...
+                                                            ? (" (" + response.error_message() + ")")
+                                                            : "")
+                                                    << " -- req_id: " << response.request_id());
+                                getSchemaPromise.setFailed(result);
+                            }
+                            return;
+                        }
+
+                        auto schema = response.schema();
+                        auto properMap = schema.properties();
```
BewareMyPower, line 1339:
```suggestion
                        const auto& schema = response.schema();
                        const auto& properMap = schema.properties();
```

Avoid unnecessary copy.

lib/LookupService.h
```diff
@@ -21,6 +21,7 @@
 
 #include <pulsar/Result.h>
 
+#include <boost/optional.hpp>
 #include <memory>
```
BewareMyPower, line 25: It's better to include `<pulsar/Schema.h>` here though it does not affect the compilation because the `.cc` files that includes `LookupService.h` already include `Schema.h`. But for this single file, it might make some static code analysis checks fail. 

image

tests/LookupServiceTest.cc
```diff
@@ -274,3 +277,82 @@ TEST(LookupServiceTest, testTimeout) {
...
+
+    auto clientImplPtr = PulsarFriend::getClientImplPtr(client_);
+    auto lookup = clientImplPtr->getLookup();
+
+    boost::optional<SchemaInfo> schemaInfo;
+    auto future = lookup->getSchema(TopicName::get(topic));
+    ASSERT_EQ(ResultOk, future.get(schemaInfo));
+    ASSERT_EQ(keyValueSchema.getSchema(), schemaInfo->getSchema());
+    ASSERT_EQ(SchemaType::KEY_VALUE, schemaInfo->getSchemaType());
+    ASSERT_TRUE(!schemaInfo->getProperties().empty());
```
BewareMyPower, line 355: Replace with `ASSERT_FALSE`

shibd:
Job failure seems to have a flaky test, can help review and rerun the job? @BewareMyPower

include/pulsar/Result.h
BewareMyPower: Could this `Result` be returned with binary protocol? If it's an error that is only related to the HTTP protocol, I think it would not be a good idea to expose it as a `Result`.

tests/ClientTest.cc
```diff
@@ -305,6 +305,7 @@ TEST(ClientTest, testCloseClient) {
         auto t0 = std::chrono::steady_clock::now();
         while ((std::chrono::steady_clock::now() - t0) < std::chrono::microseconds(i)) {
         }
+        sleep(1);
```
BewareMyPower, line 308: Please move it into another PR if you want to fix the flaky test. BTW, sleeping 1 second is unacceptable here because it's in a 1000 times loop

BewareMyPower (changes requested):

tests/ClientTest.cc
shibd: Oh, sorry. This was written by my local test and I submitted it by mistake.

lib/HTTPLookupService.cc
```diff
@@ -409,4 +430,72 @@ void HTTPLookupService::handleLookupHTTPRequest(LookupPromise promise, const std
     }
 }
 
+void HTTPLookupService::handleGetSchemaHTTPRequest(GetSchemaPromise promise, const std::string completeUrl) {
+    std::string responseData;
+    Result result = sendHTTPRequest(completeUrl, responseData);
+
+    if (result == ResultNotFound) {
+        promise.setValue(boost::none);
```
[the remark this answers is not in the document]
shibd: Get it. I add a overload methods: `Result sendHTTPRequest(std::string completeUrl, std::string& responseData, long& responseCode);`, PTAL.

BewareMyPower (approved):

Issue #137, opened 2022-12-05
Support setting log level for C Api


RobertIndie:
Currently, the log level for C API is hardcoded as `INFO`:
https://github.com/apache/pulsar-client-cpp/blob/7a7b3aa5faa802fa354d2b2e41d38ce22241d700/lib/c/c_ClientConfiguration.cc#L81

Therefore, we cannot get the debug log in the C API or Nodejs client. It's better to support setting the log level for PulsarCLogger.

PR #158, opened 2022-12-27
The C API supports setting the log level.

shibd:
### Motivation
#137

### Modifications
-  Add interface `pulsar_client_configuration_set_logger_and_level` to support set the log level.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/examples/SampleCustomLoggerCApi.c
+++ b/examples/SampleCustomLoggerCApi.c
@@ -1,0 +1,89 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include <pulsar/c/client.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+
+void format_time(char *output){
+    time_t rawtime;
+    struct tm * timeinfo;
+
+    time(&rawtime);
+    timeinfo = localtime(&rawtime);
+
+    sprintf(output, "%d %d %d %d:%d:%d",
+            timeinfo->tm_year + 1900, timeinfo->tm_mon + 1,  timeinfo->tm_mday,
+            timeinfo->tm_hour, timeinfo->tm_min, timeinfo->tm_sec);
+}
+
+void custom_logger(pulsar_logger_level_t level, const char *file, int line, const char *message,
+              void *ctx) {
+    time_t mytime = time(NULL);
+    char * time_str = ctime(&mytime);
+    // Control the log level yourself.
+    if (level >= pulsar_DEBUG) {
+        format_time(time_str);
+        printf("[%s] [%u] [%s] [%d] [%s] \n", time_str, level, file, line, message);
+    }
+}
+
+int main() {
+    pulsar_client_configuration_t *conf = pulsar_client_configuration_create();
+
+    pulsar_client_configuration_set_logger(conf, custom_logger, NULL);
+    pulsar_client_configuration_set_memory_limit(conf, 64 * 1024 * 1024);
+    pulsar_client_t *client = pulsar_client_create("pulsar://localhost:6650", conf);
+
+    pulsar_producer_configuration_t* producer_conf = pulsar_producer_configuration_create();
+    pulsar_producer_configuration_set_batching_enabled(producer_conf, 1);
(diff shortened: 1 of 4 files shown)


BewareMyPower:
Not related to this PR, I found the `.c` file is not formatted by `clang-format`. I think you can fix the code style first. The `clang-format` check could be fixed in another PR.

lib/c/c_ClientConfiguration.cc
```diff
@@ -78,7 +78,7 @@ class PulsarCLogger : public pulsar::Logger {
     PulsarCLogger(const std::string &file, pulsar_logger logger, void *ctx)
         : file_(file), logger_(logger), ctx_(ctx) {}
 
-    bool isEnabled(Level level) { return level >= pulsar::Logger::LEVEL_INFO; }
+    bool isEnabled(Level level) { return level >= pulsar::Logger::LEVEL_DEBUG; }
```
BewareMyPower, line 81: We can just return true here.

examples/SampleCustomLoggerCApi.c
```diff
@@ -0,0 +1,89 @@
...
+
+void custom_logger(pulsar_logger_level_t level, const char *file, int line, const char *message,
+              void *ctx) {
+    time_t mytime = time(NULL);
+    char * time_str = ctime(&mytime);
+    // Control the log level yourself.
+    if (level >= pulsar_DEBUG) {
+        format_time(time_str);
+        printf("[%s] [%u] [%s] [%d] [%s] \n", time_str, level, file, line, message);
+    }
```
BewareMyPower, line 45: Though it's an example, we should check `level` first to avoid `time` and `ctime` calls.

shibd:
> Not related to this PR, I found the `.c` file is not formatted by `clang-format`. I think you can fix the code style first. The `clang-format` check could be fixed in another PR.

Thanks for the reminder. I open a new PR #159 fix it.

lib/c/c_ClientConfiguration.cc
```diff
@@ -78,7 +78,7 @@ class PulsarCLogger : public pulsar::Logger {
     PulsarCLogger(const std::string &file, pulsar_logger logger, void *ctx)
         : file_(file), logger_(logger), ctx_(ctx) {}
 
-    bool isEnabled(Level level) { return level >= pulsar::Logger::LEVEL_INFO; }
+    bool isEnabled(Level level) { return true; }
```
RobertIndie, line 81: Could we support setting the level in the PulsarCLogger. Like SimpleLogger here: https://github.com/apache/pulsar-client-cpp/blob/5d5ed2acc2e89aed90b593343369f360644440e7/lib/SimpleLogger.h#L53
Otherwise, all debug output-related calculations will be called, which may have some impact on the performance.

shibd: For C users, We expose a method: `pulsar_logger `.

https://github.com/apache/pulsar-client-cpp/blob/0e86e9f1f8304fd0bbf21185c1c66d0b5ac58482/include/pulsar/c/client_configuration.h#L36-L37

`True` is always returned here in order to let user controls what level of the log is printed in this method.

Such as:

``` c
void custom_logger(pulsar_logger_level_t level, const char *file, int line, const char *message,
              void *ctx) {
    // Control the log level yourself.
    if (level >= pulsar_INFO) {
         printf("[%u] [%s] [%d] [%s] \n", level, file, line, message);
    }
}
```

This approach avoids the need to expose additional interfaces to set the log level.

RobertIndie: Take this as an example:
https://github.com/apache/pulsar-client-cpp/blob/5d5ed2acc2e89aed90b593343369f360644440e7/lib/MessageCrypto.cc#L153-L157
In this case, lines 154-156 will always be executed even though the user doesn't want to get the debug log. We usually use this approach to avoid executing some performance-critical code when we don't need to print the debug log.

shibd: Good point. I will change the solution.

shibd: @RobertIndie I added `pulsar_client_configuration_set_logger_and_level` to  change log level, PTAL.

RobertIndie (approved):

BewareMyPower (commented):
I prefer to change the definition of `pulsar_logger` and use the new function to deprecate the old function to set the logger:

```c
// NOTE: you need to include <stdbool.h> to have bool in C
typedef struct pulsar_logger_t {
    void (*log)(pulsar_logger_level_t level, const char *file, int line, const char *message, void *ctx);

    bool (*is_enabled)(pulsar_logger_level_t level);
} pulsar_logger_t;

/**
 * Deprecated. Use `pulsar_client_configuration_set_logger_t` instead.
 */
PULSAR_PUBLIC void pulsar_client_configuration_set_logger(pulsar_client_configuration_t *conf,
                                                          pulsar_logger logger, void *ctx);

PULSAR_PUBLIC void pulsar_client_configuration_set_logger_t(pulsar_client_configuration_t *conf,
                                                            pulsar_logger_t logger, void *ctx);
```

This abstraction will be more like the C++'s `Logger`.

BewareMyPower:
The design might not be good, I think how to configure the logger might need more considerations. It's better to make the configuration more flexible.

BewareMyPower:
I will merge this PR for now. It's only included in the next 3.2.0 release. If I have a better idea, I will open a PR to remove the new API in this PR.

BewareMyPower (approved):

PR #160, opened 2022-12-29
[fix] Redeliver messages that can't be decrypted.

shibd:
Master Issue: https://github.com/apache/pulsar/pull/3097

### Motivation
Customers who can't decrypt their message must be given a chance to re-decrypt the message if redelivery is enabled.

### Modifications
- When the decrypted message fails, add to `unAckedMessageTrackerPtr_` to wait for redelivery.

### Verifying this change
 - Added `testRedeliveryOfDecryptionFailedMessages` to cover it.
 
### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/ConsumerTest.cc
+++ b/tests/ConsumerTest.cc
@@ -902,9 +902,8 @@ TEST(ConsumerTest, testRedeliveryOfDecryptionFailedMessages) {
     std::string topicName = "testRedeliveryOfDecryptionFailedMessages" + std::to_string(time(nullptr));
     std::string subName = "sub-test";
 
-    // TODO remove ../
-    std::string PUBLIC_CERT_FILE_PATH = "../../test-conf//public-key.client-rsa.pem";
-    std::string PRIVATE_CERT_FILE_PATH = "../../test-conf//private-key.client-rsa.pem";
+    std::string PUBLIC_CERT_FILE_PATH = "../test-conf//public-key.client-rsa.pem";
+    std::string PRIVATE_CERT_FILE_PATH = "../test-conf//private-key.client-rsa.pem";
     std::shared_ptr<pulsar::DefaultCryptoKeyReader> keyReader =
         std::make_shared<pulsar::DefaultCryptoKeyReader>(PUBLIC_CERT_FILE_PATH, PRIVATE_CERT_FILE_PATH);
 
--- a/tests/ConsumerTest.cc
+++ b/tests/ConsumerTest.cc
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "HttpHelper.h"
+#include "NoOpsCryptoKeyReader.h"
 #include "PulsarFriend.h"
 #include "lib/ClientConnection.h"
 #include "lib/Future.h"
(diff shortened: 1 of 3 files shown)


tests/ConsumerTest.cc
```diff
@@ -895,6 +896,90 @@ TEST(ConsumerTest, testGetLastMessageIdBlockWhenConnectionDisconnected) {
...
+
+    ConsumerConfiguration consConfig3;
+    consConfig3.setConsumerType(pulsar::ConsumerShared);
+    consConfig3.setCryptoFailureAction(ConsumerCryptoFailureAction::FAIL);
+    consConfig3.setUnAckedMessagesTimeoutMs(10000);
+    Consumer consumer3;
+    ASSERT_EQ(ResultOk, client.subscribe(topicName, subName, consConfig3, consumer3));
+    auto consumer3ImplPtr = PulsarFriend::getConsumerImplPtr(consumer3);
+    consumer3ImplPtr->unAckedMessageTrackerPtr_.reset(new UnAckedMessageTrackerEnabled(
+        100, 100, PulsarFriend::getClientImplPtr(client), static_cast<ConsumerImplBase&>(*consumer3ImplPtr)));
```
BewareMyPower, line 943: Do we need another consumer for this test? I think only one consumer is enough to verify the case that messages cannot be decrypted.

tests/ConsumerTest.cc
```diff
@@ -895,6 +896,90 @@ TEST(ConsumerTest, testGetLastMessageIdBlockWhenConnectionDisconnected) {
...
+        ASSERT_EQ(ResultOk, consumer1.receive(msg, 2000));
+        ASSERT_EQ(ResultOk, consumer1.acknowledge(msg));
+        receivedMsgs.insert(msg.getDataAsString());
+    }
+
+    ASSERT_EQ(receivedMsgs.size(), numberOfMessages);
+    for (int i = 0; i < numberOfMessages; i++) {
+        std::stringstream expected;
+        expected << msgContent << i;
+        ASSERT_TRUE(receivedMsgs.count(expected.str()));
```
BewareMyPower, line 972: It's better not to convert from `size_t` to `bool` immediately, i.e. check `xxx.count(key) > 0`.

BTW, in this case, we can store the messages in an ordered set and then just compare the sent messages and received messages.

```c++
    std::set<std::string> valuesSent;
    for (int i = 0; i < numberOfMessages; i++) {
        auto value = msgContent + std::to_string(i);
        valuesSent.emplace(value);
        msg = MessageBuilder().setContent(value).build();
        ASSERT_EQ(ResultOk, producer.send(msg));
    }

    /* ... */

    // All messages would be received by consumer 1
    std::set<std::string> valuesReceived;
    for (int i = 0; i < numberOfMessages; i++) {
        ASSERT_EQ(ResultOk, consumer1.receive(msg, 2000));
        ASSERT_EQ(ResultOk, consumer1.acknowledge(msg));
        valuesReceived.emplace(msg.getDataAsString());
    }
    ASSERT_EQ(valuesSent, valuesReceived);
```

tests/ConsumerTest.cc
shibd: Yes, `consumer2` is set the `NoOpsCryptoKeyReader` to cover [change](https://github.com/apache/pulsar-client-cpp/pull/160/files#diff-102603faf4c8eaedb96f25866948c8b38eaf6064f1a77410572a46bec3f48d8dR712-R713), and `consumer3` is not set CryptoKeyReader to cover [change](https://github.com/apache/pulsar-client-cpp/pull/160/files#diff-102603faf4c8eaedb96f25866948c8b38eaf6064f1a77410572a46bec3f48d8dR690-R691)

tests/ConsumerTest.cc
```diff
@@ -895,6 +896,91 @@ TEST(ConsumerTest, testGetLastMessageIdBlockWhenConnectionDisconnected) {
...
+    }
+    ASSERT_EQ(valuesSent, valuesReceived);
+
+    // Consuming from consumer 2 and 3 again just to be sure
+    // no message should be returned since they can't decrypt the message
+    ASSERT_EQ(ResultTimeout, consumer2.receive(msg, 1000));
+    ASSERT_EQ(ResultTimeout, consumer3.receive(msg, 1000));
+
+    ASSERT_EQ(ResultOk, client.close());
+}
```
RobertIndie, line 982: Where do we verify the message is added to unAckedMessageTrackerPtr_ correctly?

BewareMyPower (approved):

tests/ConsumerTest.cc
shibd: When `consumer2` and `consumer3` receive a msg and the decrypted message failed, it will be added to `unAckedMessageTracker` and redelivery. So, just `consumer1` will receive all messages.

RobertIndie (approved):

PR #162, opened 2022-12-29
[feat] Add pulsar_logger_t as the configurable C logger

BewareMyPower:
### Motivation

The current `pulsar_client_configuration_set_logger` API can only configure the `Logger::log` method, but the` Logger::isEnabled` method cannot be configured via C API.

https://github.com/apache/pulsar-client-cpp/pull/158 added a `pulsar_client_configuration_set_logger_and_level` function to configure a log level, but it's not flexible. For example, the log level might be modified dynamically (though it's a complicated case).

### Modifications

Add a `pulsar_logger_t` struct and the related
`pulsar_client_configuration_set_logger_t` function to configure it as the C logger API. The `is_enabled` and `log` fields of the struct are the responding methods of the `isEnabled` and `log` methods in C++ `Logger`.

Then add a `LogContext` example in `SampleCustomLoggerCApi.c` to print logs to a file or standard output.

Eliminate the `pulsar_client_configuration_set_logger_and_level` function.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/examples/SampleCustomLoggerCApi.c
+++ b/examples/SampleCustomLoggerCApi.c
@@ -17,6 +17,7 @@
  * under the License.
  */
 
+#include <ctype.h>
 #include <pulsar/c/client.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -24,7 +25,7 @@
 #include <time.h>
 
 char *current_time() {
-    char *time_str = malloc(128);
+    char *time_str = (char *)malloc(128);
     struct tm *p;
     time_t now = time(0);
     p = gmtime(&now);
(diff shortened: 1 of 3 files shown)


BewareMyPower:
@shibd @RobertIndie I added a more flexible interface to configure the C logger, so I removed the function added in the previous PR (#158). The PR also documents and shows to use the `ctx`, which is now a `field` of `pulsar_logger_t`.

shibd (approved):

Demogorgon314 (approved):

Issue #167, opened 2023-01-05
Flaky test: ClientTest.testCloseClient


BewareMyPower:
[example failure](https://github.com/apache/pulsar-client-cpp/actions/runs/3797580369/jobs/6458643223)

```
FAILED TESTS (5/418):
     743 ms: ./pulsar-tests ClientTest.testReferenceCount (try #1)
   11653 ms: ./pulsar-tests ClientTest.testCloseClient (try #1)
    9692 ms: ./pulsar-tests ClientTest.testCloseClient (try #2)
    8242 ms: ./pulsar-tests ClientTest.testCloseClient (try #3)
    8126 ms: ./pulsar-tests ClientTest.testCloseClient (try #4)
```

```
// ...
2022-12-29T01:22:46.5133298Z 2022-12-29 01:22:46.253 INFO  [140291349905088] ConnectionPool:97 | Created connection for pulsar://localhost:6650
2022-12-29T01:22:46.5133649Z 2022-12-29 01:22:46.253 INFO  [140291346781760] ClientConnection:387 | [[::1]:60292 -> [::1]:6650] Connected to broker
2022-12-29T01:22:46.5133967Z 2022-12-29 01:22:46.253 INFO  [140291349905088] ClientImpl:516 | Closing Pulsar client with 0 producers and 0 consumers
2022-12-29T01:22:46.5134331Z 2022-12-29 01:22:46.254 INFO  [140291330000448] ClientConnection:1602 | [[::1]:60292 -> [::1]:6650] Connection disconnected
2022-12-29T01:22:46.5134974Z 2022-12-29 01:22:46.254 ERROR [140291346781760] ClientImpl:183 | Error Checking/Getting Partition Metadata while creating producer on persistent://public/default/client-test-close-client-1672276958 -- NotConnected
2022-12-29T01:22:46.5135316Z 2022-12-29 01:22:46.254 INFO  [140291330000448] ClientConnection:268 | [[::1]:60292 -> [::1]:6650] Destroyed connection
2022-12-29T01:22:46.5135700Z 2022-12-29 01:22:46.255 INFO  [140291349905088] ClientConnection:190 | [<none> -> pulsar://localhost:6650] Create ClientConnection, timeout=10000
2022-12-29T01:22:46.5136036Z 2022-12-29 01:22:46.255 INFO  [140291349905088] ConnectionPool:97 | Created connection for pulsar://localhost:6650
[...]

PR #168, opened 2023-01-05
Fix segfault caused by socket I/O on a closed io_service

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/167

### Motivation

Here are some debugging info when the segfault happened in `testCloseClient`. The outputs have been trimmed to make them clear.

An example crash at `async_write`:

```
#12 0x00007ffff7496dad in basic_stream_socket<...>::boost::asio::async_write /usr/include/boost/asio/impl/write.hpp:512
#13 0x00007ffff748e003 in ClientConnection::asyncWrite lib/ClientConnection.h:245
#14 0x00007ffff746e0b6 in ClientConnection::handleHandshake (this=0x555555e689d0) lib/ClientConnection.cc:502
```

Another example crash at `async_receive`:

```
#6  0x00007ffff7497247 in basic_stream_socket<...>::async_receive /usr/include/boost/asio/basic_stream_socket.hpp:677
#7  0x00007ffff748e647 in ClientConnection::asyncReceive lib/ClientConnection.h:258
#8  0x00007ffff746fa5d in ClientConnection::readNextCommand lib/ClientConnection.cc:606
```

The frame where it crashed:

```
245       if (descriptor_data->shutdown_)
(gdb) p descriptor_data
$2 = (boost::asio::detail::epoll_reactor::per_descriptor_data &) @0x555555e4a780: 0x0
```

We can see the socket descriptor is `nullptr`. The root cause is when `async_receive` or `async_write` is called, the `io_service` object might be closed. This case happened when `createProducerAsync` is called, the actual producer creation continues in another thread, while the `client.close()` happens in the current thread.

### Modifications

Check if the `ClientConnection` is closed before `async_receive` or `async_write`. To avoid the use of lock, changing the `state_` field to atomic.

### Verifications

```bash
./tests/pulsar-tests --gtest_filter='ClientTest.testCloseClient' --gtest_repeat=20
```

It never crashed after applying this patch.

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -812,7 +812,7 @@ void ClientConnection::handleIncomingMessage(const proto::CommandMessage& msg, b
 void ClientConnection::handleIncomingCommand(BaseCommand& incomingCmd) {
     LOG_DEBUG(cnxString_ << "Handling incoming command: " << Commands::messageType(incomingCmd.type()));
 
-    switch (state_) {
+    switch (state_.load()) {
         case Pending: {
             LOG_ERROR(cnxString_ << "Connection is not ready yet");
             break;
--- a/lib/ClientConnection.h
+++ b/lib/ClientConnection.h
@@ -22,6 +22,7 @@
 #include <pulsar/ClientConfiguration.h>
 #include <pulsar/defines.h>
 
+#include <atomic>
 #include <boost/any.hpp>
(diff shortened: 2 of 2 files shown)


shibd (approved):

shibd:
@RobertIndie Can you take a look? This error occurs frequently.

RobertIndie (approved):

Issue #150, opened 2022-12-19
[Feature] PIP 23: Message Tracing By Interceptors


RobertIndie:
## Motivation

https://github.com/apache/pulsar/wiki/PIP-23%3A-Message-Tracing-By-Interceptors

## Implementation

The java implementation for referring: https://github.com/apache/pulsar/pull/2471

RobertIndie:
All features from PIP 23 have been implemented.

PR #169, opened 2023-01-06
[feat] Add producer interceptor

RobertIndie:
Master Issue: #150

### Motivation

This is the producer interceptor implementation of #150.

### Modifications

* Add ProducerInterceptor interface
* Add `intercept` and `getInterceptors` in `ProducerConfiguration`.
* Triggering the interceptor when `beforeSend`, `onSendAcknowledgement`, `onPartitionsChange` and `close`.

### Verifying this change

This change added tests.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


include/pulsar/ProducerConfiguration.h
```diff
@@ -532,6 +533,10 @@ class PULSAR_PUBLIC ProducerConfiguration {
      */
     ProducerAccessMode getAccessMode() const;
 
+    ProducerConfiguration& intercept(const std::initializer_list<ProducerInterceptorPtr> interceptors);
```
BewareMyPower, line 536: An `initializer_list` as the parameter is strange. Adding an `addInterceptor` API could be better and easier to use.

```c++
ProducerConfiguration& addInterceptor(const ProducerInterceptorPtr& interceptor);
```

include/pulsar/ProducerInterceptor.h
```diff
@@ -0,0 +1,114 @@
...
+    virtual void onSendAcknowledgement(const Producer& producer, Result result, const Message& message,
+                                       const MessageId& messageID) = 0;
+
+    /**
+     * This method is called when partitions of the topic (partitioned-topic) changes.
+     *
+     * @param topicName topic name
+     * @param partitions new updated partitions
+     */
+    virtual void onPartitionsChange(const std::string& topicName, const int partitions) {}
```
BewareMyPower, line 108:
```suggestion
    virtual void onPartitionsChange(const std::string& topicName, int partitions) {}
```

Don't add `const` to value types in function parameters. See https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters

lib/PartitionedProducerImpl.cc
```diff
@@ -443,6 +445,7 @@ void PartitionedProducerImpl::handleGetPartitions(Result result,
                 producers_.push_back(producer);
             }
             // `runPartitionUpdateTask()` will be called in `handleSinglePartitionProducerCreated()`
+            onPartitionsChange(getTopic(), newNumPartitions);
```
BewareMyPower, line 448: Unlock `producersLock` before `onPartitionsChange` because it might took long time and it doesn't need to be protected by `producersLock`.

BewareMyPower (changes requested):

lib/ProducerImpl.cc
```diff
@@ -975,5 +984,33 @@ void ProducerImpl::asyncWaitSendTimeout(DurationType expiryTime) {
 
 ProducerImplWeakPtr ProducerImpl::weak_from_this() noexcept { return shared_from_this(); }
 
+Message ProducerImpl::beforeSend(const Producer& producer, const Message& message) const {
+    if (interceptors_.size() == 0) {
```
BewareMyPower, line 988:
```suggestion
    if (interceptors_.empty()) {
```

include/pulsar/ProducerConfiguration.h
RobertIndie: This is to make it consistent with [the java API](https://github.com/apache/pulsar/blob/932023da50a124384198778bf0b7daa8bf9a3ba4/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ProducerBuilder.java#L552).

Using it, the user can easily add the interceptors by this:
```cpp
conf.intercept({interceptor1, interceptor2, interceptor3});
```

BewareMyPower: Yes. I see the advantage.

BewareMyPower: Why not use a `std::vector`?

BewareMyPower (changes requested):
All the tests should check what `latch.wait` returns and the side effects made by the different interceptors.

include/pulsar/Producer.h
```diff
@@ -23,13 +23,16 @@
...
+#include <functional>
 #include <memory>
 
 namespace pulsar {
 class ProducerImplBase;
 class PulsarWrapper;
 class PulsarFriend;
 
+typedef std::function<void(Result, const MessageId& messageId)> SendCallback;
+typedef std::function<void(Result)> CloseCallback;
```
BewareMyPower, line 35: Why did you add these extra definitions?

include/pulsar/ProducerInterceptor.h
```diff
@@ -0,0 +1,114 @@
...
+class Producer;
+
+class PULSAR_PUBLIC ProducerInterceptor {
+   public:
+    virtual ~ProducerInterceptor() {}
+
+    /**
+     * Close the interceptor
+     */
+    virtual void close() {}
```
shibd, line 49: When the producer close, do we need to invoke this close()? I see this is not invoked in Java Client, maybe it a bug?

include/pulsar/ProducerInterceptor.h
```diff
@@ -0,0 +1,114 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef PULSAR_PRODUCER_INTERCEPTOR_H
+#define PULSAR_PRODUCER_INTERCEPTOR_H
+
+#include <pulsar/Message.h>
+#include <pulsar/Producer.h>
```
shibd, line 24: Line 40 `Producer` has been declared. There is no need to import `Producer.h` here.

include/pulsar/Producer.h
```diff
@@ -23,13 +23,16 @@
...
 
+#include <functional>
 #include <memory>
 
 namespace pulsar {
 class ProducerImplBase;
 class PulsarWrapper;
 class PulsarFriend;
 
+typedef std::function<void(Result, const MessageId& messageId)> SendCallback;
```
shibd, line 34: You don't need to redefine these here. Remove the import of `Producer.h` from `ProducerInterceptor.h` and the compilation will succeed

include/pulsar/ProducerInterceptor.h
shibd: ProducerInterceptor.h  -> Producer.h -> ProducerConfiguration.h -> ProducerInterceptor.h

include/pulsar/ProducerConfiguration.h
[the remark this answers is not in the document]
BewareMyPower: I just missed the code here. Generally, `std::initializer_list` should be passed by value. BTW, the `const` could be removed.

See `std::vector`'s 10th constructor: https://en.cppreference.com/w/cpp/container/vector/vector

RobertIndie: I don't have a strong opinion on it. Just because `initializer_list ` is more often used as the lightweight container type for the input parameter. I could change to using vector anyway.

tests/InterceptorsTest.cc
```diff
@@ -0,0 +1,157 @@
...
+    ProducerConfiguration conf;
+    conf.intercept({std::make_shared<TestInterceptor>(latch)});
+    Producer producer;
+    client.createProducer(topic, conf, producer);
+
+    Message msg = MessageBuilder().setContent("content").build();
+    Result result = producer.send(msg);
+    ASSERT_EQ(result, ResultOk);
+
+    latch.wait(std::chrono::seconds(5));
```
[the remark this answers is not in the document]
RobertIndie: Thanks. I have added the check for the return value

> as well as the received message, which should have set the "key" property through the interceptor.

Already did that in the `TestInterceptor.onSendAcknowledgement()`.

include/pulsar/Producer.h
RobertIndie: Fixed by https://github.com/apache/pulsar-client-cpp/pull/169#discussion_r1066689463

include/pulsar/ProducerConfiguration.h
RobertIndie: OK. I have decided to use `vector` here.

include/pulsar/ProducerInterceptor.h
RobertIndie: Yes. Fixed it. The java client also need to be fixed.

include/pulsar/ProducerInterceptor.h
RobertIndie: Thanks.

BewareMyPower:
Could you solve the conflicts?

RobertIndie:
> Could you solve the conflicts?

Done. PTAL again. Thanks.

lib/ProducerInterceptors.h
```diff
@@ -0,0 +1,56 @@
...
+#include <atomic>
+#include <utility>
+#include <vector>
+
+namespace pulsar {
+
+class ProducerInterceptors {
+   public:
+    explicit ProducerInterceptors(std::vector<ProducerInterceptorPtr> interceptors)
+        : interceptors_(std::move(interceptors)) {}
```
BewareMyPower, line 33: Even if you use `std::move(interceptors)`, the copy will still happen because this constructor accepts an argument by value, not reference.

For example,

```c++
#include <iostream>
#include <vector>
using namespace std;

struct Foo {
  Foo() { cout << "Foo()" << endl; }
  Foo(const Foo &) { cout << "Foo copy" << endl; }
  Foo(Foo &&foo) noexcept { cout << "Foo move" << endl; }
};

struct FooVector {
  FooVector(vector<Foo> v) : v_(std::move(v)) {}

  vector<Foo> v_;
};

int main(int argc, char* argv[]) {
  vector<Foo> v;
  v.push_back(Foo{});
  cout << "--------" << endl;
  FooVector foos{v};
  return 0;
}
```

You will see the following output:

```
Foo()
Foo move
--------
Foo copy
```

If you want to avoid the copy, you have to pass `std::move(v)` like:

```c++
  FooVector foos{std::move(v)};
```

However, in your case, you call it here:

```c++
auto interceptors = std::make_shared<ProducerInterceptors>(conf.getInterceptors());
```

If you changed it to `std::move(conf.getInterceptors())`, the `conf` will hold no interceptor and an unexpected behavior might happen (imagine users might retrieve the interceptors from the `ProducerConfiguration` instance).

Therefore, in short, it's better just to keep the copy semantics instead of the move semantics:

```c++
explicit ProducerInterceptors(const std::vector<ProducerInterceptorPtr>& interceptors) : interceptors_(interceptors) {}
```

This is not a critical path, so copying is acceptable here.

lib/ProducerImplBase.h
```diff
@@ -48,6 +49,9 @@ class ProducerImplBase {
     virtual void flushAsync(FlushCallback callback) = 0;
     virtual bool isConnected() const = 0;
     virtual uint64_t getNumberOfConnectedProducer() = 0;
+
+   protected:
+    ProducerInterceptorsPtr interceptors_;
```
BewareMyPower, line 54: Actually I do not suggest adding a field here. We'd better keep this abstract class like an interface in Java. It works because C++ supports multi inheritance.

I'd like to move this field to both `ProducerImpl` and `PartitionedProducerImpl`, then we can initialize this field in the initializer list like:

```c++
       conf_(conf),
       interceptors_(interceptors),
```

But I have no objection to multiple inheritance. If you keeps this design, you can add a constructor to it:

```c++
    ProducerImplBase(const ProducerInterceptorsPtr& interceptors) : interceptors_(interceptors) {}
```

Then initialize it in the initializer list in `ProducerImpl` and `PartitionedProducerImpl`:

```c++
      ProducerImplBase(interceptors),
      conf_(conf),
```

BewareMyPower (commented):
I've left two comments about the coding style. These should not be blockers so I dismissed my requested changes before.

lib/ProducerInterceptors.h
RobertIndie: It's a warning from the clang-tidy: `Clang-Tidy: Parameter 'interceptors' is passed by value and only copied once; consider moving it to avoid unnecessary copies `.

We should copy the value here and it's okay to do that because it's a pointer type. It will raise a warning here if we don't use `std::move`. But I'm okay for both.

lib/ProducerImplBase.h
RobertIndie: Keeping it as an `interface` makes sense to me. I have moved them to `ProduerImpl` and `PartitonedProducerImpl`.

BewareMyPower:
@RobertIndie The main branch is broken due to a recent upgrade of `macos-12` runner image. The fix is here: https://github.com/apache/pulsar-client-cpp/pull/199

lib/ProducerInterceptors.h
BewareMyPower: It seems that it would be better to apply the `clang-tidy` check into the CI.

BewareMyPower: BTW, I'm also okay with the current method signature.

RobertIndie: > It seems that it would be better to apply the `clang-tidy` check into the CI.

+1. I have created a issue to track it: https://github.com/apache/pulsar-client-cpp/issues/200

BewareMyPower (approved):

Anonymitaet:
Hi @RobertIndie, thanks for introducing this feature!

As we discussed just now, for the doc side:

### ✅ Pulsar feature matrix

I've ticked the boxes of ["Interceptors" for C++ producer and consumer\(https://docs.google.com/spreadsheets/d/1YHYTkIXR8-Ql103u-IMI18TXLlGStK8uJjDsOOA0T20/edit#gid=1784579914&range=48:48)

### ✅ Pulsar client docs

Could you please provide the following tech inputs? TIA!

The following docs are only added to `NEXT` since this feature is server-independent. 

- [C++ > Producer > Interceptor](https://pulsar.apache.org/docs/next/client-libraries-producers/#intercept-messages)

image

- [C++ > Consumer > Interceptor](https://pulsar.apache.org/docs/next/client-libraries-consumers/#intercept-messages)

image

Issue #163, opened 2022-12-30
[feature request] Client support set ServiceUrlProvider


shibd:
### Motivation

This is an earlier feature of the Pulsar java client and has no related PIP.

Refer to Java client PR: https://github.com/apache/pulsar/pull/2543

shibd:
I will start work on it.

PR #170, opened 2023-01-09
[feature] Support set ServiceUrlProvider when create client.

shibd:
Master Issue: https://github.com/apache/pulsar/pull/2543

### Motivation

#163 

`ServiceUrlProvider` is provided so that users can dynamically update serviceUrl, and can be implemented [#164 cluster level auto failover](https://github.com/apache/pulsar-client-cpp/issues/164).

### Modifications
- Add the `ServiceUrlProvider`  function define.
- Add the `updateServiceUrl` method in the Client.
- `Client` add new constructor method to support set  `ServiceUrlProvider` instance param.

### Verifying this change
- Add `ServiceUrlProviderTest` to cover custom `ServiceUrlProvider`.

### Documentation

- [x] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Client.h
+++ b/include/pulsar/Client.h
@@ -39,6 +39,7 @@ typedef std::function<void(Result, Consumer)> SubscribeCallback;
 typedef std::function<void(Result, Reader)> ReaderCallback;
 typedef std::function<void(Result, const std::vector<std::string>&)> GetPartitionsCallback;
 typedef std::function<void(Result)> CloseCallback;
+typedef std::function<const std::string&()> ServiceUrlProvider;
 
 class ClientImpl;
 class PulsarFriend;
@@ -65,6 +66,31 @@ class PULSAR_PUBLIC Client {
      * @throw std::invalid_argument if `serviceUrl` is invalid
      */
     Client(const std::string& serviceUrl, const ClientConfiguration& clientConfiguration);
+
+    /**
+     * Create a Pulsar client object connecting to the specified cluster address and using the default
+     * configuration.
+     *
+     * <p>Instead of specifying a static service URL string (with {@link #serviceUrl(String)}), an application
+     * can pass a {@link ServiceUrlProvider} function that dynamically provide a service URL.
+     *
+     * @param serviceUrlProvider The serviceUrlProvider used to generate ServiceUrl.
+     * @throw std::invalid_argument if `serviceUrlProvider()` return a invalid url.
+     */
+    Client(ServiceUrlProvider serviceUrlProvider);
+
+    /**
+     * Create a Pulsar client object connecting to the specified cluster address and using the specified
+     * configuration.
+     *
+     * <p>Instead of specifying a static service URL string (with {@link #serviceUrl(String)}), an application
+     * can pass a {@link ServiceUrlProvider} instance that dynamically provide a service URL.
+     *
+     * @param serviceUrlProvider The serviceUrlProvider used to generate ServiceUrl.
+     * @param clientConfiguration the client configuration to use
+     * @throw std::invalid_argument if `serviceUrlProvider()` return a invalid url.
+     */
+    Client(ServiceUrlProvider serviceUrlProvider, const ClientConfiguration& clientConfiguration);
 
     /**
      * Create a producer with default configuration
@@ -332,6 +358,17 @@ class PULSAR_PUBLIC Client {
      * @since 2.3.0
      */
     void getPartitionsForTopicAsync(const std::string& topic, GetPartitionsCallback callback);
+
+    /**
+     * Update the service URL this client is using.
+     *
+     * <p>This will force the client close all existing connections and to restart service discovery to the
+     * new service endpoint.
+     *
+     * @param serviceUrl
+     *            the new service URL this client should connect to
+     */
+    Result updateServiceUrl(const std::string& serviceUrl);
 
     /**
      *
--- a/lib/ClientImpl.cc
+++ b/lib/ClientImpl.cc
@@ -497,6 +497,24 @@ void ClientImpl::getPartitionsForTopicAsync(const std::string& topic, GetPartiti
                   std::placeholders::_2, topicName, callback));
 }
 
+Result ClientImpl::updateServiceUrl(const std::string& serviceUrl) {
+    Lock lock(mutex_);
+    if (state_ != Open) {
+        lock.unlock();
+        return ResultAlreadyClosed;
+    }
+
+    LOG_INFO("Updating service URL to " << serviceUrl);
+    try {
+        serviceNameResolver_.updateServiceUrl(serviceUrl);
+    } catch (const std::invalid_argument& e) {
+        LOG_ERROR("Invalid service-url " << serviceUrl << "provided " << e.what());
+        return ResultInvalidUrl;
+    }
+    pool_.disconnect();
+    return ResultOk;
+}
+
 void ClientImpl::closeAsync(CloseCallback callback) {
     if (state_ != Open) {
         if (callback) {
--- a/lib/ServiceNameResolver.h
+++ b/lib/ServiceNameResolver.h
@@ -21,6 +21,7 @@
 #include <assert.h>
 
 #include <atomic>
+#include <memory>
 
 #include "ServiceURI.h"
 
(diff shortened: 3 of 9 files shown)


BewareMyPower (changes requested):
We only need to support passing a `std::function<std::string()>`. The `initialize` and `close` methods are redundant. IMO, the interface was not designed well in Java caused by the abuse of virtual functions. It makes it impossible to use a lambda as the service URL provider in Java.

If you really want to wrap the initialization and close logics, you can implement the function like:

```c++
class ServiceUrlProvider {
 public:
  ServiceUrlProvider(/* we can pass any type we want */) { /* initializations ... */ }
  void close() {/* close the resource... */}
  std:string operator()() const { /* ... */ }
};
```

For example, if we need to load the available URLs from a file periodically, we can pass the file path in the constructor and close the files in `close`.

There are also internal implementations in Java client like `AutoClusterFailover`. However, it actually makes use of `PulsarClientImpl`, not `PulsarClient`.

shibd:
> We only need to support passing a std::function<std::string()>. The initialize and close methods are redundant.

 The initialize and close methods are useful. We need pass `Client&` to `ServiceUrlProvider` by `initialize` method.

https://github.com/apache/pulsar-client-cpp/blob/7d91638513405ed3e9bc5e7cd782d30e8703679d/lib/Client.cc#L43

This way, if the user implements the `ServiceUrlProvider`. When it listens for a change in the service URL (such as through zk or other storage), it can invoke the `client.updateServiceUrl` method to update the URL.

shibd:
> We only need to support passing a std::function<std::string()>. The initialize and close methods are redundant.

Yes, you are right. The initialize and close methods are redundant.

In fact, The client already provides the `updateServiceUrl` method, which the user can use to implement dynamic update service URL. As for how the user implements it and how to close his resources, this is left to the user.

I changed to pass a `std::function<const std::string&()>` when creating a `Client`.

include/pulsar/Client.h
```diff
@@ -66,6 +67,31 @@ class PULSAR_PUBLIC Client {
...
+     * Create a Pulsar client object connecting to the specified cluster address and using the default
+     * configuration.
+     *
+     * <p>Instead of specifying a static service URL string (with {@link #serviceUrl(String)}), an application
+     * can pass a {@link ServiceUrlProvider} function that dynamically provide a service URL.
+     *
+     * @param serviceUrlProvider The serviceUrlProvider used to generate ServiceUrl.
+     * @throw std::invalid_argument if `serviceUrlProvider()` return a invalid url.
+     */
+    Client(ServiceUrlProvider serviceUrlProvider);
```
BewareMyPower, line 80: Instead of adding new constructor overloads, we should add `setServiceUrlProvider` to `ClientConfiguration`.

BewareMyPower (changes requested):

lib/ServiceNameResolver.h
```diff
@@ -29,29 +29,33 @@ namespace pulsar {
 class ServiceNameResolver {
```
BewareMyPower, line 29: Add a `#include <memory>` to this file for `std::make_shared`. The compilation did not fail because the `.cc` files that include this header already include `<memory>`.

BewareMyPower (changes requested):

lib/ClientImpl.cc
```diff
@@ -497,6 +497,20 @@ void ClientImpl::getPartitionsForTopicAsync(const std::string& topic, GetPartiti
                   std::placeholders::_2, topicName, callback));
 }
 
+Result ClientImpl::updateServiceUrl(const std::string& serviceUrl) {
+    LOG_INFO("Updating service URL to " << serviceUrl);
+
+    try {
+        serviceNameResolver_.updateServiceUrl(serviceUrl);
+        lookupServicePtr_->updateServiceUrl(serviceUrl);
+    } catch (const std::exception& e) {
```
BewareMyPower, line 506: Catch `std::invalid_argument` instead of `std::exception`

BewareMyPower (changes requested):

lib/ClientImpl.cc
```diff
@@ -497,6 +497,20 @@ void ClientImpl::getPartitionsForTopicAsync(const std::string& topic, GetPartiti
                   std::placeholders::_2, topicName, callback));
 }
 
+Result ClientImpl::updateServiceUrl(const std::string& serviceUrl) {
+    LOG_INFO("Updating service URL to " << serviceUrl);
+
+    try {
+        serviceNameResolver_.updateServiceUrl(serviceUrl);
+        lookupServicePtr_->updateServiceUrl(serviceUrl);
```
BewareMyPower, line 505: Actually we don't need to add `updateServiceUrl` method to `LookupService` because in C++ client, all `LookupService` implementations just call  `serviceNameResolver_.updateServiceUrl(serviceUrl);`.

Unlike Java client, `HttpLookupService` uses a `HttpClient` class that accepts a service URL.

lib/ServiceNameResolver.h
```diff
@@ -29,29 +29,33 @@ namespace pulsar {
...
+        return serviceUri_->getScheme() == PulsarScheme::HTTP ||
+               serviceUri_->getScheme() == PulsarScheme::HTTPS;
     }
 
     const std::string& resolveHost() {
-        return serviceUri_.getServiceHosts()[(numAddresses_ == 1) ? 0 : (index_++ % numAddresses_)];
+        return serviceUri_->getServiceHosts()[(numAddresses_ == 1) ? 0 : (index_++ % numAddresses_)];
     }
 
+    void updateServiceUrl(const std::string& urlString) { serviceUri_.reset(new ServiceURI(urlString)); }
```
BewareMyPower, line 54: There is a problem that if the `urlString` contains multiple service URLs, we have to update `numAddresses_` as well. A possible solution is moving `numAddresses_` into `ServiceURI`.

BewareMyPower (changes requested):

include/pulsar/Client.h
shibd: This request the user to choose between `Client(ServiceUrlProvider serviceUrlProvider);` and `Client(const std::string& serviceUrl);` constructors.

If add `setServiceUrlProvider` to `ClientConfiguration`. When the user sets it this way:

```c++
ClientConfiguration config;
config.setServiceUrlProvider(serviceUrlProvider);
Client client(serviceUrl);
```

This will be confuse.

lib/ClientImpl.cc
shibd: Ok, I will remove `updateServiceUrl`  method of `LookupService `.

include/pulsar/Client.h
BewareMyPower: LGTM. Before this PR, `serviceUrl` is a required argument. But after this PR, only one of `serviceUrl` and `serviceUrlProvider` should be provided.

BewareMyPower, line 80:
```suggestion
    /**
     * @see Client(ServiceUrlProvider, const ClientConfiguration&)
     */
    Client(ServiceUrlProvider serviceUrlProvider) : Client(serviceUrlProvider, ClientConfiguration{}) {}
```

Avoid unnecessary duplicated code and API docs.

BewareMyPower:
> The initialize and close methods are redundant.

I reviewed the `ServiceUrlProvider` in Java client again. **I have thought `getServiceUrl()` is called each time the service URL is needed (e.g. topic lookup) but it's actually not.** The update of the service URL is actually done by `PulsarClient#updateServiceUrl`, which is triggered by the periodical task of `ServiceUrlProvide#initialize`.

If we want to implement `AutoClusterFailover` and `ControlledClusterFailover` in future, we actually don't need to implement a function that returns the service URL. Instead, we only need to implement the following methods:
- PulsarClient#updateServiceUrl
- PulsarClientImpl#updateAuthentication
- PulsarClientImpl#updateTlsTrustCertsFilePath

The way of using `ServiceUrlProvider` is bad. It's more like a provider that can modify the internal state of `PulsarClient`.

For C++ client, this PR implements the similar way with Java that the `ServiceUrlProvider` (a function that returns the service URL) is only triggerred when constructing the `Client` instance. There will be no difference between

```c++
ServiceUrlProvider provider = /* ... */;
Client client(provider());
```

and

```c++
ServiceUrlProvider provider = /* ... */;
String serviceUrl = provider();
Client client(serviceUrl);
```

l'm going to send an email to dev mail list to discuss.

shibd:
> If we want to implement AutoClusterFailover and ControlledClusterFailover in future, we actually don't need to implement a function that returns the service URL. Instead, we only need to implement the following methods:

Yes, Actually, the `ServiceUrlProvider ` implemented here is just convenient for the user to return a function in order to get the service URL from elsewhere.

If we do not expose the set  `ServiceUrlProvider`, It's no problem too. 

Use `ServiceUrlProvider`
```c++
Client client([]() {
    // get service URL from store.
    auto serviceUrl = xxx
    return serviceUrl;
})
```

`ServiceUrlProvider` is not used
```c++
    // get service URL from store.
    auto serviceUrl = xxx
    Client client(serviceUrl);
```

I exposed it just for ease of use.

shibd:
If you feel redundant to expose this method, I can remove it. What do you think? /cc @BewareMyPower

BewareMyPower:
I have sent an email: https://lists.apache.org/thread/cojz6qxpc7v0j8vy0w476ttt0f89zr3z

BewareMyPower:
In short, a natural design is to save the `ServiceUrlProvider` in `Client` and call `getServiceUrl()` method (in C++ it's just a function invocation) each time we need a service URL. Maybe we need to maintain a key-value (`serviceUrl -> ServiceNameResolver`), when the `serviceUrlProvider()` returns a different result from the key, creating a new `ServiceNameResolver` for related lookup requests.

shibd:
> In short, a natural design is to save the `ServiceUrlProvider` in `Client` and call `getServiceUrl()` method (in C++ it's just a function invocation) each time we need a service URL. Maybe we need to maintain a key-value (`serviceUrl -> ServiceNameResolver`), when the `serviceUrlProvider()` returns a different result from the key, creating a new `ServiceNameResolver` for related lookup requests.

I think you only need the client to provide the `updateServiceUrl` method. 

Users can take advantage of this approach to implement their own services. (Pseudocode)


```c++
class Store;
class AutomaticClusterService {
   public:
    explicit AutomaticClusterService(const Store& store, const Client pulsarClient)
        : store_(store), pulsarClient_(pulsarClient) {
        serviceUrlCache_ = store.queryServiceUrl();
        updateServiceUrlInterval_ = boost::posix_time::seconds(60);
        runUpdateServiceurl();
    }

    void start() {
        runUpdateServiceurl();
    }

    const std::string& getServiceUrl(){
        return serviceUrlCache_;
    }

    const void runUpdateServiceurl() {
        timer_->expires_from_now(updateServiceUrlInterval_);
        auto availableServiceUrl = store_.queryServiceUrl();
        timer_->async_wait([this](const boost::system::error_code& ec) {
          // Maybe some other implementation
          if (serviceUrlCache_ != availableServiceUrl) {
              serviceUrlCache_ = availableServiceUrl;
              pulsarClient_.updateServiceUrl(serviceUrlCache_);
            }
            runUpdateServiceurl();
        });
    }


   private:
    // External storage services
    Store store_;
    const Client pulsarClient_;
    std::string serviceUrlCache_;
    std::shared_ptr<boost::asio::deadline_timer> timer_;
    boost::posix_time::time_duration updateServiceUrlInterval_;
};

Store store;
Client client(store.queryServiceUrl());

AutomaticClusterService  automaticClusterService(store, client);
automaticClusterService.start();

//....

```

BewareMyPower:
> I think only need the client to provide the updateServiceUrl method.

This idea also looks good to me.

BewareMyPower:
@shibd We should delay this feature until a result of this [discussion](https://lists.apache.org/thread/cojz6qxpc7v0j8vy0w476ttt0f89zr3z).

BewareMyPower:
Close this PR for now.

PR #151, opened 2022-12-20
[feat] Support batch index acknowledgment

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/87

### Modifications

- Add an consumer configuration `setBatchIndexAckEnabled` to enable the batch index ACK. When it's enabled, passing the original `MessageId` instead of the `MessageIdImpl` that trunks the batch index to the ACK grouping tracker.
- Since now a `BatchedMessageIdImpl` could be accepted in the ACK grouping tracker, fix the compare logic.
- Support passing a `BitSet` in `Commands::newAck` and get the internal `BitSet` from `MessageId` in `Commands::newMultiMessageAck`.
- Skip the acknowledged batch indexes when receiving batched messages in `ConsumerImpl::receiveIndividualMessagesFromBatch`.

### Verifications

Modify `BitSetTest.testSet` to verify the `BitSet::get` method added in this commit.

Add `AcknowledgeTest.testBatchIndexAck` to test batch index ACK for all types of acknowledgment:
- Individual ACK for a single message
- Individual ACK for a list of messages
- Cumulative ACK

Add `AcknowledgeTest.testMixedCumulativeAck` to test the new compare logic between `BatchedMessageIdImpl` and `MessageIdImpl` works for cumulative ACK.

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

PR #175, opened 2023-01-17
Fix broken main branch by providing correct getBitSet implementation

BewareMyPower:
### Motivation

Currently the main branch is broken by the concurrent merge of https://github.com/apache/pulsar-client-cpp/pull/153 and https://github.com/apache/pulsar-client-cpp/pull/151. It's because when a batched message id is constructed from deserialization, there is no `getBitSet` implementation of the internal acker.

### Modifications

Add a `bool` parameter to `MessageIdImpl::getBitSet` to indicate whether the message ID is batched. The logic is similar with

https://github.com/apache/pulsar/blob/299bd70fdfa023768e94a8ee4347d39337b6cbd4/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L325-L327

and

https://github.com/apache/pulsar/blob/299bd70fdfa023768e94a8ee4347d39337b6cbd4/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L345-L347

Add a `testMessageIdFromBuild` to test the acknowledgment for a message ID without an acker could succeed for a consumer that enables batch index ACK.

### TODO

In future, https://github.com/apache/pulsar/pull/19031 might be migrated into the C++ client to fix the consumer that disables batch index ACK.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/BatchMessageAcker.h
+++ b/lib/BatchMessageAcker.h
@@ -37,13 +37,15 @@ class BatchMessageAcker {
     // by deserializing from raw bytes.
     virtual bool ackIndividual(int32_t) { return false; }
     virtual bool ackCumulative(int32_t) { return false; }
+    virtual const BitSet& getBitSet() noexcept {
+        static BitSet emptyBitSet;
+        return emptyBitSet;
+    }
 
     bool shouldAckPreviousMessageId() noexcept {
         bool expectedValue = false;
         return prevBatchCumulativelyAcked_.compare_exchange_strong(expectedValue, true);
     }
-
-    const BitSet& getBitSet() const noexcept { return bitSet_; }
 
    private:
     // When a batched message is acknowledged cumulatively, the previous message id will be acknowledged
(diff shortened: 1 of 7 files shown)


shibd:
I have a question:

> It's because when a batched message id is constructed from deserialization, there is no getBitSet implementation of the internal acker.

Is the code related to this passage here?

https://github.com/apache/pulsar-client-cpp/blob/d03ff20a96de695a6c2837d62862569faf11876b/lib/MessageIdBuilder.cc#L45-L47

Is setting bitSize equal to 0 for lazy initialization? To save memory?

If `bitSite` is initialized here, subsequent operations such as `ack` can reuse the logic here. Wouldn't it be better?

https://github.com/apache/pulsar-client-cpp/blob/f860566c05634dd097d8941ea38a2a939fcfd26e/lib/BatchedMessageIdImpl.h#L38-L40

lib/BatchMessageAcker.h
```diff
@@ -37,14 +37,16 @@ class BatchMessageAcker {
     // by deserializing from raw bytes.
```
shibd, line 37: The modification of this class to have solved the compilation error. Can we first open a new PR to merge it?

BewareMyPower: I will modify this PR after https://github.com/apache/pulsar-client-cpp/pull/177 is merged.

PR #185, opened 2023-02-02
[fix] Avoid resource leakage of AckGroupingTracker

erobot:
### Motivation

Avoid resource leakage of AckGroupingTracker.

The result of AckGroupingTracker leakage is waste of cpu. In our case, after a large number (30w+) of consumer creation failures, pulsar client io threads use 99% cpu when no operations.

There are two problems about the leakage:
* In the current code, if consumer creation failed, ConsumerImpl::ackGroupingTrackerPtr_ will not close.
* AckGroupingTrackerEnabled has race condition between close and reschedule, and may continue reschedule after close.

### Modifications

* ConsumerImpl: close ackGroupingTrackerPtr_ when shutdown
* AckGroupingTrackerEnabled: add state, and check state before reschedule

### Verifying this change

Can verify this change by adding a log and using a test program. The leakage resources are hold in boost::asio::io_service and seems not easy to write a unit test to verify.

The test program trys to trigger AckGroupingTracker leakage. If there is a leak, will print a lot of test log 'reschedule AckGroupingTrackerEnabled'. 

Add a test log:
```
void AckGroupingTrackerEnabled::scheduleTimer() {
    // ......
    this->timer_->async_wait([this, self](const boost::system::error_code& ec) -> void {
        if (!ec) {
            this->flush();
            this->scheduleTimer();
            LOG_INFO("reschedule AckGroupingTrackerEnabled");  // add a log when reschedule
        }
    });
}
```

Test program:
```
int main() {

    Consumer consumer;
    ConsumerConfiguration config;
    config.setConsumerType(ConsumerType::ConsumerExclusive);
    config.setAckGroupingTimeMs(1);

    // create exclusive consumer
    Result result = client.subscribe("persistent://public/default/my-topic", "consumer-1", config, consumer);
    if (result != ResultOk) {
        LOG_ERROR("Failed to subscribe: " << result);
        return -1;
    }

    // create other consumer, will fail
    for (int i = 0; i < 1000; ++i) {
        Result result =
            client.subscribe("persistent://public/default/my-topic", "consumer-1", config, consumer);
        assert(result != ResultOk);
        (void)result;
    }

    consumer.close();

    LOG_INFO("sleep, should not have reschedule logs below");
    std::this_thread::sleep_for(std::chrono::seconds(60));

    client.close();
    return 0;
}
```

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
Bug fix only.

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/AckGroupingTrackerEnabled.h
+++ b/lib/AckGroupingTrackerEnabled.h
@@ -71,15 +71,6 @@ class AckGroupingTrackerEnabled : public AckGroupingTracker {
     //! Method for scheduling grouping timer.
     void scheduleTimer();
 
-    //! State
-    enum State
-    {
-        NotStarted,
-        Ready,
-        Closed,
-    };
-    std::atomic<State> state_;
-
     //! The connection handler.
     HandlerBaseWeakPtr handlerWeakPtr_;
 
--- a/lib/AckGroupingTrackerEnabled.h
+++ b/lib/AckGroupingTrackerEnabled.h
@@ -21,6 +21,7 @@
 
 #include <pulsar/MessageId.h>
 
+#include <atomic>
 #include <boost/asio/deadline_timer.hpp>
 #include <cstdint>
 #include <mutex>
@@ -70,6 +71,9 @@ class AckGroupingTrackerEnabled : public AckGroupingTracker {
    protected:
     //! Method for scheduling grouping timer.
     void scheduleTimer();
+
+    //! State
+    std::atomic_bool isClosed_;
 
     //! The connection handler.
     HandlerBaseWeakPtr handlerWeakPtr_;
--- a/lib/AckGroupingTrackerEnabled.h
+++ b/lib/AckGroupingTrackerEnabled.h
@@ -71,6 +71,15 @@ class AckGroupingTrackerEnabled : public AckGroupingTracker {
     //! Method for scheduling grouping timer.
     void scheduleTimer();
 
+    //! State
+    enum State
+    {
+        NotStarted,
+        Ready,
+        Closed,
+    };
+    std::atomic<State> state_;
+
     //! The connection handler.
     HandlerBaseWeakPtr handlerWeakPtr_;
 
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -1212,6 +1212,9 @@ void ConsumerImpl::closeAsync(ResultCallback originalCallback) {
 const std::string& ConsumerImpl::getName() const { return consumerStr_; }
 
 void ConsumerImpl::shutdown() {
+    if (ackGroupingTrackerPtr_) {
+        ackGroupingTrackerPtr_->close();
+    }
     incomingMessages_.clear();
     possibleSendToDeadLetterTopicMessages_.clear();
     resetCnx();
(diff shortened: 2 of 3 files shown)


BewareMyPower:
I observed a similar bug long days ago: https://github.com/apache/pulsar/issues/8914. This PR might also fix that problem. The `scheduleTimer` method should not be called without any state check.

lib/AckGroupingTrackerEnabled.h
```diff
@@ -71,6 +71,15 @@ class AckGroupingTrackerEnabled : public AckGroupingTracker {
...
     void scheduleTimer();
 
+    //! State
+    enum State
+    {
+        NotStarted,
+        Ready,
+        Closed,
+    };
+    std::atomic<State> state_;
```
BewareMyPower, line 81: The `NotStarted` and `Closed` states are never used, would it better to just use a `atomic_bool` here?

BewareMyPower:
There might be something wrong with the Vcpkg installation, I will investigate the reason soon.

lib/AckGroupingTrackerEnabled.h
erobot: Using similar state definition style is a first thought and it can use `atomic_bool isClosed_` instead. May be more explicit.  I can change to isClosed style latter.

erobot: Updated to isClosed style.

lib/ConsumerImpl.cc
```diff
@@ -1212,6 +1212,9 @@ void ConsumerImpl::closeAsync(ResultCallback originalCallback) {
 const std::string& ConsumerImpl::getName() const { return consumerStr_; }
 
 void ConsumerImpl::shutdown() {
+    if (ackGroupingTrackerPtr_) {
```
shibd, line 1215: We can put it to the `cancelTimers()` method and remove the same logic in the `closeAsync()` method.

shibd (changes requested):

lib/ConsumerImpl.cc
shibd: > In the current code, if consumer creation failed, ConsumerImpl::ackGroupingTrackerPtr_ will not close.

BTW, In your case. It seems that the original design was to close resources in the destructor, but he didn't work.

https://github.com/apache/pulsar-client-cpp/blob/2980bbe9d3ef47eef4a1d468c3898b5a989b4df0/lib/AckGroupingTrackerEnabled.h#L48

erobot: I'm not sure should it be put into the cancelTimers(). Technically, it can be put into the cancelTimers(). But semantically closing a tracker (which means flushing the data and canceling the timer) is a little different than canceling a timer.

erobot: > BTW, In your case. It seems that the original design was to close resources in the destructor, but he didn't work.

The timer handler is a lambda function which captures a shared_ptr of AckGroupingTrackerEnabled. Because it always has a reference, so destructor would not invoke.

shibd: Get it, thanks for your explain.

shibd (approved):

BewareMyPower (approved):

PR #188, opened 2023-02-06
[Fix] NegativeAcksTracker need close when consumer closed.

shibd:
### Motivation

When consumers close, also need close `NegativeAcksTracker`.

### Modifications

- ConsumerImpl: close `negativeAcksTracker_`(cancel timer and clear nackedMessages_) when consumer shutdown or close.

### Verifying this change

- Add `ConsumerTest.testNegativeAcksTrackerClose` to cover it.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -1186,6 +1186,7 @@ void ConsumerImpl::closeAsync(ResultCallback originalCallback) {
     if (ackGroupingTrackerPtr_) {
         ackGroupingTrackerPtr_->close();
     }
+    negativeAcksTracker_.close();
 
     ClientConnectionPtr cnx = getCnx().lock();
     if (!cnx) {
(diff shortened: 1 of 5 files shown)


lib/ConsumerImpl.cc
```diff
@@ -1186,6 +1186,7 @@ void ConsumerImpl::closeAsync(ResultCallback originalCallback) {
     if (ackGroupingTrackerPtr_) {
         ackGroupingTrackerPtr_->close();
     }
+    negativeAcksTracker_.close();
```
RobertIndie, line 1189: The `shutdown` already closed it. We don't need to close it again here. The call of `shutdown` is at line 1165

shibd: These two invoke scenarios are not the same:

- `consumer.close()`: This must be closed before sending the `newCloseConsumer` command to prevent the `tracker` from sending useless requests.
- `consumer.shutdown()`: In addition to the method called by consumer.close, the destructor also calls this method. When the consumer `subscriptio`n fails, the consumer object destruction is triggered, and the `tracker` needs to be closed. This one is similar to this issue: https://github.com/apache/pulsar-client-cpp/pull/185

RobertIndie: Make sense to me.

RobertIndie (approved):

Issue #184, opened 2023-02-01
[Bug] OAuth2 authentication failed


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

- Broker: Pulsar 2.9.3.16 (StreamNative's release)
- Client: Pulsar C++ Client 3.1.1
- OS: Windows 11 (Visual Studio 2019) and Ubuntu 20.04 WSL (GCC 9)

### Minimal reproduce step

Set up an account on StreamNative cloud, create the topic `my-topic` in advance and grant the produce permission. The following example uses `test-auth` instance under the `sndev` organization, replace `YOUR-KEY-FILE-PATH` to the path of the credential file.

```c++
#include <pulsar/Client.h>
using namespace pulsar;

int main(int argc, char *argv[]) {
  std::string params = R"({
    "issuer_url": "https://auth.streamnative.cloud/",
    "private_key": "YOUR-KEY-FILE-PATH",
    "audience": "urn:sn:pulsar:sndev:test-auth"})";
  ClientConfiguration config;
  config.setAuth(AuthOauth2::create(params));
  Client client("pulsar+ssl://test-auth.sndev.snio.cloud:6651", config);

  Producer producer;
  auto result =
      client.createProducer("persistent://public/default/my-topic", producer);
  if (result != ResultOk) {
    std::cerr << "Failed to create producer: " << result << std::endl;
    return 1;
  }

  MessageId id;
  result = producer.send(MessageBuilder().setContent("hello").build(), id);
  if (result == ResultOk) {
    std::cout << "Sent to " << id << std::endl;
  } else {
    std::cerr << "Failed to send: " << result << std::endl;
  }

  client.close();
  return 0;
}
```

### What did you expect to see?

It should succeed

### What did you see instead?

Ubuntu:

```
2023-02-01 22:58:13.267 DEBUG [139889604585216] ClientConnection:591 | [<none> -> pulsar+ssl://test-auth.sndev.snio.cloud:6651] Resolved hostname test-auth.sndev.snio.cloud to SECRET_IP:6651
2023-02-01 22:58:14.564 INFO  [139889604585216] ClientConnection:387 | [MY_LOCAL_IP:38550 -> SECRET_IP:6651] Connected to broker
[...]

PR #190, opened 2023-02-07
[fix] Use ClientConfiguration::getTlsTrustCertsFilePath for the OAuth2 flow

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/184

### Motivation

`ClientConfiguration::getTlsTrustCertsFilePath` is only used for HTTP service URL lookup and the TLS handshake, but not used for the OAuth2 client credentials flow. If the issuer URL or the OAuth2 authorization server requires the CA cert, the CA cert can only be loaded from a specific path, which is determined by how to build the C++ client. e.g. it's `/etc/ssl/certs/ca-certificates.crt` when libcurl is built in Debian-based Linux distros and `/etc/pki/tls/certs/ca-bundle.crt` in RedHat-based Linux distros. A typical issue is that the Pulsar Python wheels are built in a [RedHat-based Linux image](https://github.com/pypa/manylinux), if they are installed in a Debian-based Linux distro, users can only fix this issue by copying the `ca-certificats.cert` into `/etc/pki/tls/certs/ca-bundle.crt`, while it should have been configured by the `tls_trust_certs_file_path` config.

### Modifications

Add a `AuthenticationDataProvider` implementation `InitialAuthData`, which holds the CA cert path. Then, in `AuthOauth2::getAuthData`, retrieve the path and pass it to the `ClientCredentialFlow` for HTTP requests performed by libcurl.

This solution is API and ABI compatible.

### Verifications

It's hard to add the test in CI because we need an OAuth2 server configured with the CA configured.

Follow the **How to reproduce** section in
https://github.com/apache/pulsar-client-cpp/issues/184#issuecomment-1420509146 to reproduce this issue. Apply this patch and build the `libpulsar.so` with `LINK_STATIC=ON`, then copy the `libpulsar.so` into the docker container (under `/app/lib`).

Run `./a.out` directly, you will still see the `AuthenticationError`. However, if you added the path of `libpulsar.so` to the `LD_LIBRARY_PATH`:

```bash
export LD_LIBRARY_PATH=/app/lib
./a.out
```

No error will happen. You can also replace the `/lib/libpulsar.so` with the `libpulsar.so` built from source.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -31,6 +31,7 @@
 #include "ProducerImpl.h"
 #include "PulsarApi.pb.h"
 #include "Url.h"
+#include "auth/InitialAuthData.h"
 #include "checksum/ChecksumProvider.h"
 
 DECLARE_LOG_OBJECT()
@@ -225,7 +226,8 @@ ClientConnection::ClientConnection(const std::string& logicalAddress, const std:
         std::string tlsCertificates = clientConfiguration.getTlsCertificateFilePath();
         std::string tlsPrivateKey = clientConfiguration.getTlsPrivateKeyFilePath();
 
-        AuthenticationDataPtr authData;
+        auto authData = std::dynamic_pointer_cast<AuthenticationDataProvider>(
+            std::make_shared<InitialAuthData>(clientConfiguration.getTlsTrustCertsFilePath()));
         if (authentication_->getAuthData(authData) == ResultOk && authData->hasDataForTls()) {
             tlsCertificates = authData->getTlsCertificates();
             tlsPrivateKey = authData->getTlsPrivateKey();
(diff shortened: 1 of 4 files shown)


BewareMyPower:
I opened a candidate release in my local env that includes this patch: https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/4112989933

shibd (approved):

michaeljmarshall (commented):
I am not familiar with C++, but this looks correct to me. Thanks for fixing it @BewareMyPower!

PR #206, opened 2023-02-28
[improve] Refactor client version format

RobertIndie:
### Motivation

Currently,  the `client_version ` only shows the version information. And clients in different languages use their own separate version numbers. This can lead to conflict. For example, the java client 2.10.2 uses the same value `2.10.2` as the C++ client 2.10.2. And C++ client 3.0.0 may conflict with the future java client 3.0.0. The information of `client_version` is incomplete. We could not determine what language(or specific library) the client uses. This raises inconvenience for debugging.

This PR adopts a new client version format for the C++ client.

Please see the discussion: https://lists.apache.org/thread/n59k537fhthjnzkfxtc2p4zk4l0cv3mp
### Modifications
* Change the client version value to `Pulsar-C++-vXXX`

### Verifying this change

This change added tests.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/Commands.cc
+++ b/lib/Commands.cc
@@ -266,7 +266,7 @@ SharedBuffer Commands::newConnect(const AuthenticationPtr& authentication, const
     BaseCommand cmd;
     cmd.set_type(BaseCommand::CONNECT);
     CommandConnect* connect = cmd.mutable_connect();
-    connect->set_client_version(PULSAR_VERSION_STR);
+    connect->set_client_version(std::string("Pulsar-C++-v") + PULSAR_VERSION_STR);
     connect->set_auth_method_name(authentication->getAuthMethodName());
     connect->set_protocol_version(ProtocolVersion_MAX);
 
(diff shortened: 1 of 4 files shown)


shibd (approved):

lib/Commands.cc
```diff
@@ -266,7 +266,7 @@ SharedBuffer Commands::newConnect(const AuthenticationPtr& authentication, const
     BaseCommand cmd;
     cmd.set_type(BaseCommand::CONNECT);
     CommandConnect* connect = cmd.mutable_connect();
-    connect->set_client_version(PULSAR_VERSION_STR);
+    connect->set_client_version(std::string("Pulsar-C++-v") + PULSAR_VERSION_STR);
```
shibd, line 269: change to "Pulsar-CPP-v"?

shibd:
For Node.js and Python clients, How do we set this parameter?

lib/Commands.cc
RobertIndie: The documentation prefers to use `C++` over `CPP`. I think `C++` is better for users. https://pulsar.apache.org/docs/next/client-libraries-cpp/

RobertIndie: Oh. I see that we are already using CPP here: https://github.com/apache/pulsar-client-cpp/blob/05807bdaf3a4341b22efd7c71f7b00c47cc31413/lib/HTTPLookupService.cc#L195

+1 for using CPP

RobertIndie:
> For Node.js and Python clients, How do we set this parameter?

I made a discussion here: https://lists.apache.org/thread/ldmtjrqvz6bz2flpdl89k9ogybcm9mj1
I think we need to expose the configuration of the clientVersion.

Issue #213, opened 2023-03-14
[Flaky Test] ConsumerTest.testAcknowledgeCumulativeWithPartition


BewareMyPower:
[example failure](https://github.com/apache/pulsar-client-cpp/actions/runs/4413485517/jobs/7734003726)

logs:

```
2023-03-14 08:13:56.475 INFO  [139660182783552] ProducerImpl:207 | [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-1, ] Created producer on broker [[::1]:42966 -> [::1]:6650]
/home/runner/work/pulsar-client-cpp/pulsar-client-cpp/tests/ConsumerTest.cc:307: Failure
Expected equality of these values:
  0
  tracker->size()
    Which is: 50
2023-03-14 08:13:58.792 INFO  [139660185906944] ProducerImpl:711 | Producer - [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-0, standalone-0-385] , [batching  = off]
2023-03-14 08:13:58.792 INFO  [139660185906944] ProducerImpl:711 | Producer - [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-1, standalone-0-386] , [batching  = off]
2023-03-14 08:13:58.792 WARN  [139660185906944] ConsumerImpl:150 | [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-1, t-sub, 1] Destroyed consumer which was not properly closed
2023-03-14 08:13:58.793 INFO  [139660185906944] ConsumerImpl:158 | [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-1, t-sub, 1] Closed consumer for race condition: 1
2023-03-14 08:13:58.793 WARN  [139660185906944] ConsumerImpl:150 | [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-0, t-sub, 0] Destroyed consumer which was not properly closed
2023-03-14 08:13:58.793 INFO  [139660185906944] ConsumerImpl:158 | [persistent://public/default/testAcknowledgeCumulativeWithPartition-1678781636-partition-0, t-sub, 0] Closed consumer for race condition: 0
2023-03-14 08:13:58.793 INFO  [139660185906944] ClientConnection:1200 | [[::1]:42966 -> [::1]:6650] Connection disconnected
[...]

PR #211, opened 2023-03-14
Fix Client::close might hang forever for a multi-topics consumer

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/209

### Motivation

There is a race condition when `Client::close` is called after a `MultiTopicsConsumerImpl` instance is destructed. e.g.

```c++
{
  Consumer consumer;
  // subscribe multiple topics or partitions
  client.subscribe(/* ... */, consumer);
}  // consumer will be destructed after going out of the scope here
client.close();
```

There is a chance that when `client.close()` is called first, the destructor of `consumer` is not called, so `client.consumers_` is not empty and `ClientImpl::closeAsync` will wait until the `MultiTopicsConsumerImpl::closeAsync` completes. Then, in its callback https://github.com/apache/pulsar-client-cpp/blob/242ad770f76655d92c5c1d1abfdee1319ba2ebdc/lib/MultiTopicsConsumerImpl.cc#L482

the original callback (`ClientImpl::handleClose`) will be skipped and `ClientImpl::closeAsync` will never complete.

It's also the root cause of the flaky `ConnectionFailTest`, we can see each time it stuck for 120 minutes, there must be the following log:

```
ClientImpl:589 | Closing Pulsar client with 0 producers and 1 consumers
```

When the test passed, the log was:

```
ClientImpl:589 | Closing Pulsar client with 0 producers and 0 consumers
```

### Modifications

In `MultiTopicsConsumerImpl::closeAsync`, do not capture `weakSelf` in the callback that is passed to `consumer->closeAsync`. Instead, capture the `weakSelf` in the wrapped `callback` and modify the `state_` field here. Since now `callback` (and `originalCallback`) will never be skipped, `ClientImpl::close` will not be blocked forever.

### Verifications

Run the `ConnectionFailTest` inside a `ubuntu:22.04` docker container until the following log appeared.

```
ClientImpl:589 | Closing Pulsar client with 0 producers and 1 consumers
```

We can see now the test still passed.

```
2023-03-14 06:51:39.201 INFO  [140309830422656] ClientImpl:553 | Closing Pulsar client with 0 producers and 1 consumers
2023-03-14 06:51:39.201 INFO  [140309830422656] ConsumerImpl:1188 | [persistent://public/default/test-connection-fail-51678776698-partition-4, reader-e0617a7614, 9] Closing consumer for topic persistent://public/default/test-connection-fail-51678776698-partition-4
[...]

PR #214, opened 2023-03-14
Fix flaky testAcknowledgeCumulativeWithPartition 

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/213

### Motivation

In `ConsumerTest.testAcknowledgeCumulativeWithPartition`, the last two
`MessageId`s are acknowledged because they are treated as the latest two
`MessageId`s of the two partitions. However, this assumption is false
because `MultiTopicsConsumerImpl` does not guarantee receiving messages
in a round robin way. Here is an example output when I added the debug
logs for the `MessageId` received.

```
(519,48,1,-1)
(519,49,1,-1)
(518,48,0,-1)
(518,49,0,-1)
```

We can see the last two `MessageId`s are both from partition 0.

### Modifications

Use an array `latestMsgIds` to record the latest `MessageId` of a given
partition in this test.

### DIFF
--- a/tests/ConsumerTest.cc
+++ b/tests/ConsumerTest.cc
@@ -19,6 +19,7 @@
 #include <gtest/gtest.h>
 #include <pulsar/Client.h>
 
+#include <array>
 #include <chrono>
 #include <ctime>
 #include <map>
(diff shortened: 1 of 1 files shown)


BewareMyPower:
This PR is based on https://github.com/apache/pulsar-client-cpp/pull/211 to avoid being affected by the very flaky `ConnectFailTest`, so let's wait #211 is merged first.

shibd (approved):

PR #228, opened 2023-03-23
[fix] Consumer batch receive will cause data loss.

shibd:
### Motivation

Currently, the batch receive method may cause data loss.

The root cause is, when `peekMsg` can't add to `messages`(The batch has satisfied the batch receive policy), it has actually been popped.

https://github.com/apache/pulsar-client-cpp/blob/ba7f59d2bb28b72fa6d1d01abada71090537b45e/lib/ConsumerImpl.cc#L652-L657

### Modifications

- Peek first, pop only when it can be added.

### Verifying this change

- Change `testBatchReceive` test to cover this case: 
Send 100 messages first, batch receive 10 messages at a time, and assert that all messages can be received.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -649,10 +649,11 @@ void ConsumerImpl::executeNotifyCallback(Message& msg) {
 void ConsumerImpl::notifyBatchPendingReceivedCallback(const BatchReceiveCallback& callback) {
     auto messages = std::make_shared<MessagesImpl>(batchReceivePolicy_.getMaxNumMessages(),
                                                    batchReceivePolicy_.getMaxNumBytes());
-    Message peekMsg;
-    while (incomingMessages_.pop(peekMsg, std::chrono::milliseconds(0)) && messages->canAdd(peekMsg)) {
-        messageProcessed(peekMsg);
-        Message interceptMsg = interceptors_->beforeConsume(Consumer(shared_from_this()), peekMsg);
+    Message msg;
+    while (incomingMessages_.peek(msg) && messages->canAdd(msg)) {
+        incomingMessages_.pop(msg);
+        messageProcessed(msg);
+        Message interceptMsg = interceptors_->beforeConsume(Consumer(shared_from_this()), msg);
         messages->add(interceptMsg);
     }
     auto self = get_shared_this_ptr();
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -650,8 +650,8 @@ void ConsumerImpl::notifyBatchPendingReceivedCallback(const BatchReceiveCallback
     auto messages = std::make_shared<MessagesImpl>(batchReceivePolicy_.getMaxNumMessages(),
                                                    batchReceivePolicy_.getMaxNumBytes());
     Message msg;
-    while (incomingMessages_.peek(msg) && messages->canAdd(msg)) {
-        incomingMessages_.pop(msg);
+    while (incomingMessages_.popIf(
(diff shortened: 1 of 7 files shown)


lib/ConsumerImpl.cc
```diff
@@ -649,10 +649,11 @@ void ConsumerImpl::executeNotifyCallback(Message& msg) {
 void ConsumerImpl::notifyBatchPendingReceivedCallback(const BatchReceiveCallback& callback) {
     auto messages = std::make_shared<MessagesImpl>(batchReceivePolicy_.getMaxNumMessages(),
                                                    batchReceivePolicy_.getMaxNumBytes());
-    Message peekMsg;
-    while (incomingMessages_.pop(peekMsg, std::chrono::milliseconds(0)) && messages->canAdd(peekMsg)) {
-        messageProcessed(peekMsg);
-        Message interceptMsg = interceptors_->beforeConsume(Consumer(shared_from_this()), peekMsg);
+    Message msg;
+    while (incomingMessages_.peek(msg) && messages->canAdd(msg)) {
+        incomingMessages_.pop(msg);
```
BewareMyPower, line 654: The combination of `peek` and `pop` is not atomic. So you have to use a lock to synchronize these two operations.

BewareMyPower (changes requested):

lib/ConsumerImpl.cc
shibd: I noticed that Java client has the same problem. 

https://github.com/apache/pulsar/blob/9ab9d0170a0a955875656d683839dd516c335bde/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerBase.java#L981-L999

> So you have to use a lock to synchronize these two operations.

But the lock feels a bit heavy because we also need to synchronize the take operation of `receive`. 

https://github.com/apache/pulsar/blob/e0b50c9ec5f12d0fb8275f235d8ac00e87a9099e/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L445-L446

The current implementation does not lose or duplicate data, that just may return batches that exceed `batchReceivePolicy.MaxNumBytes`.

BewareMyPower: Correctness is more important than performance. The race condition problem is hard to analysis if it happened. We don't have to copy everything from the Java client, including the bugs.

BTW, the `peek` and `pop` methods in C++ client already acquire the lock.

shibd: > The current implementation does not lose or duplicate data, that just may return batches that exceed batchReceivePolicy.MaxNumBytes.

I mean, if we don't think it's a bug, there's no need to lock it here.

Because we can't accurately control the batch returned, it must be equal to `batchReceivePolicy.MaxNumBytes`

- If we lock here, it is possible to return batches less than `batchReceivePolicy.MaxNumBytes`. 
- If we don't lock here, it is possible to return batches greater than `batchReceivePolicy.MaxNumBytes`.

We can discuss it first.

BewareMyPower: Don't assume any possible result caused by a race condition.

For example, what if `incomingMessages_` is empty when `pop` is called?

shibd: Make sense to me. 

I added the `popIf` method on the queue. This avoids adding locks in multiple places in the consumer(all pop, clear, peek call places). PTAL.

shibd:
> Should we also check `isClosedNoMutex` for both `BlockingQueue` and `UnboundedBlockingQueue`?

oh, I missed it. added check, PTAL.

BewareMyPower (approved):

PR #230, opened 2023-03-24
[feat] Support consumer batch receive on C client.

shibd:
### Motivation

Support consumer batch receive on C client.

### Modifications
- Add messages on C.
- Support batch receive and async batch receive API.

### Verifying this change

- Add `ConsumerTest.testCBatchReceive` to cover it.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/c/consumer_configuration.h
+++ b/include/pulsar/c/consumer_configuration.h
@@ -89,6 +89,15 @@ typedef enum
     pulsar_consumer_regex_sub_mode_AllTopics = 2
 } pulsar_consumer_regex_subscription_mode;
 
+typedef struct {
+    // Max num message, if less than 0, it means no limit.
+    int maxNumMessage;
+    // Max num bytes, if less than 0, it means no limit.
+    long maxNumBytes;
+    // If less than 0, it means no limit.
+    long timeoutMs;
+} pulsar_consumer_batch_receive_policy_t;
+
 /// Callback definition for MessageListener
 typedef void (*pulsar_message_listener)(pulsar_consumer_t *consumer, pulsar_message_t *msg, void *ctx);
 
@@ -326,6 +335,22 @@ PULSAR_PUBLIC void pulsar_consumer_configuration_set_regex_subscription_mode(
 
 PULSAR_PUBLIC pulsar_consumer_regex_subscription_mode
 pulsar_consumer_configuration_get_regex_subscription_mode(
+    pulsar_consumer_configuration_t *consumer_configuration);
+
+/**
+ * Set batch receive policy.
+ *
+ * The default value: {maxNumMessage: -1, maxNumBytes: 10 * 1024 * 1024, timeoutMs: 100}
+ * @param consumer_configuration  the consumer conf object.
+ * @param maxNumMessage default: Max num message, if less than 0, it means no limit.
+ * @param maxNumBytes Max num bytes, if less than 0, it means no limit.
+ * @param timeoutMs If less than 0, it means no limit.
+ */
+PULSAR_PUBLIC void pulsar_consumer_configuration_set_batch_receive_policy(
+    pulsar_consumer_configuration_t *consumer_configuration,
+    const pulsar_consumer_batch_receive_policy_t *batch_receive_policy);
+
+PULSAR_PUBLIC pulsar_consumer_batch_receive_policy_t pulsar_consumer_configuration_get_batch_receive_policy(
     pulsar_consumer_configuration_t *consumer_configuration);
 
 // const CryptoKeyReaderPtr getCryptoKeyReader()
--- a/include/pulsar/c/messages.h
+++ b/include/pulsar/c/messages.h
@@ -1,0 +1,62 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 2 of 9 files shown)


include/pulsar/c/consumer_configuration.h
```diff
@@ -304,6 +304,28 @@ PULSAR_PUBLIC void pulsar_consumer_configuration_set_start_message_id_inclusive(
...
+    long timeoutMs);
+
+PULSAR_PUBLIC int pulsar_consumer_configuration_get_batch_receive_policy_max_num_message(
+    pulsar_consumer_configuration_t *consumer_configuration);
+
+PULSAR_PUBLIC long pulsar_consumer_configuration_get_batch_receive_policy_max_num_bytes(
+    pulsar_consumer_configuration_t *consumer_configuration);
+
+PULSAR_PUBLIC long pulsar_consumer_configuration_get_batch_receive_policy_timeout_ms(
+    pulsar_consumer_configuration_t *consumer_configuration);
```
BewareMyPower, line 327: It might be better to define a struct that represents the batch receive policy to avoid too many setters.

```c
typedef struct {
    int maxNumMessage;
    int maxNumBytes;
    long timeoutMs;
} pulsar_batch_receive_policy_t;
```

shibd: With struct, do users need to manually delete that? 

```c
pulsar_batch_receive_policy_t  policy = {100, 100, 1000};
pulsar_consumer_configuration_set_batch_receive_policy(consumer, policy);

delete policy;

```

BewareMyPower: No. Only the memory allocated from the heap needs to be deleted.

BewareMyPower: BTW, since the size of `pulsar_batch_receive_policy_t` is greater than 8 bytes, it's better to pass the pointer to the function, i.e. the function accepts a const pointer argument.

shibd: If passing a pointer, users need to allocate/new this struct and manually free memory. 

```c
    pulsar_consumer_batch_receive_policy_t *batch_receive_policy;
    batch_receive_policy =
        (pulsar_consumer_batch_receive_policy_t *)malloc(sizeof(pulsar_consumer_batch_receive_policy_t));
    batch_receive_policy->maxNumMessage = 100;
    batch_receive_policy->maxNumBytes = -1;
    batch_receive_policy->timeoutMs = -1;
    pulsar_consumer_configuration_set_batch_receive_policy(consumer_conf, batch_receive_policy);
    delete batch_receive_policy;
```

Is this user friendly in C?

include/pulsar/c/messages.h
```diff
@@ -0,0 +1,62 @@
...
+ * under the License.
+ */
+
+#pragma once
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <pulsar/defines.h>
```
BewareMyPower, line 26:
```suggestion
#include <pulsar/defines.h>

#include "message.h"
```

Include `message.h` to have `pulsar_message_t`'s definition.

image

tests/c/c_ConsumerTest.cc
```diff
@@ -0,0 +1,78 @@
...
+        pulsar_message_t *message = pulsar_message_create();
+        pulsar_message_set_content(message, data, strlen(data));
+        pulsar_result res = pulsar_producer_send(producer, message);
+        ASSERT_EQ(pulsar_result_Ok, res);
+        pulsar_message_free(message);
+    }
+
+    // Batch receive messages
+    pulsar_messages_t *msgs = pulsar_messages_create();
+    pulsar_result res = pulsar_consumer_batch_receive(consumer, &msgs);
```
BewareMyPower, line 57: We don't need the `pulsar_messages_create` function for users. Unlike `pulsar_message_t`, which should be created by users when sending messages, `pulsar_messages_t` should only be created by `pulsar_consumer_batch_receive`.

```suggestion
    pulsar_messages_t *msgs = NULL;
    pulsar_result res = pulsar_consumer_batch_receive(consumer, &msgs);
    ASSERT_TRUE(msgs != NULL);
```

When `res` is `pulsar_result_OK`, it's guaranteed that `msgs` points to an object allocated by the library and should be deallocated by `pulsar_messages_free`.

With the current implementation, there is a memory leak that the instance created by `pulsar_messages_create` won't be released.

```bash
$ valgrind --leak-check=full ./tests/pulsar-tests --gtest_filter='C_ConsumerConfigurationTest.*'
...
==5249== LEAK SUMMARY:
==5249==    definitely lost: 360 bytes in 12 blocks
==5249==    indirectly lost: 12,336 bytes in 87 blocks
==5249==      possibly lost: 0 bytes in 0 blocks
```

BewareMyPower (changes requested):

tests/c/c_ConsumerTest.cc
BewareMyPower: Even with my suggested change, there is still a memory leak. The other methods might also have potential memory leaks.

shibd: > When res is pulsar_result_OK, it's guaranteed that msgs points to an object allocated by the library and should be deallocated by pulsar_messages_free.

With the current implementation, the user needs to manually free the `pulsar_messages_t`, is there anything wrong with this?
```
    pulsar_messages_t *msgs = pulsar_messages_create();
    pulsar_result res = pulsar_consumer_batch_receive(consumer, &msgs);
    // do something.
    pulsar_messages_free(msgs);
```

> With the current implementation, there is a memory leak that the instance created by pulsar_messages_create won't be released.

You seem to be executing `C_ConsumerConfigurationTest`, in this test, the configured memory is not freed, I added it.

BewareMyPower:
```c
    // 1. You allocated a `pulsar_messages_t` object via `pulsar_messages_created()` and
    //   let `msgs` points to this piece of memory
    pulsar_messages_t *msgs = pulsar_messages_create();
    // 2. After `pulsar_consumer_batch_receive`, `msgs` points to another piece of memory that
    //   is allocated inside `pulsar_consumer_batch_receive`.
    pulsar_result res = pulsar_consumer_batch_receive(consumer, &msgs);
    // 3. The memory allocated inside `pulsar_consumer_batch_receive` is deallocated.
    pulsar_messages_free(msgs);
    // NOTE: Now, the memory allocated by `pulsar_messages_created()` could never be deallocated!
```

PR #231, opened 2023-03-24
[fix] Use unique topic name in ConsumerInterceptorsTest

### LINKED ISSUE #229: Flaky test: ConsumerInterceptorsTest.TestNegativeAcksSend
[example failure](https://github.com/apache/pulsar-client-cpp/actions/runs/4501817436/jobs/7922762277?pr=228)

```
FAILED TESTS (5/488):
     264 ms: ./pulsar-tests Pulsar/ConsumerInterceptorsTest.TestNegativeAcksSend/0 (try #1)
     257 ms: ./pulsar-tests Pulsar/ConsumerInterceptorsTest.TestNegativeAcksSend/2 (try #1)
     220 ms: ./pulsar-tests Pulsar/ConsumerInterceptorsTest.TestNegativeAcksSend/2 (try #2)
     158 ms: ./pulsar-tests Pulsar/ConsumerInterceptorsTest.TestNegativeAcksSend/2 (try #3)
     175 ms: ./pulsar-tests Pulsar/ConsumerInterceptorsTest.TestNegativeAcksSend/2 (try #4)
```

RobertIndie:
Fixes #229

### Motivation

Using `time(nulltpr)` is not enough for generating the unique topic name. Especially when running a test for less than a second.

### Modifications

* Use `unique_str` to generate the topic name in ConsumerInterceptorsTest

### Verifying this change

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/InterceptorsTest.cc
+++ b/tests/InterceptorsTest.cc
@@ -21,6 +21,7 @@
 #include <pulsar/ConsumerInterceptor.h>
 #include <pulsar/ProducerInterceptor.h>
 
+#include <random>
 #include <utility>
 
 #include "HttpHelper.h"
@@ -272,8 +273,11 @@ class ConsumerTestInterceptor : public ConsumerInterceptor {
 class ConsumerInterceptorsTest : public ::testing::TestWithParam<std::tuple<TopicType, int>> {
    public:
     void SetUp() override {
-        topic_ = "persistent://public/default/InterceptorsTest-ConsumerInterceptors-" +
-                 std::to_string(time(nullptr));
+        std::random_device rd;
+        std::mt19937 mt(rd());
+        std::uniform_int_distribution<int> dist;
+        topic_ =
+            "persistent://public/default/InterceptorsTest-ConsumerInterceptors-" + std::to_string(dist(mt));
 
         switch (std::get<0>(GetParam())) {
             case Partitioned:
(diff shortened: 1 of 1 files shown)


tests/InterceptorsTest.cc
```diff
@@ -272,8 +273,11 @@ class ConsumerTestInterceptor : public ConsumerInterceptor {
 class ConsumerInterceptorsTest : public ::testing::TestWithParam<std::tuple<TopicType, int>> {
    public:
     void SetUp() override {
-        topic_ = "persistent://public/default/InterceptorsTest-ConsumerInterceptors-" +
-                 std::to_string(time(nullptr));
```
shibd, line 276: You can just splice `GetParam()`.

RobertIndie: Using `GetParam()` couldn't solve the issue. Multiple tests share the same params. If we want to splice `GetParam()`, we also need to add the `current test method` name to the test topic name. But this PR seems simpler.

BewareMyPower:
I think the root cause is that the time unit of `time()` is second.

IMO, you can reuse the `unique_str` function defined in `BasicEndToEndTest.cc` because the time unit is nanosecond. `AcknowledgeTest`, `MemoryLimitTest` and `TypedMessageTest` all reuse this function.

RobertIndie:
> I think the root cause is that the time unit of `time()` is second.

That's also correct.

 
> IMO, you can reuse the `unique_str` function defined in `BasicEndToEndTest.cc` because the time unit is nanosecond. `AcknowledgeTest`, `MemoryLimitTest` and `TypedMessageTest` all reuse this function.

Thanks for that information. `unique_str()` is a more elegant way to generate a unique topic.

BewareMyPower (approved):

Issue #102, opened 2022-11-03
[feature request] Support ACK response


BewareMyPower:
### Motivation

Catch up https://github.com/apache/pulsar/pull/8996. The `acknowledge` and `acknowledgeCumulative` methods should be synchronized APIs, but they don't wait until the ACK response is received. #8161 adds the ACK response at broker side introduced from 2.8.0 and v17 protocol

PR #232, opened 2023-03-24
Support waiting for the ACK response

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/102

### Modifications

Add a consumer configuration to set `ackReceiptEnabled`, if it's true, when sending a CommandAck request, generate the request id and set the `request_id` field so that the broker will respond with a `CommandAckResponse`. Here we pass the shared pointer to an atomic integer so that we no longer need to hold a weak pointer to `ClientImpl`.

Pass the user-defined callback to the ACK grouping tracker, when the `ackReceiptEnabled` is true, trigger the callback only after receiving the ACK response.

### Verifications

- Add `testAckReceiptEnabled` to verify when `ackReceiptEnabled` is true, the callback will be triggered after `ackGroupingMaxTime`.
- Support configuring `ackReceiptEnabled` for existing `TEST_P` parameterized tests in `AcknowledgeTest`.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/AckGroupingTracker.cc
+++ b/lib/AckGroupingTracker.cc
@@ -19,6 +19,9 @@
 
 #include "AckGroupingTracker.h"
 
+#include <atomic>
+#include <limits>
+
 #include "BitSet.h"
 #include "ClientConnection.h"
 #include "Commands.h"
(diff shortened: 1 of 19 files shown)


BewareMyPower:
Need to rebase after https://github.com/apache/pulsar-client-cpp/pull/231 is merged.

lib/Commands.cc
```diff
@@ -441,15 +442,16 @@ SharedBuffer Commands::newProducer(const std::string& topic, uint64_t producerId
 }
 
 SharedBuffer Commands::newAck(uint64_t consumerId, int64_t ledgerId, int64_t entryId, const BitSet& ackSet,
-                              CommandAck_AckType ackType, CommandAck_ValidationError validationError) {
+                              CommandAck_AckType ackType, uint64_t requestId) {
```
RobertIndie, line 445: The `ConsumerImpl. discardCorruptedMessage ` still passes the validationError to this function. And It will be incorrectly identified as the requestId here.
https://github.com/apache/pulsar-client-cpp/blob/14623ca6f5fbd84a4baf32a2c16673011d85f98c/lib/ConsumerImpl.cc#L828-L829

RobertIndie (changes requested):

lib/Commands.cc
BewareMyPower: Good find. I think we'd better to use two overloads of `Commands::newAck` to avoid the similar misuse.

lib/ConsumerImpl.cc
```diff
@@ -1089,12 +1105,13 @@ void ConsumerImpl::acknowledgeAsync(const MessageId& msgId, ResultCallback callb
     const auto& msgIdToAck = pair.first;
     const bool readyToAck = pair.second;
     if (readyToAck) {
-        ackGroupingTrackerPtr_->addAcknowledge(msgIdToAck);
+        ackGroupingTrackerPtr_->addAcknowledge(msgIdToAck, callback);
+    } else {
+        if (callback) {
```
shibd, line 1110: Why do direct callback when readyToAck is false? 

If batchIndexAck = false, and the batch message does not have ack all messages, here also needs to wait(if ackReceiptEnabled == true)?

lib/AckGroupingTracker.cc
```diff
@@ -62,25 +74,39 @@ static std::ostream& operator<<(std::ostream& os, const std::set<MessageId>& msg
...
+        if (waitResponse_) {
+            const auto requestId = requestIdSupplier_();
+            cnx->sendRequestWithId(Commands::newMultiMessageAck(consumerId_, msgIds, requestId), requestId)
+                .addListener([callback](Result result, const ResponseData&) {
+                    if (callback) {
+                        callback(result);
+                    }
+                });
+        } else {
+            cnx->sendCommand(Commands::newMultiMessageAck(consumerId_, msgIds));
```
shibd, line 97: Here also need to callback?

lib/ConsumerImpl.cc
BewareMyPower: Yes. It has the same behavior with the Java client. See https://github.com/apache/pulsar/blob/7a99e74220bdff7a0dff3dbbf3befb2316f3a6c8/pulsar-client/src/main/java/org/apache/pulsar/client/impl/PersistentAcknowledgmentsGroupingTracker.java#L214

Even if the actual CommandAck request is not sent, the shared bit set of these `BatchMessageIdImpl` instances has already been modified, which could be treated as acknowledged.

shibd (approved):

RobertIndie (approved):

Anonymitaet:
Hi @RobertIndie, 

Just now, I discussed with @BewareMyPower on how we add docs for this PR and summarized the thoughts below. He suggests that you're an expert in this area and can provide tech inputs for this PR. Are you available? Thanks!

========================================

Here is the doc plan:

## ✅ Pulsar feature matrix

I've created ["Wait for response" in Consumer > Ack](https://docs.google.com/spreadsheets/d/1YHYTkIXR8-Ql103u-IMI18TXLlGStK8uJjDsOOA0T20/edit#gid=1784579914&range=91:91) and added versions for Pulsar and C++.

## ✅ Pulsar client docs

1. Create a topic "Wait for response" in [Client > Work with consumer](https://pulsar.apache.org/docs/next/client-libraries-consumers/#acknowledge-messages), and add a simple feature description (this info can be fetched from [here](https://github.com/apache/pulsar-client-cpp/pull/232/files#diff-258280272b1786eac0ecaafc163e189b87faa2a63ab740225da0b459f528358aR633-R639)  and code examples (Java and C++).

image

2. This doc should be added to 5 doc versions (NEXT, 2.11.x, 2.10.x, 2.9.x, 2.8.x) since this feature is available from Pulsar 2.8.x. 

- Original doc file of NEXT is [here](https://github.com/apache/pulsar-site/edit/main/docs/client-libraries-tableviews.md).
- Original doc files of all versioned docs are [here](https://github.com/apache/pulsar-site/tree/main/versioned_docs).

We can update one version first and apply the same doc changes to the rest once it gets approved. 

========================================

Feel free to reach out to me if you have questions. TIA!

PR #236, opened 2023-03-29
Speed up the process to start standalone for tests

BewareMyPower:
### Motivation

It took too much time to start Pulsar standalone for tests because of the slow JVM startup when running `pulsar-admin` commands.

### Modifications

Use `curl` to send REST requests directly. After this change, the startup time reduced from 1m39.597s to 32.451s in my local env.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/build-support/start-mim-test-service-inside-container.sh
+++ b/build-support/start-mim-test-service-inside-container.sh
@@ -31,8 +31,24 @@ bin/pulsar tokens create \
             --secret-key file:///pulsar/data/tokens/secret.key \
             > /pulsar/data/tokens/token.txt
 
+# Unset the HTTP proxy to avoid the REST requests being affected
+export http_proxy=
+TOKEN=$(bin/pulsar tokens create \
+            --subject superUser \
+            --secret-key file:///pulsar/data/tokens/secret.key)
+
+# Create "standalone" cluster if it does not exist
+put() {
+    curl -H "Authorization: Bearer $TOKEN" \
+        -L http://localhost:8081/admin/v2/$1 \
+        -H 'Content-Type: application/json' \
+        -X PUT \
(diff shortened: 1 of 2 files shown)


RobertIndie (commented):
Is it better to use pulsar-shell instead of using raw HTTP calls?
We can use the no-interactive mode of the pulsar shell: https://pulsar.apache.org/docs/next/administration-pulsar-shell/#run-commands-sequentially

shibd (approved):

RobertIndie:
@BewareMyPower  Why do we need to download it separately? This script is running inside the docker. The pulsar after 2.11 already contains the `pulsar-shell`. We just need to use it like the `pulsar-admin`.

Using the `curl` would reduce the readability. It makes the script more complicate.

BewareMyPower:
When I tried to write shell scripts for `pulsar-shell`, I found it's hard to split it into two lines. For example, I met the following error:

```
[5/29] ? admin namespaces create public/default -c standalone
[6/29] Executing admin namespaces grant-permission
The following options are required: [--actions], [--role]
```

while my script is:

```bash
admin namespaces create public/default -c standalone
admin namespaces grant-permission \
    --actions produce,consume --role anonymous \
    public/default
```

The backslash seems not recognized, which is bad from my view.

BTW, I think the readability is not bad even with the raw REST APIs.
1. I wrote the comments for each REST call.
2. The `curl` command is encapsulated into a shell function `put <path> <body>`.

In addition, it can also be easily used to verify an older Pulsar image (< 2.11.0) to test the backward compatibility.

@RobertIndie

RobertIndie (approved):
Thanks for the explanation. Make sense to me.

PR #237, opened 2023-03-29
[feat] Support dead letter topic for C client.

shibd:
### Motivation

#114  Support dead letter topic for C client.

### Modifications

- Support pass dead letter topic policy by consumer config of C client.

### Verifying this change

- Add `testCDeadLetterTopic` to cover simple dlq case(other case cover by C++ test). 
- Enhance `testCApiConfig` to cover config set.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/c/consumer_configuration.h
+++ b/include/pulsar/c/consumer_configuration.h
@@ -369,7 +369,7 @@ PULSAR_PUBLIC pulsar_consumer_batch_receive_policy_t pulsar_consumer_configurati
 
 PULSAR_PUBLIC void pulsar_consumer_configuration_set_dlq_policy(
     pulsar_consumer_configuration_t *consumer_configuration,
-    pulsar_consumer_config_dead_letter_policy_t *dlq_policy);
+    const pulsar_consumer_config_dead_letter_policy_t *dlq_policy);
 
 /**
  * Get the dlq policy
--- a/include/pulsar/c/consumer_configuration.h
+++ b/include/pulsar/c/consumer_configuration.h
@@ -98,6 +98,21 @@ typedef struct {
     long timeoutMs;
 } pulsar_consumer_batch_receive_policy_t;
 
+typedef struct {
+    // Name of the dead topic where the failing messages are sent.
+    // The default value is: sourceTopicName + "-" + subscriptionName + "-DLQ"
+    const char *dead_letter_topic;
+    // Maximum number of times that a message is redelivered before being sent to the dead letter queue.
+    // - The maxRedeliverCount must be greater than 0.
+    // - The default value is {INT_MAX} (DLQ is not enabled)
+    const int max_redeliver_count;
+    // Name of the initial subscription name of the dead letter topic.
+    // If this field is not set, the initial subscription for the dead letter topic is not created.
+    // If this field is set but the broker's `allowAutoSubscriptionCreation` is disabled, the DLQ producer
+    // fails to be created.
+    const char *initial_subscription_name;
+} pulsar_consumer_config_dead_letter_policy_t;
+
 /// Callback definition for MessageListener
 typedef void (*pulsar_message_listener)(pulsar_consumer_t *consumer, pulsar_message_t *msg, void *ctx);
 
@@ -352,6 +367,13 @@ PULSAR_PUBLIC void pulsar_consumer_configuration_set_batch_receive_policy(
 
 PULSAR_PUBLIC pulsar_consumer_batch_receive_policy_t pulsar_consumer_configuration_get_batch_receive_policy(
     pulsar_consumer_configuration_t *consumer_configuration);
+
+PULSAR_PUBLIC void pulsar_consumer_configuration_set_dlq_policy(
+    pulsar_consumer_configuration_t *consumer_configuration,
+    pulsar_consumer_config_dead_letter_policy_t *dlq_policy);
+
+PULSAR_PUBLIC pulsar_consumer_config_dead_letter_policy_t
+pulsar_consumer_configuration_get_dlq_policy(pulsar_consumer_configuration_t *consumer_configuration);
 
 // const CryptoKeyReaderPtr getCryptoKeyReader()
 //
--- a/include/pulsar/c/consumer_configuration.h
+++ b/include/pulsar/c/consumer_configuration.h
@@ -100,14 +100,13 @@ typedef struct {
 
 typedef struct {
     // Name of the dead topic where the failing messages are sent.
-    // The default value is: sourceTopicName + "-" + subscriptionName + "-DLQ"
+    // If it's null, use sourceTopicName + "-" + subscriptionName + "-DLQ" as the value
     const char *dead_letter_topic;
     // Maximum number of times that a message is redelivered before being sent to the dead letter queue.
-    // - The maxRedeliverCount must be greater than 0.
-    // - The default value is {INT_MAX} (DLQ is not enabled)
-    const int max_redeliver_count;
+    // If it's not greater than 0, treat it as INT_MAX, it means DLQ disable.
+    int max_redeliver_count;
     // Name of the initial subscription name of the dead letter topic.
-    // If this field is not set, the initial subscription for the dead letter topic is not created.
+    // If it's null, the initial subscription for the dead letter topic is not created.
     // If this field is set but the broker's `allowAutoSubscriptionCreation` is disabled, the DLQ producer
     // fails to be created.
     const char *initial_subscription_name;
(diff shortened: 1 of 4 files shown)


lib/c/c_ConsumerConfiguration.cc
```diff
@@ -250,3 +251,36 @@ pulsar_consumer_regex_subscription_mode pulsar_consumer_configuration_get_regex_
...
+                                                  const char *initial_subscription_name) {
+    auto policyBuilder = pulsar::DeadLetterPolicyBuilder().maxRedeliverCount(max_redeliver_count);
+    if (dead_letter_topic != nullptr) {
+        policyBuilder.deadLetterTopic(dead_letter_topic);
+    }
+    if (initial_subscription_name != nullptr) {
+        policyBuilder.initialSubscriptionName(initial_subscription_name);
+    }
+    consumer_configuration->consumerConfiguration.setDeadLetterPolicy(policyBuilder.build());
+}
```
RobertIndie, line 266: It seems that this API isn't exposed at all. I don't think we need this API. The following API is enough.

RobertIndie (changes requested):

RobertIndie (approved):

include/pulsar/c/consumer_configuration.h
```diff
@@ -98,6 +98,21 @@ typedef struct {
...
+    // Maximum number of times that a message is redelivered before being sent to the dead letter queue.
+    // - The maxRedeliverCount must be greater than 0.
+    // - The default value is {INT_MAX} (DLQ is not enabled)
+    const int max_redeliver_count;
+    // Name of the initial subscription name of the dead letter topic.
+    // If this field is not set, the initial subscription for the dead letter topic is not created.
+    // If this field is set but the broker's `allowAutoSubscriptionCreation` is disabled, the DLQ producer
+    // fails to be created.
+    const char *initial_subscription_name;
+} pulsar_consumer_config_dead_letter_policy_t;
```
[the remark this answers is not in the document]
BewareMyPower: In addition, I think we'd better not to use raw struct. In C, the initial field values of a struct is not determined. For example,

```c
#include <stdio.h>
#include <stdlib.h>

typedef struct {
  int x;
  const char *s;
} A;

int main(int argc, char *argv[]) {
  A a;
  printf("%d | %s\n", a.x, a.s);
  return 0;
}
```

```bash
$ gcc 0.c -O0
$ ./a.out
891409248 | (null)
$ gcc 0.c -O2
$ ./a.out
0 | (null)
```

Unlike C++ (from 11), a struct could not have initial values. 

In addition, we'd better not to return a struct because its size could be greater than 8 bytes. POSIX APIs usually add a pointer argument as the output argument. Take [setrlimit/getrlimit](https://linux.die.net/man/2/setrlimit) for example:

```c
int getrlimit(int resource, struct rlimit *rlim);
int setrlimit(int resource, const struct rlimit *rlim);

struct rlimit {
    rlim_t rlim_cur;  /* Soft limit */
    rlim_t rlim_max;  /* Hard limit (ceiling for rlim_cur) */
};
```

These problems exist for `pulsar_consumer_batch_receive_policy_t` as well. I will push a PR to fix it.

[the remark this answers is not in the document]
BewareMyPower: Regarding the default value, I just thought again and now I think it's okay to use raw struct types but we must document it well, especially for the **default values**. For example,

```c
typedef struct {
    // ...
    // If it's null, use sourceTopicName + "-" + subscriptionName + "-DLQ" as the value
    const char *dead_letter_topic;
    // ...
    // If it's not greater than 0, treat it as INT_MAX
    const int max_redeliver_count;
    // ...
    // If it's null, the initial subscription for the dead letter topic is not created.
    const char *initial_subscription_name;
} pulsar_consumer_config_dead_letter_policy_t;
```

i.e. a `NULL` string and a zero integer are treated as special default values. And the caller should set all these fields explicitly.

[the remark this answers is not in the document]
shibd: > In addition, I think we'd better not to use raw struct. In C, the initial field values of a struct is not determined. For example

emm, I agree with your case. But, what is a better example of this case?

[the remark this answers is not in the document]
shibd: > i.e. a NULL string and a zero integer are treated as special default values. And the caller should set all these fields explicitly.

I agree, Users need to set these fields, and we do not promise the behavior of not setting these fields, which should be the consensus of C programming.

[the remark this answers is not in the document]
BewareMyPower: > But, what is a better example of this case?

I checked some POSIX API designs. There are two typical ways.

1. Require users to set all fields explicitly.
2. Use the getter to initialize the current struct.

The `setrlimit`/`getrlimit` pair is an example.

```c
// Set two customized values
struct rlimit rl;
rl.rlim_cur = 100;
rl.rlim_max = 1024;
setrlimit(RLIMIT_XXX, &rl);
```

or

```c
// Get the default values first, then only modify one field
struct rlimit rl;
getrlimit(RLIMIT_XXX, &rl);
rl.rlim_max = 1024;
setrlimit(RLIMIT_XXX, &rl);
```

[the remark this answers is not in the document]
BewareMyPower: If you prefer the OOP style, you can also add a function like:

```c
struct A {
    int x;
    const char* s;
};

void init_a(struct A* a) {
    a.x = 0;
    a.s = NULL;
}
```

Then it requires users to use `init_a` to initialize struct `A`. The `init_a` function acts like C++'s constructor.

BTW, I think all of them are okay.

include/pulsar/c/consumer_configuration.h
```diff
@@ -353,6 +367,22 @@ PULSAR_PUBLIC void pulsar_consumer_configuration_set_batch_receive_policy(
 PULSAR_PUBLIC pulsar_consumer_batch_receive_policy_t pulsar_consumer_configuration_get_batch_receive_policy(
     pulsar_consumer_configuration_t *consumer_configuration);
 
+PULSAR_PUBLIC void pulsar_consumer_configuration_set_dlq_policy(
+    pulsar_consumer_configuration_t *consumer_configuration,
+    pulsar_consumer_config_dead_letter_policy_t *dlq_policy);
```
BewareMyPower, line 372:
```suggestion
    const pulsar_consumer_config_dead_letter_policy_t *dlq_policy);
```

tests/c/c_ConsumerConfigurationTest.cc
```diff
@@ -19,6 +19,8 @@
 #include <gtest/gtest.h>
 #include <pulsar/c/consumer_configuration.h>
 
+#include "climits"
```
BewareMyPower, line 22:
```suggestion
#include <climits>
```

tests/c/c_ConsumerTest.cc
```diff
@@ -121,3 +121,81 @@ TEST(c_ConsumerTest, testBatchReceive) {
     pulsar_client_free(client);
     pulsar_client_configuration_free(conf);
 }
+
+TEST(C_ConsumerConfigurationTest, testCDeadLetterTopic) {
```
BewareMyPower, line 125:
```suggestion
TEST(C_ConsumerTest, testCDeadLetterTopic) {
```

tests/c/c_ConsumerTest.cc
```diff
@@ -121,3 +121,81 @@ TEST(c_ConsumerTest, testBatchReceive) {
...
+        pulsar_message_free(message);
+    }
+
+    // Consumer dlq topic
+    pulsar_consumer_t *dlq_consumer;
+    pulsar_consumer_configuration_t *dlq_consumer_conf = pulsar_consumer_configuration_create();
+    result = pulsar_client_subscribe(client, dlq_topic_name, sub_name, dlq_consumer_conf, &dlq_consumer);
+    ASSERT_EQ(pulsar_result_Ok, result);
+    for (int i = 0; i < num; ++i) {
+        pulsar_message_t *message = pulsar_message_create();
```
BewareMyPower, line 176:
```suggestion
        pulsar_message_t *message = NULL;
```

Fix the memory leak. `message` will pointer to another instance after `pulsar_consumer_receive`.

tests/c/c_ConsumerTest.cc
```diff
@@ -121,3 +121,81 @@ TEST(c_ConsumerTest, testBatchReceive) {
...
+        pulsar_message_set_content(message, data, strlen(data));
+        pulsar_result res = pulsar_producer_send(producer, message);
+        ASSERT_EQ(pulsar_result_Ok, res);
+        pulsar_message_free(message);
+    }
+
+    // Redelivery all messages
+    for (int i = 1; i <= max_redeliver_count * num + num; ++i) {
+        pulsar_message_t *message = pulsar_message_create();
```
BewareMyPower, line 161:
```suggestion
        pulsar_message_t *message = NULL;
```

tests/c/c_ConsumerTest.cc
```diff
@@ -121,3 +121,81 @@ TEST(c_ConsumerTest, testBatchReceive) {
...
+    pulsar_message_t *message = pulsar_message_create();
+    pulsar_result res = pulsar_consumer_receive_with_timeout(dlq_consumer, &message, 200);
+    ASSERT_EQ(pulsar_result_Timeout, res);
+    pulsar_message_free(message);
+
+    ASSERT_EQ(pulsar_result_Ok, pulsar_consumer_unsubscribe(consumer));
+    ASSERT_EQ(pulsar_result_AlreadyClosed, pulsar_consumer_close(consumer));
+    ASSERT_EQ(pulsar_result_Ok, pulsar_consumer_unsubscribe(dlq_consumer));
+    ASSERT_EQ(pulsar_result_AlreadyClosed, pulsar_consumer_close(dlq_consumer));
+    ASSERT_EQ(pulsar_result_Ok, pulsar_producer_close(producer));
```
BewareMyPower, line 190:
```suggestion
```

They can be removed. We only need to close the client.

BTW, `close` returns `AlreadyClosed` after `unsubscribe` is a bug. We don't need to protect this wrong behavior currently.

BewareMyPower (changes requested):

BewareMyPower (approved):

Issue #241, opened 2023-04-02
[Bug][KeySharedPolicy] bug where KeySharedPolicy::setStickyRanges duplicated ranges.


hyperevo:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

OS: 5.10.157-1-MANJARO #1 SMP PREEMPT Fri Dec 2 21:02:47 UTC 2022 x86_64 GNU/Linux
Pulsar version: 3.2.0-pre

### Minimal reproduce step

Run the following unit test:

```
TEST_F(KeySharedPolicyTest, testStickyConsumerExpected) {
    // Test whether the saved range vector is as expected
    KeySharedPolicy ksp;
    ksp.setStickyRanges({StickyRange(0, 300), StickyRange(400, 500)});

    std::vector<std::pair<int, int>> expectedStickyRange {{0, 300}, {400, 500}};
    ASSERT_EQ(ksp.getStickyRanges(),expectedStickyRange);
}
```

### What did you expect to see?

I expect ksp->ranges to be a vector of the form { (0, 300), (400, 500) }

### What did you see instead?

I see a vector of the form { (0, 300), (400, 500), (0, 300), (400, 500) }. 

### Anything else?

The cause is that this for loop is nested when it shouldn't be. https://github.com/apache/pulsar-client-cpp/blob/86d66bdb09bdb596a2a5c25b0a09f3d0182d683e/lib/KeySharedPolicy.cc#L68

I will make a fix and add some tests and put it in a pr.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

tisonkun:
Fixed by https://github.com/apache/pulsar-client-cpp/pull/242

PR #242, opened 2023-04-02
[Bug Fix][KeySharedPolicy] Fixed bug where KeySharedPolicy::setStickyRanges duplicated ranges.

hyperevo:
*[fix] Fixed bug where KeySharedPolicy::setStickyRanges appended duplicate ranges to KeySharedPolicy->ranges. The for loop was nested when it shouldn't have been nested. 

This fixes the issue #241 

*Added a test that checks to see if the KeySharedPolicy->ranges variable is set as expected after calling KeySharedPolicy::setStickyRanges

*Added overloaded KeySharedPolicy::setStickyRanges function that takes the StickyRanges vector. This is convenient in case a developer wants to provide a vector. It is also required for adding KeySharedPolicy to the python pulsar client which uses pybind11. std::initializer_list type conversion is not supported with pybind11. See https://github.com/pybind/pybind11/issues/1302#issuecomment-369090893
I am in the process of adding support for KeySharedPolicy to the python-pulsar-client and will submit a pr soon.

*Added a test that checks to see if the KeySharedPolicy->ranges variable is the same when using the two different overloaded KeySharedPolicy::setStickyRanges functions.

### Modifications

Moved for loop in KeySharedPolicy::setStickyRanges to unnested location to eliminate duplicate vector entries.

Added simple overloaded version of the same function that a parameter of type StickyRanges in addition to existing function that takes a parameter of type std::initializer_list<StickyRange>

Added two new tests in KeySharedPolicyTest.cc to validate the changes.

### Verifying this change

Added two new tests in KeySharedPolicyTest.cc to validate the changes.

### Documentation

- [ ] `doc-required` 

- [ *] `doc-not-needed` 
(Please explain why)
Bug fix did not change functionality in any new ways that are not covered by existing documentation.

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/KeySharedPolicy.h
+++ b/include/pulsar/KeySharedPolicy.h
@@ -99,6 +99,12 @@ class PULSAR_PUBLIC KeySharedPolicy {
     KeySharedPolicy& setStickyRanges(std::initializer_list<StickyRange> ranges);
 
     /**
+     * @param ranges used with sticky mode
+     */
+    KeySharedPolicy& setStickyRanges(StickyRanges ranges);
+
+
+    /**
      * @return ranges used with sticky mode
      */
     StickyRanges getStickyRanges() const;
--- a/lib/KeySharedPolicy.cc
+++ b/lib/KeySharedPolicy.cc
@@ -51,7 +51,16 @@ KeySharedPolicy &KeySharedPolicy::setAllowOutOfOrderDelivery(bool allowOutOfOrde
 bool KeySharedPolicy::isAllowOutOfOrderDelivery() const { return impl_->allowOutOfOrderDelivery; }
 
 KeySharedPolicy &KeySharedPolicy::setStickyRanges(std::initializer_list<StickyRange> ranges) {
-    if (ranges.size() == 0) {
+    StickyRanges v;
+    for(StickyRange i:ranges)
+    {
+        v.push_back(i);
+    }
+    return this->setStickyRanges(v);
+}
+
+KeySharedPolicy &KeySharedPolicy::setStickyRanges(StickyRanges ranges) {
+    if (ranges.empty()) {
         throw std::invalid_argument("Ranges for KeyShared policy must not be empty.");
     }
     for (StickyRange range : ranges) {
@@ -65,9 +74,9 @@ KeySharedPolicy &KeySharedPolicy::setStickyRanges(std::initializer_list<StickyRa
                 throw std::invalid_argument("Ranges for KeyShared policy with overlap.");
             }
         }
-        for (StickyRange range : ranges) {
-            impl_->ranges.push_back(range);
-        }
+    }
+    for (StickyRange range : ranges) {
+        impl_->ranges.push_back(range);
     }
     return *this;
 }
--- a/tests/KeySharedPolicyTest.cc
+++ b/tests/KeySharedPolicyTest.cc
@@ -201,3 +201,25 @@ TEST_F(KeySharedPolicyTest, InvalidStickyRanges) {
     ASSERT_THROW(ksp.setStickyRanges({StickyRange(0, 65536)}), std::invalid_argument);
     ASSERT_THROW(ksp.setStickyRanges({StickyRange(0, 10), StickyRange(9, 20)}), std::invalid_argument);
 }
+
+TEST_F(KeySharedPolicyTest, testStickyConsumerExpected) {
+    // Test whether the saved range vector is as expected
+    KeySharedPolicy ksp;
+    ksp.setStickyRanges({StickyRange(0, 300), StickyRange(400, 500)});
+
+    std::vector<std::pair<int, int>> expectedStickyRange {{0, 300}, {400, 500}};
+    ASSERT_EQ(ksp.getStickyRanges(),expectedStickyRange);
+}
+
+TEST_F(KeySharedPolicyTest, testStickyConsumerVectors) {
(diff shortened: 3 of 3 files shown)


hyperevo:
Here is the pulsar python client pr that is dependent on this pr https://github.com/apache/pulsar-client-python/pull/109

shibd:
@hyperevo Please fix the code format.

You can use `./build-support/docker-format.sh`

include/pulsar/KeySharedPolicy.h
```diff
@@ -98,6 +98,12 @@ class PULSAR_PUBLIC KeySharedPolicy {
      */
     KeySharedPolicy& setStickyRanges(std::initializer_list<StickyRange> ranges);
```
BewareMyPower, line 99: The method that accepts an `initializer_list` is not a good API. I think we should mark it as deprecated.

include/pulsar/KeySharedPolicy.h
```diff
@@ -98,6 +98,12 @@ class PULSAR_PUBLIC KeySharedPolicy {
...
 
+    /**
+     * @param ranges used with sticky mode
+     */
+    KeySharedPolicy& setStickyRanges(StickyRanges ranges);
```
BewareMyPower, line 104: Use `const StickyRanges&` instead of `StickyRanges`

lib/KeySharedPolicy.cc
```diff
@@ -51,7 +51,16 @@ KeySharedPolicy &KeySharedPolicy::setAllowOutOfOrderDelivery(bool allowOutOfOrde
 bool KeySharedPolicy::isAllowOutOfOrderDelivery() const { return impl_->allowOutOfOrderDelivery; }
 
 KeySharedPolicy &KeySharedPolicy::setStickyRanges(std::initializer_list<StickyRange> ranges) {
-    if (ranges.size() == 0) {
+    StickyRanges v;
+    for(StickyRange i:ranges)
+    {
+        v.push_back(i);
+    }
+    return this->setStickyRanges(v);
```
BewareMyPower, line 59: I think we can convert a `std::initializer_list<T>` to a `std::vector<T>` implicitly.

BewareMyPower (changes requested):
Please fix the code format as well

include/pulsar/KeySharedPolicy.h
hyperevo: Done.

include/pulsar/KeySharedPolicy.h
hyperevo: Done.

lib/KeySharedPolicy.cc
hyperevo: Done.

hyperevo:
Thanks for reviewing! I made the changes requested and ran ./build-support/docker-format.sh to format correctly in the latest commit.

BewareMyPower:
Please fix the build error:

```
/home/runner/work/pulsar-client-cpp/pulsar-client-cpp/lib/KeySharedPolicy.cc: In member function ‘pulsar::KeySharedPolicy& pulsar::KeySharedPolicy::setStickyRanges(std::initializer_list<std::pair<int, int> >)’:
/home/runner/work/pulsar-client-cpp/pulsar-client-cpp/lib/KeySharedPolicy.cc:54:47: error: expected primary-expression before ‘v’
   54 |     return this->setStickyRanges(StickyRanges v(ranges));
```

.gitignore
```diff
@@ -69,6 +69,7 @@ include/pulsar/Version.h
 *.ninja*
 .clangd/
 compile_commands.json
+.cmake-build-debug/
```
BewareMyPower, line 72: Please revert this unrelated change. The cmake build directory can be any directory.

lib/KeySharedPolicy.cc
```diff
@@ -51,7 +51,11 @@ KeySharedPolicy &KeySharedPolicy::setAllowOutOfOrderDelivery(bool allowOutOfOrde
 bool KeySharedPolicy::isAllowOutOfOrderDelivery() const { return impl_->allowOutOfOrderDelivery; }
 
 KeySharedPolicy &KeySharedPolicy::setStickyRanges(std::initializer_list<StickyRange> ranges) {
-    if (ranges.size() == 0) {
+    return this->setStickyRanges(v(ranges));
```
BewareMyPower, line 54: > This should be fixed now.

Actually it's not fixed. The `v(ranges)` here is not a valid C++ semantic.

BewareMyPower (changes requested):

lib/KeySharedPolicy.cc
hyperevo: Oops. Deleted the wrong thing. I just fixed this and recommitted.

.gitignore
hyperevo: Anyways. I have removed this now.

hyperevo: note: cmake-build-debug is a default directory that Jetbrains Clion IDE creates when the project is compiled.

BewareMyPower: > Is this really worth worrying about?

Not a serious problem. But it's better to explain why did you add a new item into the `.gitignore`.

> note: cmake-build-debug is a default directory that the Jetbrains Clion IDE creates when the project is compiled.

And I agree that this explanation is reasonable.

BewareMyPower (approved):

PR #253, opened 2023-04-17
PIP-254: Support configuring client version with a description suffix

BewareMyPower:
### Motivation

The C++ catch up of [PIP-254](https://github.com/apache/pulsar/issues/19705)

### Modifications

Different from the Java implementation, this PR adds private methods to set or get the description in `ClientConfiguration`. Since `PulsarWrapper` is used as a friend class of nearly all public classes, the 3rd party library that is based on the C++ SDK can set the description by defining a `PulsarWrapper` class.

For the `ConnectionPool` and any `ClientConnection` instance created, store the client version and pass it into `Commands::newConnect`.

Modify the `ClientTest.testClientVersion` to verify the changes.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ConnectionPool.h
+++ b/lib/ConnectionPool.h
@@ -41,7 +41,8 @@ using ExecutorServiceProviderPtr = std::shared_ptr<ExecutorServiceProvider>;
 class PULSAR_PUBLIC ConnectionPool {
    public:
     ConnectionPool(const ClientConfiguration& conf, ExecutorServiceProviderPtr executorProvider,
-                   const AuthenticationPtr& authentication, bool poolConnections = true);
+                   const AuthenticationPtr& authentication, bool poolConnections,
+                   const std::string& clientVersion);
 
     /**
      * Close the connection pool.
@@ -80,6 +81,7 @@ class PULSAR_PUBLIC ConnectionPool {
     typedef std::map<std::string, ClientConnectionWeakPtr> PoolMap;
     PoolMap pool_;
     bool poolConnections_;
+    const std::string clientVersion_;
     mutable std::mutex mutex_;
     std::atomic_bool closed_{false};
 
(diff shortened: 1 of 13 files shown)


lib/ConnectionPool.h
```diff
@@ -41,7 +41,8 @@ using ExecutorServiceProviderPtr = std::shared_ptr<ExecutorServiceProvider>;
 class PULSAR_PUBLIC ConnectionPool {
    public:
     ConnectionPool(const ClientConfiguration& conf, ExecutorServiceProviderPtr executorProvider,
-                   const AuthenticationPtr& authentication, bool poolConnections = true);
+                   const AuthenticationPtr& authentication, bool poolConnections,
```
shibd, line 44: Why delete the default value? Set default values can reduce incoming parameters.

BewareMyPower: Because I added a parameter `clientVersion` at the tail. The default parameter not at the tail is meaningless.

Additional, in C++, a default parameter works by generating one more function symbol, it's better not to add a default value just for tests. In the `lib/` subdirectory, the constructor of `ConnectionPool` is only called in `ClientImpl`'s constructor, so we only need one constructor.

shibd: So, unless there is a special requirement for a long parameter that requires a default value, in other cases, it is better not to set a default value?

BewareMyPower: IMO, the default arguments feature is used when the overloads can be simplified. For example, assuming we'd have two overloads:

```c++
// Overload 1
void f(int x, int y) { /* ... */ }
// Overload 2
void f(int x) { f(x, 0); }
```

Then we can simplify them to:

```c++
void f(int x, int y = 0) { /* ... */ }
```

It would be more clear.

However, in the case here, we don't need any overload as I've explained before:

> In the lib/ subdirectory, the constructor of ConnectionPool is only called in ClientImpl's constructor, so we only need one constructor.

If the overloads are only used for tests, we should not add them.

RobertIndie: +1 for not using default value here.

RobertIndie (approved):

shibd (approved):

PR #256, opened 2023-04-20
Improve the C APIs for batch receive policy and fix the wrong documents

BewareMyPower:
### Motivation

Currently, the `pulsar_consumer_configuration_get_batch_receive_policy` function returns a struct whose size is greater than 8, which is generally not acceptable in C API design because of the copy overhead.

And the API documents are wrong that they say if the value is less than 0, it means no limit. However, if the value is exactly 0, it also means no limit.

### Modifications

Use the POSIX style for these C APIs. The setter returns -1 if it failed. The getter accepts an output argument to store the value to get.

Then test these changes in `testCApiConfig`. The docs are also fixed.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/c/consumer_configuration.h
+++ b/include/pulsar/c/consumer_configuration.h
@@ -89,12 +89,14 @@ typedef enum
     pulsar_consumer_regex_sub_mode_AllTopics = 2
 } pulsar_consumer_regex_subscription_mode;
 
+// Though any field could be non-positive, if all of them are non-positive, this policy will be treated as
+// invalid
 typedef struct {
-    // Max num message, if less than 0, it means no limit.
-    int maxNumMessage;
-    // Max num bytes, if less than 0, it means no limit.
+    // Max num messages, a non-positive value means no limit.
+    int maxNumMessages;
+    // Max num bytes, a non-positive value means no limit.
     long maxNumBytes;
-    // If less than 0, it means no limit.
+    // The receive timeout, a non-positive value means no limit.
     long timeoutMs;
 } pulsar_consumer_batch_receive_policy_t;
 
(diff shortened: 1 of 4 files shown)


include/pulsar/c/consumer_configuration.h
```diff
@@ -89,12 +89,14 @@ typedef enum
...
 } pulsar_consumer_regex_subscription_mode;
 
+// Though any field could be non-positive, if all of them are non-positive, this policy will be treated as
+// invalid
 typedef struct {
-    // Max num message, if less than 0, it means no limit.
-    int maxNumMessage;
-    // Max num bytes, if less than 0, it means no limit.
+    // Max num messages, a non-positive value means no limit.
+    int maxNumMessages;
```
shibd, line 96: Refer to your comment on #237 

> In addition, I think we'd better not to use raw struct. In C, the initial field values of a struct is not determined. 

If users do not set this field, then this field is an undefined value.

For examples:
```
pulsar_consumer_batch_receive_policy_t batch_receive_policy{.maxNumeBytes = 1000}
pulsar_consumer_configuration_set_batch_receive_policy(consumer_conf, &batch_receive_policy);
```
The `maxNumMessages` and `timeoutMs` is an indeterminate value.

I think about it. For C users, he should know this behavior about undefined values in the struct.

So, We don't provide a default value to users, and we just to declaration is clear what the behavior is when the value is `xxx`.

Do you agree with me?

shibd (approved):

include/pulsar/c/consumer_configuration.h
BewareMyPower: Yes, see my additional explanations here:
- https://github.com/apache/pulsar-client-cpp/pull/237#discussion_r1172743122
- https://github.com/apache/pulsar-client-cpp/pull/237#discussion_r1172745690

BewareMyPower:
It's weird there is no helpful error prompt. I reran the CI first.

BewareMyPower:
I reverted this PR and could still reproduce the build failure in my local env.

```
/home/xyz/pulsar-client-cpp/build/lib/libpulsar.so: file not recognized: file format not recognized
collect2: error: ld returned 1 exit status
make[2]: *** [tests/CMakeFiles/ConnectionFailTest.dir/build.make:129: tests/ConnectionFailTest] Error 1
make[1]: *** [CMakeFiles/Makefile2:734: tests/CMakeFiles/ConnectionFailTest.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
...
[ 87%] Linking CXX executable pulsar-tests
[ 87%] Built target pulsar-tests
make: *** [Makefile:136: all] Error 2
```

It seems related to compiling with 8 threads. Maybe we need some further improvements for the build process.

shibd (approved):

PR #257, opened 2023-04-25
Support get the SchemaInfo from a topic and the schema version

BewareMyPower:
### Motivation

Currently there is no public API to get the SchemaInfo from the broker. However, when a consumer tries to decode a message of Avro schema, the writer schema that was stored in the broker is required. The C++ client is not responsible to decode the message, but the extended library like the Python client needs the ability to get the writer schema from the broker. There is a workaround that we can send a REST request to query the schema, but it requires the admin permission.

### Modification

- Add the `Client::getSchemaInfoAsync` method to get the schema info asynchronously. When the schema of the given version does not exist, the `SchemaInfo` whose type is `NONE` will be returned.
- Add the `Message::getLongSchemaVersion` method to get the schema version of a message. The existing `getSchemaVersion` method is hard to use because it returns a byte array, which users need to know how to decode.
- Provide `fromBigEndianBytes` and `toBigEndianBytes` functions to perform conversion between a byte array and the long value of the schema version. Add `Int64SerDes` to test them.
- Fix the `LookupServiceTest` that initializes `client_` only once (even with the `GetParam()` method) so the HTTP URL was never tested.
- Add `testGetSchemaByVersion` to test `getSchemaInfoAsync`.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Client.h
+++ b/include/pulsar/Client.h
@@ -408,11 +408,8 @@ class PULSAR_PUBLIC Client {
      * Asynchronously get the SchemaInfo of a topic and a specific version.
      *
      * @topic the topic name
-     * @version the schema version byte array, see Message::getLongSchemaVersion.
-     * @callback the callback that is triggered when the SchemaInfo is retrieved successfully or not.
-     *
-     * NOTE: If there is no schema registered or the given version does not exist, a SchemaInfo whose type is
-     * SchemaType::NONE will be returned in the callback.
+     * @version the schema version, see Message::getLongSchemaVersion
+     * @callback the callback that is triggered when the SchemaInfo is retrieved successfully or not
      */
     void getSchemaInfoAsync(const std::string& topic, int64_t version,
                             std::function<void(Result, const SchemaInfo&)> callback);
--- a/include/pulsar/Client.h
+++ b/include/pulsar/Client.h
@@ -404,6 +404,19 @@ class PULSAR_PUBLIC Client {
      */
     uint64_t getNumberOfConsumers();
 
+    /**
+     * Asynchronously get the SchemaInfo of a topic and a specific version.
+     *
+     * @topic the topic name
+     * @version the schema version byte array, see Message::getLongSchemaVersion.
+     * @callback the callback that is triggered when the SchemaInfo is retrieved successfully or not.
+     *
+     * NOTE: If there is no schema registered or the given version does not exist, a SchemaInfo whose type is
+     * SchemaType::NONE will be returned in the callback.
+     */
+    void getSchemaInfoAsync(const std::string& topic, int64_t version,
+                            std::function<void(Result, const SchemaInfo&)> callback);
+
    private:
     Client(const std::string& serviceUrl, const ClientConfiguration& clientConfiguration,
            bool poolConnections);
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -177,7 +177,15 @@ class PULSAR_PUBLIC Message {
     bool hasSchemaVersion() const;
 
     /**
+     * Get the schema version.
+     *
+     * @return the the schema version on success or -1 if the message does not have the schema version
+     */
+    int64_t getLongSchemaVersion() const;
+
+    /**
      * Get the schema version
+     * @deprecated Use getLongSchemaVersion instead
      */
     const std::string& getSchemaVersion() const;
 
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -184,8 +184,7 @@ class PULSAR_PUBLIC Message {
     int64_t getLongSchemaVersion() const;
 
     /**
-     * Get the schema version
-     * @deprecated Use getLongSchemaVersion instead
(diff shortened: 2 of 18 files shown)


include/pulsar/Message.h
```diff
@@ -176,8 +176,16 @@ class PULSAR_PUBLIC Message {
...
+    /**
+     * Get the schema version.
+     *
+     * @return the the schema version on success or -1 if the message does not have the schema version
+     */
+    int64_t getLongSchemaVersion() const;
+
     /**
      * Get the schema version
+     * @deprecated Use getLongSchemaVersion instead
```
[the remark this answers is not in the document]
BewareMyPower: Okay, let's not mark it as deprecated in this PR.

include/pulsar/Message.h
```diff
@@ -177,7 +177,14 @@ class PULSAR_PUBLIC Message {
     bool hasSchemaVersion() const;
 
     /**
-     * Get the schema version
+     * Get the schema version.
+     *
+     * @return the the schema version on success or -1 if the message does not have the schema version
+     */
+    int64_t getLongSchemaVersion() const;
```
shibd, line 184: For the Java client, it returns a bytes array. This may be a consensus that is already acceptable to users, is it necessary for us to change it to int?

include/pulsar/Client.h
```diff
@@ -404,6 +404,19 @@ class PULSAR_PUBLIC Client {
      */
     uint64_t getNumberOfConsumers();
 
+    /**
+     * Asynchronously get the SchemaInfo of a topic and a specific version.
+     *
+     * @topic the topic name
+     * @version the schema version byte array, see Message::getLongSchemaVersion.
```
shibd, line 411: This note does not match the parameter type.

include/pulsar/Client.h
```diff
@@ -404,6 +404,19 @@ class PULSAR_PUBLIC Client {
...
+     * @callback the callback that is triggered when the SchemaInfo is retrieved successfully or not.
+     *
+     * NOTE: If there is no schema registered or the given version does not exist, a SchemaInfo whose type is
```
shibd, line 414: I think we can add a new `Result` : `TopicSchemaNotFound`

When schema is not found, the Result on the callback is `TopicSchemaNotFound`.

What do you think?

include/pulsar/Message.h
BewareMyPower: Yes. The Java API design is terrible. It really harms the user experience. Actually the byte array always represents a big-endian encoded 8 bytes integer, unless it's empty, which indicates no schema version. I believe it's a mistake made by the author. You can also see `Schemas#getSchemaInfo` in the `pulsar-admin` module, you can see the parameter type is long.

```java
    SchemaInfo getSchemaInfo(String topic, long version) throws PulsarAdminException;

    CompletableFuture<SchemaInfo> getSchemaInfoAsync(String topic, long version);
```

However, there is no document that tell users how to get the human-readable representation of the byte array. Returning a byte array is definitely a mistake. The Python client also wraps this API and returns a string to users. And the users can nearly do nothing with it.

I will open a discussion later in the mail list. But for now, I think it's good to add such a method.

include/pulsar/Client.h
BewareMyPower: In Java client, the `TopicNotFound` error is ignored and an empty optional of `SchemaInfo` is returned. So I thought it should not be treated as an exceptional case and users should process this error. I just use a `NONE` schema to replace an empty optional.

But I just thought again and IMO an error should be returned if we cannot get the schema for the given topic and version. So I will adopt the suggestion you made so that users should not handle the schema type.

BewareMyPower: I won't add a new `Result`. The server side returns a `TopicNotFound` error. https://github.com/apache/pulsar/blob/f7c0b3c49c9ad8c28d0b00aa30d727850eb8bc04/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ClientCnx.java#L999

So I use `ResultTopicNotFound` here.

include/pulsar/Message.h
shibd: Get it, thanks.

shibd (approved):

include/pulsar/Message.h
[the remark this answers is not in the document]
BewareMyPower: @RobertIndie PTAL again.

RobertIndie (approved):

armanjupriya-er:
@BewareMyPower : Can this PR be back merged to Pulsar 2.10 CPP client?

BewareMyPower:
@armanjupriya-er The Pulsar CPP client version is no longer associated with the Pulsar version. So there is no need to "back merged". Just use the latest client

armanjupriya-er:
> @armanjupriya-er The Pulsar CPP client version is no longer associated with the Pulsar version. So there is no need to "back merged". Just use the latest client

Thank you @BewareMyPower of the quick reply.

Issue #268, opened 2023-05-15
[Bug]  C++ versions less than C++14 are not supported.


mattisonchao:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

3.2.0-candidate-3

### Minimal reproduce step

```
cmake .
make
```

### What did you expect to see?

run `make` success.

### What did you see instead?

```
[ 64%] Building CXX object tests/CMakeFiles/pulsar-tests.dir/AcknowledgeTest.cc.o
In file included from /Users/mattisonchao/Projects/apache/pulsar-client-cpp/tests/AcknowledgeTest.cc:19:
In file included from /usr/local/include/gtest/gtest.h:64:
In file included from /usr/local/include/gtest/gtest-assertion-result.h:46:
In file included from /usr/local/include/gtest/gtest-message.h:57:
/usr/local/include/gtest/internal/gtest-port.h:270:2: error: C++ versions less than C++14 are not supported.
#error C++ versions less than C++14 are not supported.
 ^
In file included from /Users/mattisonchao/Projects/apache/pulsar-client-cpp/tests/AcknowledgeTest.cc:19:
In file included from /usr/local/include/gtest/gtest.h:65:
In file included from /usr/local/include/gtest/gtest-death-test.h:43:
In file included from /usr/local/include/gtest/internal/gtest-death-test-internal.h:47:
In file included from /usr/local/include/gtest/gtest-matchers.h:48:
In file included from /usr/local/include/gtest/gtest-printers.h:115:
/usr/local/include/gtest/internal/gtest-internal.h:636:54: error: too few template arguments for class template 'less'
  typedef ::std::map<std::string, CodeLocation, std::less<>> RegisteredTestsMap;
                                                     ^
/Library/Developer/CommandLineTools/SDKs/MacOSX13.0.sdk/usr/include/c++/v1/__functional/operations.h:473:29: note: template is declared here
struct _LIBCPP_TEMPLATE_VIS less
                            ^
In file included from /Users/mattisonchao/Projects/apache/pulsar-client-cpp/tests/AcknowledgeTest.cc:19:
In file included from /usr/local/include/gtest/gtest.h:65:
In file included from /usr/local/include/gtest/gtest-death-test.h:43:
In file included from /usr/local/include/gtest/internal/gtest-death-test-internal.h:47:
In file included from /usr/local/include/gtest/gtest-matchers.h:48:
[...]

PR #269, opened 2023-05-15
Support specifying the C++ standard for some higher dependencies

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/268

### Motivation

When users build from source with a dependency that requires a higher C++ version than 11, it will fail because currently the C++ standard is pinned to 11. For example, the latest GTest requires C++14 now. So currently there is no way to build the tests except modifying the CMakeLists.txt file.

### Modifications

Make `CMAKE_CXX_STANDARD` configurable and test it for the macOS workflow because the `gtest` dependency installed from `brew` requires C++14. Add the description into the README for it.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,8 +88,9 @@ find_package(Threads REQUIRED)
 MESSAGE(STATUS "Threads library: " ${CMAKE_THREAD_LIBS_INIT})
 
 set(Boost_NO_BOOST_CMAKE ON)
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_C_STANDARD 11)
+if (NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+endif ()
 
 # Compiler specific configuration:
 # https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,7 @@ set(Boost_NO_BOOST_CMAKE ON)
 if (NOT CMAKE_CXX_STANDARD)
     set(CMAKE_CXX_STANDARD 11)
 endif ()
+set(CMAKE_C_STANDARD 11)
 
 # Compiler specific configuration:
 # https://stackoverflow.com/questions/10046114/in-cmake-how-can-i-test-if-the-compiler-is-clang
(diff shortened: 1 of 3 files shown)


mattisonchao (approved):

CMakeLists.txt
```diff
@@ -88,7 +88,9 @@ find_package(Threads REQUIRED)
 MESSAGE(STATUS "Threads library: " ${CMAKE_THREAD_LIBS_INIT})
 
 set(Boost_NO_BOOST_CMAKE ON)
-set(CMAKE_CXX_STANDARD 11)
+if (NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 11)
+endif ()
 set(CMAKE_C_STANDARD 11)
```
tisonkun, line 94: For options alignment, perhaps perform the same change for `CMAKE_C_STANDARD`?

BewareMyPower: No. The C standard is only applied to
- The C headers (`*.h` files) under https://github.com/apache/pulsar-client-cpp/tree/main/include/pulsar/c
- The C examples, i.e. the `*.c` files under https://github.com/apache/pulsar-client-cpp/tree/main/examples

They are never affected by any external dependency.

We should still stick to the C++11 and C11 standard for this project, but we cannot avoid the minimum standard version requirements from external C++ dependencies.

tisonkun: Thanks for your explanation! That sounds good to me.

tisonkun (approved):

tisonkun:
For reference - another way is pinning the version of gtest, like [gtest.cmake in Kvrocks](https://github.com/apache/incubator-kvrocks/blob/75db3e42df8a715339abaf02d640f3c62d43304a/cmake/gtest.cmake).

BewareMyPower:
> another way is pinning the version of gtest

Yes. But for most C++ developers, it's still much easier to install the binary dependencies via the default package manager on Linux and macOS

The `FetchContent` feature of CMake just downloads the source code. It still needs to build and install the dependencies. And as you can see we still need to pay some extra efforts to it: https://github.com/apache/incubator-kvrocks/blob/75db3e42df8a715339abaf02d640f3c62d43304a/cmake/utils.cmake

Actually, we can achieve this goal via Vcpkg as well. Vcpkg can specify dependency versions. See example here: https://github.com/apache/pulsar-client-python/pull/23/files#diff-e62a7794d36dac6a2bbd197522dfe4d80d8d77f2e54e4772271e14e0103829e1

But still, for development, installing binaries directly will be much easier unless the dependencies are depended by source.

PR #275, opened 2023-05-30
add cmake-build-debug dir to gitignore

Shoothzj:
### Motivation
add cmake-build-debug dir to gitignore. It shouldn't committed to the repo

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`


.gitignore
```diff
@@ -83,6 +83,7 @@ Makefile
 cmake_install.cmake
 CMakeFiles
 CMakeCache.txt
+/cmake-build-debug
```
BewareMyPower, line 86: Should we change it to `./cmake-build-debug`? And CMake can specifiy an arbitrary build directory. For example, the command `cmake -B build` will create a `build` directory to contain CMake build files.

Though it makes sense if `cmake-build-debug` is generated by a specific IDE by default. For example, Visual Studio creates `out/` directory as the build directory. If then, please add a comment to explain.

.gitignore
```diff
@@ -83,6 +83,8 @@ Makefile
 cmake_install.cmake
 CMakeFiles
 CMakeCache.txt
+# cmake files generated by CLion
+/cmake-build-debug
```
RobertIndie, line 87: It's better to use `/cmake-build-*` just like what JetBrains suggested: 
https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Global/JetBrains.gitignore#L43-L44

Shoothzj:
@BewareMyPower @RobertIndie I think we can merge this PR

BewareMyPower (approved):

PR #276, opened 2023-05-30
[feat][consumer] Support parse broker metadata

Shoothzj:
### Motivation
pulsar-client-cpp doesn't support parse broker metadata. Which will error whiling connect to  enabledbrokerMetada pulsar.
This PR makes pulsar-client-cpp can consume messages well, but haven't exposing index interface yet.

See also

https://github.com/apache/pulsar-client-go/pull/745
https://github.com/apache/pulsar/blob/e38091044c428af002b16110531497e2abc897d2/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ConsumerImpl.java#L1289

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

PR #277, opened 2023-06-01
[feat][Message] Add getIndex method on Message

Shoothzj:
### Motivation
After #276 , We already can consume messages contains index well. Now we can add a method allow user to getIndex.
Index is an optional brokerMetadata. If the index not exists, we will return -1

### Verifying this change
- Add the assertions to assert the index not equals -1.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/Message.h
+++ b/include/pulsar/Message.h
@@ -31,6 +31,7 @@
 namespace pulsar {
 namespace proto {
 class CommandMessage;
+class BrokerEntryMetadata;
 class MessageMetadata;
 class SingleMessageMetadata;
 }  // namespace proto
@@ -125,6 +126,12 @@ class PULSAR_PUBLIC Message {
     void setMessageId(const MessageId& messageId) const;
 
     /**
+     * Get the index of this message, if it doesn't exist, return -1
+     * @return
+     */
+    int64_t getIndex() const;
+
+    /**
      * Get the partition key for this message
      * @return key string that is hashed to determine message's topic partition
      */
@@ -195,9 +202,11 @@ class PULSAR_PUBLIC Message {
     MessageImplPtr impl_;
 
     Message(MessageImplPtr& impl);
-    Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload);
+    Message(const MessageId& messageId, proto::BrokerEntryMetadata& brokerEntryMetadata,
+            proto::MessageMetadata& metadata, SharedBuffer& payload);
     /// Used for Batch Messages
-    Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload,
+    Message(const MessageId& messageId, proto::BrokerEntryMetadata& brokerEntryMetadata,
+            proto::MessageMetadata& metadata, SharedBuffer& payload,
             proto::SingleMessageMetadata& singleMetadata, const std::shared_ptr<std::string>& topicName);
     friend class PartitionedProducerImpl;
     friend class MultiTopicsConsumerImpl;
--- a/lib/Message.cc
+++ b/lib/Message.cc
@@ -70,17 +70,21 @@ Message::Message() : impl_() {}
 
 Message::Message(MessageImplPtr& impl) : impl_(impl) {}
 
-Message::Message(const MessageId& messageId, proto::MessageMetadata& metadata, SharedBuffer& payload)
+Message::Message(const MessageId& messageId, proto::BrokerEntryMetadata& brokerEntryMetadata,
+                 proto::MessageMetadata& metadata, SharedBuffer& payload)
     : impl_(std::make_shared<MessageImpl>()) {
     impl_->messageId = messageId;
+    impl_->brokerEntryMetadata = brokerEntryMetadata;
     impl_->metadata = metadata;
     impl_->payload = payload;
 }
 
-Message::Message(const MessageId& messageID, proto::MessageMetadata& metadata, SharedBuffer& payload,
+Message::Message(const MessageId& messageID, proto::BrokerEntryMetadata& brokerEntryMetadata,
+                 proto::MessageMetadata& metadata, SharedBuffer& payload,
                  proto::SingleMessageMetadata& singleMetadata, const std::shared_ptr<std::string>& topicName)
     : impl_(std::make_shared<MessageImpl>()) {
     impl_->messageId = messageID;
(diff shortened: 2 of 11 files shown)


Shoothzj:
@BewareMyPower PTAL

lib/Message.cc
```diff
@@ -136,6 +140,15 @@ void Message::setMessageId(const MessageId& messageID) const {
     return;
 }
 
+const int64_t Message::getIndex() const {
+    if (!impl_) {
+        return -1;
+    } else {
+        // casting uint64_t to int64_t, server definition ensures that's safe
+        return static_cast<int64_t>(impl_->brokerEntryMetadata.index());
```
BewareMyPower, line 148: I think we should check `has_index()` as well because `index` is an optional field.

```c++
if (!impl_ || !impl_->has_index()) {
    return -1;
```

include/pulsar/Message.h
```diff
@@ -124,6 +125,12 @@ class PULSAR_PUBLIC Message {
      */
     void setMessageId(const MessageId& messageId) const;
 
+    /**
+     * Get the index of this message, if it doesn't exist, return -1
+     * @return
+     */
+    const int64_t getIndex() const;
```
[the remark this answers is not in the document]
Shoothzj: @BewareMyPower Fixed

lib/MessageImpl.cc
```diff
@@ -20,7 +20,8 @@
 
 namespace pulsar {
 
-MessageImpl::MessageImpl() : metadata(), payload(), messageId(), cnx_(0), topicName_(), redeliveryCount_() {}
+MessageImpl::MessageImpl()
+    : metadata(), brokerEntryMetadata(), payload(), messageId(), cnx_(0), topicName_(), redeliveryCount_() {}
```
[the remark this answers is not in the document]
Shoothzj: @BewareMyPower removed

lib/Message.cc
Shoothzj: @BewareMyPower Fixed

BewareMyPower (changes requested):
image

You can see the `BrokerEntryMetadata` was not added when messages were from the C++ producer.

From the server side, you can see:

```java
if (cnx.getRemoteEndpointProtocolVersion() < ProtocolVersion.v18.getValue()
        || !cnx.supportBrokerMetadata()
        || !cnx.getBrokerService().getPulsar().getConfig()
        .isExposingBrokerEntryMetadataToClientEnabled()) {
    Commands.skipBrokerEntryMetadataIfExist(metadataAndPayload);
}
```

The client needs to set the feature flags in `newConnect`, just like the Java client does:

```java
    private static void setFeatureFlags(FeatureFlags flags) {
        flags.setSupportsAuthRefresh(true);
        flags.setSupportsBrokerEntryMetadata(true);
        flags.setSupportsPartialProducer(true);
    }
```

tests/brokermetadata/BrokerMetadataTest.cc
```diff
@@ -25,19 +25,58 @@ using namespace pulsar;
 TEST(BrokerMetadataTest, testConsumeSuccess) {
     Client client{"pulsar://localhost:6650"};
     Producer producer;
-    Result producerResult = client.createProducer("persistent://public/default/testConsumeSuccess", producer);
+    ProducerConfiguration producerConfiguration;
+    producerConfiguration.setBatchingEnabled(false);
```
BewareMyPower, line 29: The only difference between these two tests is here. I think you can use `INSTANTIATE_TEST_SUITE_P` to reduce duplicated code. See `AcknowledgeTest.cc` for example.

But it's only an improvement. If you think it's too complicated, just resolve this comment.

tests/brokermetadata/BrokerMetadataTest.cc
```diff
@@ -25,19 +25,58 @@ using namespace pulsar;
...
+    Producer producer;
+    Result producerResult = client.createProducer("persistent://public/default/topic-batch", producer);
+    ASSERT_EQ(producerResult, ResultOk);
+    Consumer consumer;
+    Result consumerResult = client.subscribe("persistent://public/default/topic-batch", "sub", consumer);
+    ASSERT_EQ(consumerResult, ResultOk);
+    for (int i = 0; i < 10; i++) {
+        std::string content = "testConsumeSuccess" + std::to_string(i);
+        const auto msg = MessageBuilder().setContent(content).build();
+        Result sendResult = producer.send(msg);
```
BewareMyPower, line 68: If you want to verify the batch send case,  you should use `sendAsync` instead of `send` because batching is meaningless for synchronous send.

tests/brokermetadata/BrokerMetadataTest.cc
Shoothzj: @BewareMyPower I am fine with this improvement. Since we need to use async, I think we don't need this improvement

tests/brokermetadata/BrokerMetadataTest.cc
Shoothzj: @BewareMyPower Fixed

tests/brokermetadata/BrokerMetadataTest.cc
```diff
@@ -20,24 +20,74 @@
...
-    Result receiveResult = consumer.receive(receivedMsg);
-    ASSERT_EQ(receiveResult, ResultOk);
-    ASSERT_EQ(receivedMsg.getDataAsString(), "testConsumeSuccess");
+    for (int i = 0; i < 10; i++) {
+        Result receiveResult =
+            consumer.receive(receivedMsg, 1000);  // Assumed that we wait 1000 ms for each message
+        ASSERT_EQ(receiveResult, ResultOk);
+        ASSERT_EQ(receivedMsg.getDataAsString(), "testConsumeSuccess" + std::to_string(i));
+        ASSERT_GE(receivedMsg.getIndex(), 0);
+        ASSERT_LT(receivedMsg.getIndex(), 10);
```
BewareMyPower, line 89: Why did you check `0 < index < 10`? The `index` should be equal to `i`.

Shoothzj: @BewareMyPower Sorry, I missed push a commit. I will fix it after going to home.

Shoothzj:
ping @BewareMyPower

BewareMyPower (approved):

PR #283, opened 2023-06-13
Remove log4cxx dependency and its specified configuration

BewareMyPower:
### Motivation

See https://github.com/apache/pulsar/pull/1827, which also marked the log4cxx configurations as deprecated.

### Modifications

Remove all log4cxx related stuff. log4cxx users can implement their own `LoggerFactory` like the built-in implementation before this PR.

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [x] `doc` 

- [ ] `doc-complete`


RobertIndie:
This is a breaking change, and it fails the Python client build. I think we need to discuss it on the mailing list.

BewareMyPower:
Yes, we should also remove the configs from Python as well.

Issue #261, opened 2023-04-28
[Bug] Unexpected error logs when closing client on Windows


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

- Pulsar 2.11.0
- Pulsar Client 3.2.0 candidate 1: https://dist.apache.org/repos/dist/dev/pulsar/pulsar-client-cpp/pulsar-client-cpp-3.2.0-candidate-1

### Minimal reproduce step

```c++
#include <pulsar/Client.h>
using namespace pulsar;

int main(int argc, char *argv[]) {
  // The server ran inside WSL2 on Windows
  Client client("pulsar://172.22.48.50:6650");

  Producer producer;
  auto result =
      client.createProducer("persistent://public/default/test-topic", producer);
  if (result != ResultOk) {
    std::cerr << "Failed to create producer: " << result << std::endl;
    return 1;
  }

  client.close();
  return 0;
}
```

The Pulsar standalone ran on the Ubuntu 20.04 WSL2 subsystem whose IP is `172.22.48.50`.

### What did you expect to see?

No error logs appeared.

### What did you see instead?

There was an error log when closing the client:

```
2023-04-28 22:08:06.875 INFO  [12684] D:\a\pulsar-client-cpp\pulsar-client-cpp\lib\ProducerImpl:755 | [persistent://public/default/test-topic, standalone-5-42] Closing producer for topic 000001ED65D25180
2023-04-28 22:08:06.877 INFO  [11036] D:\a\pulsar-client-cpp\pulsar-client-cpp\lib\ProducerImpl:719 | [persistent://public/default/test-topic, standalone-5-42] Closed producer 0
2023-04-28 22:08:06.878 INFO  [3112] D:\a\pulsar-client-cpp\pulsar-client-cpp\lib\ClientConnection:1205 | [172.22.48.1:5376 -> 172.22.48.50:6650] Connection disconnected
2023-04-28 22:08:06.878 INFO  [3112] D:\a\pulsar-client-cpp\pulsar-client-cpp\lib\ClientConnection:1205 | [172.22.48.1:5377 -> 172.22.48.50:6650] Connection disconnected
[...]

PR #284, opened 2023-06-14
Close the socket gracefully on Windows

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/261

### Motivation

When closing the socket on Windows, the `ERROR_CONNECTION_ABORTED` error might be returned in the callback of `async_receive`. It's caused by the `socket::close` method is not portable enough, see https://www.boost.org/doc/libs/1_81_0/doc/html/boost_asio/reference/basic_stream_socket/close/overload2.html

> For portable behaviour with respect to graceful closure of a connected socket, call shutdown() before closing the socket.

### Modifications

Call `shutdown` in `ClientConnection::closeSocket` before calling `close`. In addition, when no bytes are returned or the `eof` error happened,  print the logs with debug level whatever the error code is in the read callback.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -609,13 +609,13 @@ void ClientConnection::handleRead(const boost::system::error_code& err, size_t b
     incomingBuffer_.bytesWritten(bytesTransferred);
 
     if (err || bytesTransferred == 0) {
-        if (err) {
-            if (err == boost::asio::error::operation_aborted) {
-                LOG_DEBUG(cnxString_ << "Read operation was canceled: " << err.message());
-            } else {
-                LOG_ERROR(cnxString_ << "Read operation failed: " << err.message());
-            }
-        }  // else: bytesTransferred == 0, which means server has closed the connection
+        if (bytesTransferred == 0 || err == boost::asio::error::eof) {
+            LOG_DEBUG(cnxString_ << "Server closed the connection: " << err.message());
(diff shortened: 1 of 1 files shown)


BewareMyPower:
Here are some tests results with debug logs based on this patch.

Without calling `shutdown`, the error log is always:

```
# error code: connection_aborted ("The network connection was aborted by the local system")
2023-06-14 13:14:15.190 DEBUG [11468] ..\..\..\lib\ClientConnection:615 | [[::1]:10740 -> [::1]:6650] Server closed the connection:  由本地系统中止网络连接。
```

With calling `shutdown`, the error log will be:

```
# error code: operation_aborted ("Operation cancelled")
2023-06-14 13:15:08.607 DEBUG [21080] ..\..\..\lib\ClientConnection:613 | [[::1]:10745 -> [::1]:6650] Read operation was canceled: 由于线程退出或应用程序请求，已中止 I/O 操作。
```

There is a little chance that the error code could be `eof` (`value()` is 2 on Windows) when I debugged it in my IDE so I handled the EOF error the same way with `bytesTransferred == 0` because they both mean the socket client calls `recv` while the socket server has already closed.

image

RobertIndie (approved):

PR #290, opened 2023-06-21
Fix build failure with the Protobuf 23.3

BewareMyPower:
### Motivation

The current CMakeLists.txt does not work with the latest Protobuf (23.3.0). It's because currently the Module mode is used to find Protobuf, while the `FindProtobuf.cmake` is not updated to find the Abseil dependency.

See https://github.com/protocolbuffers/protobuf/issues/12292#issuecomment-1529680040

### Modifications

For macOS, use the Config mode to find Protobuf. It's because in other systems, the Module mode works well. Besides, enable `PROTOBUF_USE_DLLS` as a workaround for
https://github.com/protocolbuffers/protobuf/pull/12983 is not released.

Pin the default C++ standard to 17 for macOS so that users don't need to set the C++ standard manually.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/.github/workflows/ci-pr-validation.yaml
+++ b/.github/workflows/ci-pr-validation.yaml
@@ -50,11 +50,7 @@ jobs:
       - name: Install deps (macOS)
         if: ${{ startsWith(matrix.os, 'macos') }}
         run: |
-          # Install protobuf v21.12
-          wget https://raw.githubusercontent.com/Homebrew/homebrew-core/2d47ed2eac09d59ffc2c92b1d804f1c232188c88/Formula/protobuf.rb
-          brew install --formula ./protobuf.rb
-          brew install pkg-config wireshark
-
+          brew install pkg-config wireshark protobuf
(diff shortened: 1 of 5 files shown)


shibd:
In this way, the macOS developer needs to use C++ 17 to compile this project.

We need to update README.

BewareMyPower:
> In this way, the macOS developer needs to use C++ 17 to compile this project.

Yes. BTW, I think we can make C++17 as the default standard for macOS users because it's now required by the library dependency, not the test dependency. I marked this PR as drafted first.

BewareMyPower:
@shibd I've updated the README and pinned the C++ standard to 17 for macOS so that we don't need to set `CMAKE_CXX_STANDARD` explicitly with the default dependencies installed on macOS anymore.

shibd (approved):

Issue #292, opened 2023-06-27
[Bug] Race condition when closing during seek


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

- OS: macOS 13.4
- Pulsar: 3.0.0
- Client: master (633f4bbe8c182128da09803172676b9d6af05057)
- Clang++: 14.0.3

### Minimal reproduce step

See https://github.com/BewareMyPower/pulsar-client-cpp-demo/blob/main/seek_close_loop.cc

```c++
  for (int i = 0; i < 100; i++) {
    Consumer consumer;
    if (auto result = client.subscribe(topic, "sub", consumer);
        result != ResultOk) {
      LOG_ERROR(i << " Failed to subscribe: " << result);
      return 2;
    }
    consumer.seekAsync(MessageId::earliest(), [](auto result) {
      LOG_INFO("Seek to earliest: " << result);
    });
    if (auto result = consumer.close(); result != ResultOk) {
      LOG_ERROR("Failed to close: " << result);
    }
  }
```

i.e. run a seek-close loop. Use `seekAsync` so that when `close` is called, the consumer is during seek.

### What did you expect to see?

The application exits with 0.

### What did you see instead?

The application exited with 2 that the 2nd time to subscribe failed with `ResultRetryable`.

```
2023-06-27 15:33:18.865 ERROR [0x1e2c85e00] log_utils:30 | 1 Failed to subscribe: Retryable
2023-06-27 15:33:18.865 INFO  [0x16d26f000] log_utils:39 | Seek to earliest: Ok
```

### Anything else?

https://github.com/apache/pulsar/pull/7819 tried to fix this issue but it does not fix it.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #293, opened 2023-06-27
Fix retriable errors not handled well when creating producer or consumer

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/292

### Motivation

When a consumer failed to subscribe due to a retriable error, the time point comparation is wrong:

https://github.com/apache/pulsar-client-cpp/blob/633f4bbe8c182128da09803172676b9d6af05057/lib/ConsumerImpl.cc#L321

`creationTimestamp_ + operationTimeut_` is the deadline, `TimeUtils::now()` is the current time, we should use `>` instead of `<` here to compare them. Otherwise, if the consumer encountered a retriable error and the deadline is not exceeded, the consumer won't reconnect and fail with `ResultRetryable`.

### Modifications

Reverse the comparation between the deadline and the current time. When it times out, completing the future with `ResultTimeout` instead of the `result` itself, which is always `ResultRetryable`.

Add `ConsumerTest.testRetrySubscribe` to verify this change.

### TODO

Support configuring the operation timeout in milliseconds.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -318,10 +318,13 @@ void ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result r
             scheduleReconnection(get_shared_this_ptr());
         } else {
             // Consumer was not yet created, retry to connect to broker if it's possible
-            if (isRetriableError(result) && (creationTimestamp_ + operationTimeut_ < TimeUtils::now())) {
+            if (isRetriableError(result) && (TimeUtils::now() - creationTimestamp_ < operationTimeut_)) {
                 LOG_WARN(getName() << "Temporary error in creating consumer : " << strResult(result));
                 scheduleReconnection(get_shared_this_ptr());
             } else {
+                if (isRetriableError(result)) {
+                    result = ResultTimeout;
+                }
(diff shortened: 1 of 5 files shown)


lib/ConsumerImpl.cc
```diff
@@ -318,10 +318,13 @@ void ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result r
             scheduleReconnection(get_shared_this_ptr());
         } else {
             // Consumer was not yet created, retry to connect to broker if it's possible
-            if (isRetriableError(result) && (creationTimestamp_ + operationTimeut_ < TimeUtils::now())) {
```
shibd, line 321: Producer has same issue: https://github.com/apache/pulsar-client-cpp/blob/e79357e5b40eef6397d256f31b049ce418f15066/lib/ProducerImpl.cc#L284

BewareMyPower: Fixed, PTAL again.

shibd (approved):

PR #294, opened 2023-06-27
feat: Support table view for C client.

shibd:
### Motivation

Support table view for C client.

### Modifications

- Add table_view for C client.
- Add table_view_configuration for C client.

### Verifying this change

- Add CTableViewTest cover it.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/include/pulsar/c/table_view.h
+++ b/include/pulsar/c/table_view.h
@@ -1,0 +1,147 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#pragma once
+
+#include <pulsar/defines.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <pulsar/c/message.h>
+#include <pulsar/c/messages.h>
+#include <pulsar/c/result.h>
+#include <stdint.h>
+
+typedef struct _pulsar_table_view pulsar_table_view_t;
+
+typedef void (*pulsar_table_view_action)(const char *key, const char *value, void *ctx);
+typedef void (*pulsar_result_callback)(pulsar_result, void *);
+
+/**
+ * Move the latest value associated with the key.
+ *
+ * NOTE:
+ * 1. Once the value has been retrieved successfully,
+ * the associated value will be removed from the table view until next time the value is updated.
+ * 2. Once the value has been retrieved successfully, `*value` will point to the memory that is allocated
+ * internally. You have to call `delete value` to free it.
+ *
+ * Example:
+ *
+ * ```c
+ * pulsar_table_view_t *table_view;
+ * char *value;
+ * while (true) {
+ *     if (pulsar_table_view_retrieve_value(table_view, "key", &value)) {
+ *         printf("value is update to: %s", value);
+ *     } else {
+ *         // sleep for a while or print the message that value is not updated
+ *     }
+ * }
+ * delete value;
+ * ```
+ *
+ * @param table_view
+ * @param key
+ * @param value the value associated with the key
+ * @return true if there is an associated value of the key, otherwise false
+ */
+PULSAR_PUBLIC bool pulsar_table_view_retrieve_value(pulsar_table_view_t *table_view, const char *key,
+                                                    char **value);
+
+/**
+ * It's similar with `pulsar_table_view_retrieve_value` except the associated value not will be removed from
+ * the table view.
+ *
+ * NOTE:
+ * Once the value has been get successfully, `*value` will point to the memory that is allocated internally.
+ * You have to call `delete value` to free it.
+ *
+ * @param table_view
+ * @param key
+ * @param value the value associated with the key
+ * @return true if there is an associated value of the key, otherwise false
+ */
+PULSAR_PUBLIC bool pulsar_table_view_get_value(pulsar_table_view_t *table_view, const char *key,
+                                               char **value);
+
+/**
+ * Check if the key exists in the table view.
+ * @param table_view
+ * @param key
+ * @return true if the key exists in the table view
+ */
+PULSAR_PUBLIC bool pulsar_table_view_contain_key(pulsar_table_view_t *table_view, const char *key);
+
+/**
+ * Get the size of the elements.
+ * @param table_view
+ * @return
+ */
+PULSAR_PUBLIC int pulsar_table_view_size(pulsar_table_view_t *table_view);
+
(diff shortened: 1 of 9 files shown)


include/pulsar/c/table_view.h
```diff
@@ -0,0 +1,147 @@
...
+ * pulsar_table_view_t *table_view;
+ * char *value;
+ * while (true) {
+ *     if (pulsar_table_view_retrieve_value(table_view, "key", &value)) {
+ *         printf("value is update to: %s", value);
+ *     } else {
+ *         // sleep for a while or print the message that value is not updated
+ *     }
+ * }
+ * delete value;
```
BewareMyPower, line 58: It's a C header, `delete` is a C++ keyword. You should not use `delete` here. I see the memory that `value` points to is allocated by `malloc` in `str_malloc_and_copy`. You should use `free` instead of `delete`.

BewareMyPower (changes requested):

tests/c/c_TableViewTest.cc
```diff
@@ -0,0 +1,92 @@
...
+    pulsar_table_view_configuration_set_subscription_name(table_view_conf, sub_name);
+    pulsar_table_view_t *table_view;
+    result = pulsar_client_create_table_view(client, topic_name, table_view_conf, &table_view);
+    ASSERT_EQ(pulsar_result_Ok, result);
+
+    char *v1;
+    ASSERT_EQ(pulsar_table_view_size(table_view), 2);
+    ASSERT_TRUE(pulsar_table_view_get_value(table_view, "key1", &v1));
+    ASSERT_STREQ(v1, "content");
+    delete v1;
```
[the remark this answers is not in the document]
BewareMyPower: Even in C++, the behavior of malloc with delete is undefined.

https://stackoverflow.com/questions/10854210/behaviour-of-malloc-with-delete-in-c

BewareMyPower:
In addition, I don't think use a null-terminated string as the tableview value is good for C API. The value is actually a byte array rather than a null-terminated string. Unlike the `TableView` class, it's okay to return a `std::string`, which is actually a byte array in C++. For example, if the producer sends 4 bytes: `[0x01, 0x00, 0x02, 0x00]`, you can never know the length.

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

static void get_value(char **value) {
  char *data = (char *)malloc(4);
  data[0] = 0x01;
  data[1] = 0x00;
  data[2] = 0x02;
  data[3] = 0x00;
  *value = data;
}

int main(int argc, char *argv[]) {
  char *value;
  get_value(&value);
  printf("length: %zu\n", strlen(value));
  return 0;
}
```

The code example is a demo to show the difference between a byte array and a null-terminated string.

So it's better to add a length parameter like:

```c
#include <stdio.h>
#include <stdlib.h>

static void get_value(char **value_ptr, size_t *value_len) {
  char *data = (char *)malloc(4);
  data[0] = 0x01;
  data[1] = 0x00;
  data[2] = 0x02;
  data[3] = 0x00;
  *value_ptr = data;
  *value_len = 4;
}

int main(int argc, char *argv[]) {
  char *value;
  size_t len;
  get_value(&value, &len);
  printf("length: %zu\n", len);
  for (size_t i = 0; i < len; i++) {
    printf("0x%02x%c", value[i], (i < len - 1) ? ' ' : '\n');
  }
  return 0;
}
```

Output:

```
length: 4
0x01 0x00 0x02 0x00
```

shibd:
> In addition, I don't think use a null-terminated string as the tableview value is good for C API. The value is actually a byte array rather than a null-terminated string. Unlike the TableView class, it's okay to return a std::string, which is actually a byte array in C++. For example, if the producer sends 4 bytes: [0x01, 0x00, 0x02, 0x00], you can never know the length.

Thanks explain, that makes sense to me.

New [commit](https://github.com/shibd/pulsar-client-cpp/commit/c2471a27137dffae46e51903b8a6c2ddda6559f6) modifications:
1. Add `*value_size` param on `pulsar_table_view_get_value`, `pulsar_table_view_retrieve_value` method and `pulsar_table_view_action`.
2. Change `*value` type from `char **value` to `void **value`.
3. Remove `pulsar_string_map_t *pulsar_table_view_snapshot(pulsar_table_view_t *table_view)` method. Because on this case.
 -  `pulsar_string_map_get` also does not return value_size, 
 -  this **string_map** may not be suitable for `table_view` for C.
 -  I think the `table_view` currently provided interface can meet the user use case.
 -  If needed, I'll add it in another PR.


PTAL.

tests/c/c_TableViewTest.cc
```diff
@@ -0,0 +1,97 @@
...
+    // Send messages
+    const int num = 10;
+    const char *key1 = "key1";
+    const char *key2 = "key2";
+    size_t data_size = 4;
+    int *data = (int *)malloc(data_size);
+    data[0] = 0x01;
+    data[1] = 0x00;
+    data[2] = 0x02;
+    data[3] = 0x00;
```
BewareMyPower, line 48: The pointer you used is wrong. You allocated 4 (`data_size`) bytes. However, you casted the `void*` to `int*`, so the access to `data[1]` to `data[3]` is invalid.

Assuming `p` is the address returned by `malloc` and the machine is little-endian, the value stored are actually :

| p | p + 1 | p + 2 | p + 3 |
| :- | :- | :- | :- |
| 0x01 | 0x00 | 0x00 | 0x00 |

`data[1]` is `p+4`.

Take the following example:

```c
#include <stdio.h>
#include <stdlib.h>

int main(int argc, char *argv[]) {
  int *data = (int *)malloc(4);
  data[0] = 0x01;
  data[1] = 0x02;
  printf("%p\n%p\n", &data[0], &data[1]);
  char *p = (char *)data;
  for (int i = 0; i < 4; i++) {
    printf("0x%02x ", p[i]);
  }
  printf("\n");
  free(data);
  return 0;
}
```

```bash
$ gcc 1.c
$ ./a.out
0x600001838030
0x600001838034
0x01 0x00 0x00 0x00 
```

BewareMyPower (changes requested):

include/pulsar/c/table_view.h
```diff
@@ -0,0 +1,139 @@
...
+PULSAR_PUBLIC pulsar_result pulsar_table_view_close(pulsar_table_view_t *table_view);
+
+/**
+ * Async close the table view and stop the broker to push more messages
+ * @param table_view
+ * @param callback
+ * @param ctx
+ */
+PULSAR_PUBLIC void pulsar_table_view_close_async(pulsar_table_view_t *table_view,
+                                                 pulsar_result_callback callback, void *ctx);
```
BewareMyPower, line 135: Could you also add a test for this? If you added a new API, please cover it in tests.

include/pulsar/c/table_view.h
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Example:
+ *
+ * ```c
+ * pulsar_table_view_t *table_view;
+ * char *value;
+ * size_t value_size;
+ * while (true) {
+ *     if (pulsar_table_view_retrieve_value(table_view, "key", &value, &value_size)) {
+ *         printf("value is update to: %s", value);
```
BewareMyPower, line 54: You still assumes the `value` is a null-terminated string here. You should treat it as a byte array.

```c
void* value;
size_t value_size;
while (true) {
    if (pulsar_table_view_retrieve_value(table_view, "key", &value, &value_size)) {
        for (size_t i = 0; i < value_size; i++) {
            printf("0x%02x%c", ((char*) value)[i], (i + 1 == value_size) ? '\n': ' ');
```

BewareMyPower (commented):
LGTM except the example is still wrong.

BewareMyPower (approved):

PR #296, opened 2023-06-29
Fix the wrong backoff computation when retrying

BewareMyPower:
### Motivation

All the retryable operations share the same `Backoff` object in `RetryableLookupService`, so if the reconnection happens for some times, the delay of retrying will keeps the maximum value (30 seconds).

### Modifications

Refactor the design of the `RetryableLookupService`:
- Add a `RetryableOperation` class to represent a retryable operation, each instance has its own `Backoff` object. The operation could only be executed once.
- Add a `RetryableOperationCache` class to represent a map that maps a specific name to its associated operation. It's an optimization that if an operation (e.g. find the owner topic of topic A) was not complete while the same operation was executed, the future would be reused.
- In `RetryableLookupService`, just maintain some caches for different operations.
- Add `RetryableOperationCacheTest` to verify the behaviors.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/RetryableLookupService.h
+++ b/lib/RetryableLookupService.h
@@ -18,21 +18,15 @@
  */
 #pragma once
 
-#include <algorithm>
-#include <memory>
-
-#include "Backoff.h"
-#include "ExecutorService.h"
-#include "LogUtils.h"
 #include "LookupDataResult.h"
 #include "LookupService.h"
-#include "SynchronizedHashMap.h"
+#include "NamespaceName.h"
+#include "RetryableOperationCache.h"
 #include "TopicName.h"
 
 namespace pulsar {
 
-class RetryableLookupService : public LookupService,
-                               public std::enable_shared_from_this<RetryableLookupService> {
+class RetryableLookupService : public LookupService {
    private:
     friend class PulsarFriend;
     struct PassKey {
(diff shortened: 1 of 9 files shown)


BewareMyPower:
`BasicEndToEndTest.testLookupThrottling` is stuck, I will fix this test ASAP.

lib/RetryableLookupService.h
```diff
@@ -50,117 +44,50 @@ class RetryableLookupService : public LookupService,
...
     }
 
-    template <typename T>
-    Future<Result, T> executeAsync(const std::string& key, std::function<Future<Result, T>()> f) {
-        Promise<Result, T> promise;
-        executeAsyncImpl(key, f, promise, timeout_);
-        return promise.getFuture();
+    size_t getNumberOfPendingTasks() const {
+        return lookupCache_->size() + partitionLookupCache_->size() + namespaceLookupCache_->size() +
+               getSchemaCache_->size();
```
RobertIndie, line 72: It's only used for testing. Do we need to expose it here?

BewareMyPower: I removed them. PTAL again.

lib/RetryableOperationCache.h
```diff
@@ -0,0 +1,117 @@
...
+                timer = executorProvider_->get()->createDeadlineTimer();
+            } catch (const std::runtime_error& e) {
+                LOG_ERROR("Failed to retry lookup for " << key << ": " << e.what());
+                Promise<Result, T> promise;
+                promise.setFailed(ResultConnectError);
+                return promise.getFuture();
+            }
+
+            auto operation = RetryableOperation<T>::create(key, std::move(func), timeoutSeconds_, timer);
+            auto future = operation->run();
```
shibd, line 73: We can call run on unlock after.

tests/TableViewTest.cc
```diff
@@ -23,7 +23,9 @@
 #include <future>
 
 #include "HttpHelper.h"
+#include "LogUtils.h"
 #include "PulsarFriend.h"
+#include "TopicName.h"
```
shibd, line 28: Is this redundant?

shibd (approved):
Nice catch! Left some small comments.

BewareMyPower:
It seems some tests failed after merging the main branch. Mark it as drafted currently.

```
    1278 ms: ./pulsar-tests ClientTest.testConnectTimeout (try #1)
    1259 ms: ./pulsar-tests ClientTest.testConnectTimeout (try #2)
     237 ms: ./pulsar-tests ClientTest.testCloseClient (try #1)
    1280 ms: ./pulsar-tests ClientTest.testConnectTimeout (try #3)
     241 ms: ./pulsar-tests ClientTest.testCloseClient (try #2)
     242 ms: ./pulsar-tests ClientTest.testCloseClient (try #3)
     311 ms: ./pulsar-tests ClientTest.testCloseClient (try #4)
    1410 ms: ./pulsar-tests ClientTest.testConnectTimeout (try #4)
    1056 ms: ./pulsar-tests Pulsar/MessageChunkingTest.testSeekChunkMessages/3 (try #1)
```

tests/TableViewTest.cc
BewareMyPower: No. This file uses `TopicDomain` in `TopicName.h`, which was included from another header before this patch.

lib/RetryableOperationCache.h
BewareMyPower: If we ran the operation after inserting it without the lock, there will be a race that:

| Thread | Method | Operation |
| :- | :- | :- |
| A | `run` | Insert the `operation`  |
| B | `clear` | Remove the `operation` from the cache |
| B | `clear` | Call `operation->cancel()` |
| A | `run` | Call `operation->run()` |

We should not call `run` after `cancel` on the same operation. So it would be more safe to call `run` before inserting it.

shibd: Thanks explain, agree.

BewareMyPower:
@shibd @RobertIndie Now all tests passed, PTAL again.

shibd (approved):

RobertIndie (approved):

Issue #298, opened 2023-07-03
[Bug] Synchronous send might stuck or crash when sending many messages


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

- OS: macOS 13.4
- Pulsar: 3.0.0
- Client: 3.2.0 or the latest commit (804f87b)

### Minimal reproduce step

```c++
#include <pulsar/Client.h>
using namespace pulsar;

int main(int argc, char* argv[]) {
  Client client("pulsar://localhost:6650");

  Producer producer;
  Result result =
      client.createProducer("persistent://public/default/my-topic", producer);
  if (result != ResultOk) {
    return -1;
  }

  // Send synchronously
  for (int i = 0; i < 10000; i++) {
    Message msg = MessageBuilder().setContent("content").build();
    Result res = producer.send(msg);
    if (res != ResultOk) {
      std::cerr << "Failed to send " << i << res << std::endl;
      break;
    }
    // NOTE: It might stuck, if we logged for each message, it would work well.
    if (i % 100 == 99) {
      std::cout << std::time(nullptr) << " " << i << std::endl;
    }
  }

  client.close();
  return 0;
}
```

Use C++ 17. You can use this project to reproduce.

```bash
git clone https://github.com/bewaremypower/pulsar-client-cpp-demo.git
cd pulsar-client-cpp-demo
cmake -B build
cmake --build build
./build/send_example.out
```

### What did you expect to see?

All messages were sent successfully.

### What did you see instead?

It stuck after sending 900 messages.

```
1688385630 99
1688385630 199
1688385630 299
1688385630 399
1688385631 499
1688385631 599
1688385631 699
1688385631 799
1688385631 899
```

Then I used `lldb` to attach this process and print `c` to continue running.

```
1688385892 1099
1688385892 1199
...
1688385901 5999
1688385901 6099
libc++abi: terminating due to uncaught exception of type std::__1::system_error: condition_variable wait failed: Invalid argument
```

It continues sending 5000 messages and crashed.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

BewareMyPower:
Asynchronous send works well.

```c++
#include <pulsar/Client.h>

#include <atomic>
#include <chrono>
#include <thread>
using namespace pulsar;

int main(int argc, char* argv[]) {
[...]

PR #299, opened 2023-07-03
Fix the buggy Future and Promise implementations

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/298

### Motivation

Currently the `Future` and `Promise` are implemented manually by managing conditional variables. However, the conditional variable sometimes behaviors incorrectly on macOS, while the existing `future` and `promise` from the C++ standard library works well.

### Modifications

Redesign `Future` and `Promise` based on the utilities in the standard `<future>` header. In addition, fix the possible race condition when `addListener` is called after `setValue` or `setFailed`:
- Thread 1: call `setValue`, switch existing listeners and call them one by one out of the lock.
- Thread 2: call `addListener`, detect `complete_` is true and call the listener directly.

Now, the previous listeners and the new listener are called concurrently in thread 1 and 2.

### Verifications

Run the reproduce code in #298 for 10 times and found it never failed or hang.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/Future.h
+++ b/lib/Future.h
@@ -19,86 +19,99 @@
 #ifndef LIB_FUTURE_H_
 #define LIB_FUTURE_H_
 
-#include <condition_variable>
+#include <atomic>
 #include <functional>
+#include <future>
 #include <list>
 #include <memory>
 #include <mutex>
-
-using Lock = std::unique_lock<std::mutex>;
+#include <utility>
 
 namespace pulsar {
 
 template <typename Result, typename Type>
-struct InternalState {
-    std::mutex mutex;
-    std::condition_variable condition;
-    Result result;
-    Type value;
-    bool complete;
+class InternalState {
+   public:
+    using Listener = std::function<void(Result, const Type &)>;
+    using Pair = std::pair<Result, Type>;
 
-    std::list<typename std::function<void(Result, const Type&)> > listeners;
+    void addListener(Listener listener) {
+        if (completed()) {
+            // Allow get_future() being called multiple times, only the 1st time will wait() be called to wait
+            // until all previous listeners are done.
+            try {
+                listenersPromise_.get_future().wait();
+            } catch (const std::future_error &e) {
+                if (e.code() != std::future_errc::future_already_retrieved) {
+                    throw e;
+                }
+            }
+            listener(future_.get().first, future_.get().second);
+        } else {
+            std::lock_guard<std::mutex> lock{mutex_};
+            listeners_.emplace_back(listener);
+        }
+    }
+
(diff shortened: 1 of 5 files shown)


Demogorgon314 (approved):
Great work! After this change, the reproduce code never failed or hang.

shibd (approved):

lib/Future.h
```diff
@@ -19,162 +19,124 @@
...
+            // Allow get_future() being called multiple times, only the 1st time will wait() be called to wait
+            // until all previous listeners are done.
+            try {
+                listenersPromise_.get_future().wait();
+            } catch (const std::future_error &e) {
+                if (e.code() != std::future_errc::future_already_retrieved) {
+                    throw e;
+                }
+            }
+            listener(future_.get().first, future_.get().second);
```
RobertIndie, line 52: Is it possible that two thread call `addListener` after the future is completed and then result in these two listeners being executed concurrently?

lib/Future.h
```diff
@@ -19,162 +19,124 @@
...
         } else {
-            state->listeners.push_back(callback);
+            std::lock_guard<std::mutex> lock{mutex_};
+            listeners_.emplace_back(listener);
```
RobertIndie, line 55: It seems to be a race condition here.
* If one thread named A is going into the `addListener` and finds that it's not completed. 
* But then another thread named B completes the future and locks this mutex. 
* Thread A is blocking until all previous listeners are executed.
* Thread A get the lock. And push the listener to the list. But this listener wouldn't be executed anymore.

This will result in the loss of the listener in some cases.

lib/Future.h
BewareMyPower: Yes. We need another synchronization mechanism for this case.

lib/Future.h
BewareMyPower: Good suggestion. I will find a way to fix it.

BewareMyPower:
@RobertIndie PTAL again.

lib/Future.h
```diff
@@ -19,162 +19,133 @@
...
+        const auto &pair = future_.get();
+        result = pair.second;
+        return pair.first;
+    }
 
-template <typename Result, typename Type>
-class Promise {
-   public:
-    Promise() : state_(std::make_shared<InternalState<Result, Type> >()) {}
+    // Only for test
```
RobertIndie, line 74:
```suggestion
    // Only public for test
```

RobertIndie (approved):
LGTM!

PR #300, opened 2023-07-10
[fix] Fix the assertion of `ConsumerWrapper`.receiveAtMost

RobertIndie:
### Motivation

Currently, the `ConsumerWrapper`.receiveAtMost doesn't wait for timeout after receiving enough messages. It couldn't assert that the consumer receives at most N messages. The test will be passed if we assert `receiveAtMost(3)` even if the consumer is able to receive 10 messages.

### Modifications

- Assert for timeout for the next message after receiving enough messages.

### Verifying this change

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/ConsumerWrapper.h
+++ b/tests/ConsumerWrapper.h
@@ -65,6 +65,7 @@ class ConsumerWrapper {
             ASSERT_EQ(ResultOk, consumer_.receive(msg, 3000));
             messageIdList_.emplace_back(msg.getMessageId());
         }
+        ASSERT_EQ(ResultTimeout, consumer_.receive(msg, 1000));
     }
 
     unsigned long getNumAcked(CommandAck_AckType ackType) const;


BewareMyPower (commented):
It's designed to be the current behavior. We can receive some messages first, then try to receive the rest messages.

You'd better add the assertion explicitly after calling `receiveAtMost`. From the references of `receiveAtMost` (in `AcknowledgeTest`), I think we don't need to assert there is no more messages.

And as you can see, `consumer_.receive(msg, 1000)` will take at least 1 second, which makes the test slow.

BTW,If you want to assert no more messages, you can send a message explicitly and receive it. It's faster than waiting for the 1 second timeout.

RobertIndie:
> It's designed to be the current behavior. We can receive some messages first, then try to receive the rest messages.
> You'd better add the assertion explicitly after calling receiveAtMost. From the references of receiveAtMost (in AcknowledgeTest), I think we don't need to assert there is no more messages

Currently, the `receiveAtMost` doesn't implement the 'at most' guarantee. This would confuse the developer.
I think we could have two options:
1. Implement the at most guarantee for `receiveAtMost`
2. Change it to `receiveAtLeast`

BewareMyPower:
Oh it's right. But `receiveAtLeast` is also wrong because this method cannot receive over N messages.

The proper implementation of `receiveAtMost` might be:

```c++
    Result receiveAtMost(int numMessages) {
        Message msg;
        for (int i = 0; i < numMessages; i++) {
            auto result = consumer_.receive(msg, 3000);
            if (result != ResultOk) {
                return result;
            }
            messageIdList_.emplace_back(msg.getMessageId());
        }
        return ResultOk;
    }
```

BewareMyPower:
If this method never return messages whose count is other than N, both `receiveAtMost` and `receiveLeast` are not good names.

> It seems the only difference is that we move the assertion out of this method. 

Yes, it now can return messages whose count is less or equal to N. Users can check the size to determine whether there are al least N messages. For example, if you sent 5 messages and `receiveAtMost(5)` returns only `[0, 1, 2]`, the test should fail. You can either check the returned result or the  `messageIdList_` field.

shibd:
For now, the code does not match the PR described. Can you update? @RobertIndie

Issue #306, opened 2023-08-26
[Performance] Unexpected copy of OpSendMsg


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

3.3.0

### Minimal reproduce step

Modify the `OpSendMsg` to:

```c++
    OpSendMsg() { std::cout << "OpSendMsg(): " << sizeof(*this) << " " << sizeof(metadata_) << " bytes\n"; }
    OpSendMsg(const proto::MessageMetadata& metadata, const SharedBuffer& payload,
              const SendCallback& sendCallback, uint64_t producerId, uint64_t sequenceId, int sendTimeoutMs,
              uint32_t messagesCount, uint64_t messagesSize, ChunkMessageIdImplPtr chunkedMessageId = nullptr)
        : metadata_(metadata),  // the copy happens here because OpSendMsg of chunks are constructed with
                                // a shared metadata object
          payload_(payload),
          sendCallback_(sendCallback),
          producerId_(producerId),
          sequenceId_(sequenceId),
          timeout_(TimeUtils::now() + milliseconds(sendTimeoutMs)),
          messagesCount_(messagesCount),
          messagesSize_(messagesSize),
          chunkedMessageId_(chunkedMessageId) {
        std::cout << "OpSendMsg metadata: " << sizeof(metadata_) << " bytes" << std::endl;
    }

    OpSendMsg(const OpSendMsg& rhs)
        : metadata_(rhs.metadata_),
          payload_(rhs.payload_),
          sendCallback_(rhs.sendCallback_),
          producerId_(rhs.producerId_),
          sequenceId_(rhs.sequenceId_),
          timeout_(rhs.timeout_),
          messagesCount_(rhs.messagesCount_),
          messagesSize_(rhs.messagesSize_),
          trackerCallbacks_(rhs.trackerCallbacks_),
          chunkedMessageId_(rhs.chunkedMessageId_) {
        std::cout << "OpSendMsg copy" << std::endl;
    }
```

Modify the `example/SampleProducer.cc` to:

```c++
    Producer producer;
    ProducerConfiguration conf;
    conf.setBatchingMaxMessages(5);
    Result result = client.createProducer("persistent://public/default/my-topic", conf, producer);
    if (result != ResultOk) {
        LOG_ERROR("Error creating producer: " << result);
        return -1;
    }

    for (int i = 0; i < 100; i++) {
[...]

PR #308, opened 2023-08-28
Avoid copying OpSendMsg when sending messages

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/306

### Motivation

`OpSendMsg` is a struct whose size is 400 bytes. We should avoid the copy operation on it.

### Modifications

Pass the `unique_ptr<OpSendMsg>` everywhere instead of `OpSendMsg`.
- Use `unique_ptr<OpSendMsg>` as the element of the pending message queue in `ProducerImpl` and disable the copy constructor and assignment for `OpSendMsg`.
- Add `SendArgument`, which includes the necessary fields to construct a `CommandSend` request. Use `shared_ptr` rather than `unique_ptr` to store `SendArgument` in `OpSendMsg` because the producer might need to resend the message so the `SendArgument` object could be shared by `ProducerImpl` and `ClientConnection`.

This patch is more like a refactor because the compiler optimization might reduce unnecessary copying.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/BatchMessageContainerBase.cc
+++ b/lib/BatchMessageContainerBase.cc
@@ -37,23 +37,13 @@ BatchMessageContainerBase::BatchMessageContainerBase(const ProducerImpl& produce
       producerId_(producer.producerId_),
       msgCryptoWeakPtr_(producer.msgCrypto_) {}
 
-Result BatchMessageContainerBase::createOpSendMsgHelper(OpSendMsg& opSendMsg,
-                                                        const FlushCallback& flushCallback,
-                                                        const MessageAndCallbackBatch& batch) const {
-    opSendMsg.sendCallback_ = batch.createSendCallback();
-    opSendMsg.messagesCount_ = batch.messagesCount();
-    opSendMsg.messagesSize_ = batch.messagesSize();
+BatchMessageContainerBase::~BatchMessageContainerBase() {}
 
-    if (flushCallback) {
-        auto sendCallback = opSendMsg.sendCallback_;
-        opSendMsg.sendCallback_ = [sendCallback, flushCallback](Result result, const MessageId& id) {
-            sendCallback(result, id);
-            flushCallback(result);
-        };
-    }
-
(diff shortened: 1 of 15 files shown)


BewareMyPower:
Before this patch, `~OpSendMsg` could be called in various places (Search `.*~OpSendMsg.*` in the FlameGraph):
- Directly in `ProducerImpl::ackReceived` (4.69%)
- `pendingMessagesQueue_.pop_front()` in `ProducerImpl::ackReceived` (0.47%)
- Destructor of an Asio callback (`std::tuple<std::shared_ptr<pulsar::ClientConnection>, pulsar::OpSendMsg>`) (1.56%)
- `BatchMessageContainerBase::processAndClear` (0.78%)
-  Directly in `ackReceived` (another call in event loop) (0.16%) 

graph-3 3 0

After this patch, `~OpSendMsg` could only be called in `pendingMessagesQueue_.pop_front()` (3.65%)

graph

RobertIndie (approved):

lib/BatchMessageContainerBase.cc
```diff
@@ -70,45 +60,18 @@ Result BatchMessageContainerBase::createOpSendMsgHelper(OpSendMsg& opSendMsg,
...
-            std::vector<OpSendMsg> opSendMsgs;
-            std::vector<Result> results = createOpSendMsgs(opSendMsgs, flushCallback);
-            for (size_t i = 0; i < results.size(); i++) {
-                opSendMsgCallback(results[i], opSendMsgs[i]);
-            }
-        }  // else numBatches is 0, do nothing
-    }
-    clear();
+    return OpSendMsg::create(impl->metadata, batch.messagesCount(), batch.messagesSize(),
+                             producerConfig_.getSendTimeout(), batch.createSendCallback(flushCallback),
```
shibd, line 73: Can use `sendCallback` to replace `batch.createSendCallback(flushCallback)`?

PR #313, opened 2023-09-11
[fix] Add the curl wrapper to avoid inconsistent curl options

BewareMyPower:
### Motivation

When libcurl is used in `AuthOauth2`, the `CURLOPT_NOSIGNAL` option is not set, i.e. it will be the default value so that the `Curl_resolv_timeout` function might crash in multi-threading environment.

```
#2 0xf630 in _L_unlock_13 from /lib64/libpthread.so.0 (0x34)
#3 0x2e6c7f in Curl_failf from /usr/local/bin/***/libpulsar.so (0x6f)
#4 0x30a285 in Curl_resolv_timeout from /usr/local/bin/***/libpulsar.so (0x95)
```

Since there are many duplicated code when calling curl C APIs, it's hard to notice that `CURLOPT_NOSIGNAL` is not configured in `AuthOauth2`.

### Modifications

Introduce a `CurlWrapper` class that sets the same options to reduce the duplicated code and adapting consistent behaviors unless a few options.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/CurlWrapper.h
+++ b/lib/CurlWrapper.h
@@ -1,0 +1,180 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


BewareMyPower:
The CI for wireshark build on macOS is broken. Let's ignore it for now. I have opened an issue: https://github.com/apache/pulsar-client-cpp/issues/314

shibd:
> The CI for wireshark build on macOS is broken. Let's ignore it for now. I have opened an issue: #314

Let's fix these CI first. Ignoring failed ci is not a good idea.

BewareMyPower:
> Let's fix these CI first. Ignoring failed ci is not a good idea.

They are not caused by the code issue. It should not block any fix.

RobertIndie (approved):

shibd (approved):

PR #317, opened 2023-09-18
Avoid accessing a null ClientConnection instance

BewareMyPower:
### Motivation

We observed server null `ClientConnection` accesses in test environment. See the `this=0x0` outputs in the following two typical stacks.

```
#8  bytesWritten (this=0xb8, size=371) at lib/SharedBuffer.h:166
#9  pulsar::ClientConnection::handleRead (this=0x0, err=..., bytesTransferred=371, minReadSize=4) at lib/ClientConnection.cc:609
```

```
#12 0x00007f33202933d2 in unique_lock (__m=..., this=0x7f3311c82800) at /opt/rh/devtoolset-7/root/usr/include/c++/7/bits/std_mutex.h:197
#13 pulsar::ClientConnection::sendPendingCommands (this=0x0) at lib/ClientConnection.cc:1071
#14 0x00007f3320293d2d in pulsar::ClientConnection::handleSendPair (this=0x0, err=...) at lib/ClientConnection.cc:1066
```

Though `shared_from_this()` is always passed to the `std::bind` function, when the method of `ClientConnection` is called, the pointer is still `null`.

### Modifications

First, replace all `std::bind` calls with the lambda expression that catches `std::weak_ptr<ClientConnection>` and perform null checks explicitly on the value returned by the `lock()` method.

Since now all asio callbacks don't hold a `shared_ptr`, the owner of the `ClientConnection` object should be `ConnectionPool`, i.e. the pool maintains some connections, while all asio callbacks use `weak_ptr` to test if the connection is present.

Second, make `ClientConnection::getConnection` return `shared_ptr` rather than `weak_ptr` so that the caller side does not need to check if `lock()` returns null in the callback of this future.

We cannot make `ConnectionPool::getConnectionAsync` return `shared_ptr` because it could return the future of `connectPromise_`, which is hold by `ClientConnection` itself. We should avoid holding a `shared_ptr` of `ClientConnection` because its owner is `ConnectionPool`.

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -24,6 +24,7 @@
 #include <fstream>
 
 #include "Commands.h"
+#include "ConnectionPool.h"
 #include "ConsumerImpl.h"
 #include "ExecutorService.h"
 #include "LogUtils.h"
@@ -157,7 +158,8 @@ std::atomic<int32_t> ClientConnection::maxMessageSize_{Commands::DefaultMaxMessa
 ClientConnection::ClientConnection(const std::string& logicalAddress, const std::string& physicalAddress,
                                    ExecutorServicePtr executor,
                                    const ClientConfiguration& clientConfiguration,
-                                   const AuthenticationPtr& authentication, const std::string& clientVersion)
+                                   const AuthenticationPtr& authentication, const std::string& clientVersion,
+                                   ConnectionPool& pool)
     : operationsTimeout_(seconds(clientConfiguration.getOperationTimeoutSeconds())),
       authentication_(authentication),
       serverProtocolVersion_(proto::ProtocolVersion_MIN),
@@ -180,7 +182,8 @@ ClientConnection::ClientConnection(const std::string& logicalAddress, const std:
       outgoingBuffer_(SharedBuffer::allocate(DefaultBufferSize)),
       consumerStatsRequestTimer_(executor_->createDeadlineTimer()),
       maxPendingLookupRequest_(clientConfiguration.getConcurrentLookupRequest()),
-      clientVersion_(clientVersion) {
+      clientVersion_(clientVersion),
+      pool_(pool) {
     LOG_INFO(cnxString_ << "Create ClientConnection, timeout=" << clientConfiguration.getConnectionTimeout());
     if (clientConfiguration.isUseTls()) {
 #if BOOST_VERSION >= 105400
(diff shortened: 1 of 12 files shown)


lib/ClientConnection.cc
```diff
@@ -597,10 +636,15 @@ void ClientConnection::handleResolve(const boost::system::error_code& err,
 
 void ClientConnection::readNextCommand() {
     const static uint32_t minReadSize = sizeof(uint32_t);
+    auto weakSelf = weak_from_this();
     asyncReceive(
         incomingBuffer_.asio_buffer(),
-        customAllocReadHandler(std::bind(&ClientConnection::handleRead, shared_from_this(),
```
shibd, line 602: This is the root cause, right? ClientConnection object may have been destroyed before calling `handleRead`

BewareMyPower: I think not. Ideally, If `ClientConnection` is destroyed when calling `shared_from_this()`, `std::bad_weak_ptr` will be thrown. See the following example:

```c++
#include <memory>
#include <iostream>

struct Demo : std::enable_shared_from_this<Demo> {
  void f() {
    auto self = shared_from_this();
    std::cout << self->x << std::endl;
  }
  int x = 100;
};

int main() {
  auto demo = std::make_shared<Demo>();
  auto p = demo.get();
  demo.reset();
  p->f();
}
```

I built it and ran on CentOS 7 with GCC 7 and saw:

```
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted
```

In addition, `readNextCommand` is only called in `handleSentPulsarConnect`, which is also called in the `std::bind` wrapper with `shared_from_this()`:

```c++
     asyncWrite(buffer.const_asio_buffer(), std::bind(&ClientConnection::handleSentPulsarConnect,
                                                      shared_from_this(), std::placeholders::_1, buffer));
```

So when `handleSentPulsarConnect` is called, the `shared_ptr<ClientConnection>` should not be expired.

----

The analysis above is based on the assumption that there is nothing wrong with the following use:

```c++
// When f() is called, since shared_from_this() is passed to std::bind, the ClientConnection object should be valid
async_xxx(xxx, std::bind(&ClientConnection::f, shared_from_this(), args...));
```

So I suspect that there might be something wrong with how we use `std::bind`. Replacing it with lambda is more efficient, readable and **the null check is performed explicitly**.

BewareMyPower: As you can see, `handleSendPair` was also called on a null object.

shibd: Ok, I agree to use `lambda` instead of `std::bind`.

shibd (approved):

merlimat (approved):

Issue #325, opened 2023-10-08
[Bug] Segmentation fault when sending messages after receiving an error 


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

- Pulsar 3.1.0
- Client: https://github.com/apache/pulsar-client-cpp/commit/eea59bb4458c6ea190f51732d41fdb32c9ad6327

Additional broker configs:

```properties
brokerDeduplicationEnabled=true
maxMessageSize=1024000
```

### Minimal reproduce step

To simulate the send error is returned, we need to run the following code on a Pulsar release that does not include https://github.com/apache/pulsar/pull/20948

```c++
#include <pulsar/Client.h>

#include <chrono>
#include <iostream>
#include <thread>

using namespace pulsar;

int main() {
    Client client("pulsar://127.0.0.1:6650");

    Producer producer;
    ProducerConfiguration conf;
    conf.setBatchingEnabled(false);
    conf.setChunkingEnabled(true);
    client.createProducer("persistent://public/default/my-topic", conf, producer);

    producer.sendAsync(MessageBuilder().setContent(std::string(1024000 * 5, 'a')).build(),
                       [](Result result, const MessageId& msgId) {
                           std::cout << "XYZ send " << result << " " << msgId << std::endl;
                       });
    std::this_thread::sleep_for(std::chrono::seconds(1));

    client.close();
}
```

### What did you expect to see?

The application exits normally.

### What did you see instead?

```
2023-10-08 11:41:38.699 WARN  [0x16f05b000] ClientConnection:1529 | [127.0.0.1:55532 -> 127.0.0.1:6650] Received send error from server: Cannot determine whether the message is a duplicate at this time
2023-10-08 11:41:38.699 INFO  [0x16f05b000] ClientConnection:1340 | [127.0.0.1:55532 -> 127.0.0.1:6650] Connection disconnected
2023-10-08 11:41:38.699 INFO  [0x16f05b000] HandlerBase:141 | [persistent://public/default/test-send-chunks, standalone-12-4] Schedule reconnection in 0.1 s
2023-10-08 11:41:38.699 INFO  [0x16f05b000] ConnectionPool:122 | Remove connection for pulsar://127.0.0.1:6650
[...]

PR #326, opened 2023-10-08
Fix segmentation fault when sending messages after receiving an error

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/325

### Motivation

https://github.com/apache/pulsar-client-cpp/pull/317 introduces a bug that might cause segmentation fault when sending messages after receiving an error, see
https://github.com/apache/pulsar-client-cpp/issues/325#issuecomment-1751914150 for the detailed explanation.

### Modifications

When calling `asyncWrite`, capture the `shared_ptr` instead of the `weak_ptr` to extend the lifetime of the `socket_` or `tlsSocket_` field in `ClientConnection`. Since the lifetime is extended, in some callbacks, check `isClosed()` before other logic.

Add a `ChunkDedupTest` to reproduce this issue based on Pulsar 3.1.0. Run the test for 10 times to ensure it won't crash after this patch.

### DIFF
--- a/tests/chunkdedup/docker-compose.yml
+++ b/tests/chunkdedup/docker-compose.yml
@@ -1,0 +1,44 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


tests/chunkdedup/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+# under the License.
+#
+
+version: '3'
+networks:
+  pulsar:
+    driver: bridge
+services:
+  standalone:
+    image: apachepulsar/pulsar:3.1.0 # Ensure https://github.com/apache/pulsar/pull/20948 is not included
```
RobertIndie, line 26: It's better to add a comment like "Don't change the pulsar version here.". This test could only work before 3.1.0, right?

RobertIndie (approved):

shibd (approved):

tests/chunkdedup/docker-compose.yml
BewareMyPower: Yes.

PR #332, opened 2023-10-20
Avoid blocking the message listener threads

erobot:
### Motivation

The message listener thread blocks when the receiver queue of `MultiTopicsConsumerImpl` is full. As message listener threads are used by all consumers in the same `Client`, if one slow consumer blocks the listener threads, all other consuemrs can no longer receive new messages.

### Modifications

1. Modify `MultiTopicsConsumerImpl` to use `UnboundedBlockingQueue` to avoid blocking
2. Modify the permit update logic: Increase permit only after messages consumed from `MultiTopicsConsumerImpl`

### Verifying this change

  - Add a test `testNoListenerThreadBlocking`

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
It's just an optimization.

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/MessageImpl.h
+++ b/lib/MessageImpl.h
@@ -47,6 +47,7 @@ class MessageImpl {
     int redeliveryCount_;
     bool hasSchemaVersion_;
     const std::string* schemaVersion_;
+    std::shared_ptr<class ConsumerImpl> consumerPtr_;
 
     const std::string& getPartitionKey() const;
     bool hasPartitionKey() const;
--- a/lib/MessageImpl.h
+++ b/lib/MessageImpl.h
@@ -47,7 +47,7 @@ class MessageImpl {
     int redeliveryCount_;
     bool hasSchemaVersion_;
     const std::string* schemaVersion_;
-    std::shared_ptr<class ConsumerImpl> consumerPtr_;
+    std::weak_ptr<class ConsumerImpl> consumerPtr_;
 
     const std::string& getPartitionKey() const;
     bool hasPartitionKey() const;
--- a/tests/ConsumerTest.cc
+++ b/tests/ConsumerTest.cc
@@ -1335,4 +1335,67 @@ TEST(ConsumerTest, testRetrySubscribe) {
     // milliseconds
 }
 
+TEST(ConsumerTest, testNoListenerThreadBlocking) {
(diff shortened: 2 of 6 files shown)


tests/ConsumerTest.cc
```diff
@@ -1335,4 +1335,67 @@ TEST(ConsumerTest, testRetrySubscribe) {
...
+    }
+    producer.flush();
+    producer.close();
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    // check consumer1 prefetch num
+    auto multiConsumerImpl = PulsarFriend::getMultiTopicsConsumerImplPtr(consumer1);
+    int prefetchNum = multiConsumerImpl->getNumOfPrefetchedMessages();
+    ASSERT_LE(prefetchNum, receiverQueueSizeAcrossPartitions);
```
BewareMyPower, line 1382: You can use `waitUntil` instead of the trivial sleep.

```c++
    waitUntil(std::chrono::seconds(1), [consumer1] {
        auto multiConsumerImpl = PulsarFriend::getMultiTopicsConsumerImplPtr(consumer1);
        return multiConsumerImpl->getNumOfPrefetchedMessages() == receiverQueueSizeAcrossPartitions;
    });
```

lib/MessageImpl.h
```diff
@@ -47,6 +47,7 @@ class MessageImpl {
     int redeliveryCount_;
     bool hasSchemaVersion_;
     const std::string* schemaVersion_;
+    std::shared_ptr<class ConsumerImpl> consumerPtr_;
```
BewareMyPower, line 50: Should you store a `std::weak_ptr`? If we have N `Message` objects, the reference count of the `ConsumerImpl` will be at least `N + 1`, which is unnecessary and it might extend the lifetime of internal consumers unexpectedly.

tests/ConsumerTest.cc
erobot: Fixed

BewareMyPower (approved):

lib/MessageImpl.h
erobot: Changed to `std::weak_ptr`.

PR #336, opened 2023-10-30
Added support for multiple connections to each broker

### LINKED ISSUE #221: [feature request] Support the connectionsPerBroker config
See https://github.com/apache/pulsar/blob/e0098ee0c34643371f53ee921bdc361e163f6387/pulsar-client-api/src/main/java/org/apache/pulsar/client/api/ClientBuilder.java#L275

Currently the connection pool is enabled and cannot be disabled because the constructor is private. https://github.com/apache/pulsar-client-cpp/blob/998c73df164f7e08a5e4e4b33f9d979db4518e07/include/pulsar/Client.h#L408-L409

So if there are many producers created by the same `Client`, including the case that a topic has many partitions, all producers will share the same TCP connection and the number of send requests might reach the limit: https://github.com/apache/pulsar/blob/e0098ee0c34643371f53ee921bdc361e163f6387/pulsar-broker-common/src/main/java/org/apache/pulsar/broker/ServiceConfiguration.java#L681

merlimat:
Fixes #221

### Motivation

Support having more than 1 connection to each broker.

### DIFF
--- a/lib/ClientConfiguration.cc
+++ b/lib/ClientConfiguration.cc
@@ -40,6 +40,13 @@ ClientConfiguration& ClientConfiguration::setMemoryLimit(uint64_t memoryLimitByt
 
 uint64_t ClientConfiguration::getMemoryLimit() const { return impl_->memoryLimit; }
 
+ClientConfiguration& ClientConfiguration::setConnectionsPerBroker(int connectionsPerBroker) {
+    impl_->connectionsPerBroker = connectionsPerBroker;
+    return *this;
+}
+
+int ClientConfiguration::getConnectionsPerBroker() const { return impl_->connectionsPerBroker; }
+
 ClientConfiguration& ClientConfiguration::setAuth(const AuthenticationPtr& authentication) {
     impl_->authenticationPtr = authentication;
     return *this;
--- a/lib/ClientConfiguration.cc
+++ b/lib/ClientConfiguration.cc
@@ -41,6 +41,9 @@ ClientConfiguration& ClientConfiguration::setMemoryLimit(uint64_t memoryLimitByt
 uint64_t ClientConfiguration::getMemoryLimit() const { return impl_->memoryLimit; }
 
 ClientConfiguration& ClientConfiguration::setConnectionsPerBroker(int connectionsPerBroker) {
+    if (connectionsPerBroker <= 0) {
+        throw std::invalid_argument("connectionsPerBroker should be greater than 0");
+    }
     impl_->connectionsPerBroker = connectionsPerBroker;
     return *this;
 }
--- a/lib/ConnectionPool.cc
+++ b/lib/ConnectionPool.cc
@@ -34,15 +34,13 @@ DECLARE_LOG_OBJECT()
 namespace pulsar {
 
 ConnectionPool::ConnectionPool(const ClientConfiguration& conf, ExecutorServiceProviderPtr executorProvider,
-                               const AuthenticationPtr& authentication,
-                               const std::string& clientVersion)
+                               const AuthenticationPtr& authentication, const std::string& clientVersion)
     : clientConfiguration_(conf),
       executorProvider_(executorProvider),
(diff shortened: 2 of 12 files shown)


lib/ConnectionPool.cc
```diff
@@ -34,13 +34,13 @@ DECLARE_LOG_OBJECT()
...
-                               const AuthenticationPtr& authentication, bool poolConnections,
-                               const std::string& clientVersion)
+                               const AuthenticationPtr& authentication, const std::string& clientVersion)
     : clientConfiguration_(conf),
       executorProvider_(executorProvider),
       authentication_(authentication),
-      poolConnections_(poolConnections),
-      clientVersion_(clientVersion) {}
+      clientVersion_(clientVersion),
+      randomDistribution_(0, conf.getConnectionsPerBroker()),
```
BewareMyPower, line 42:
```suggestion
      randomDistribution_(0, conf.getConnectionsPerBroker() + 1),
```

The `uniform_int_distribution` range is left-closed and right-open

BewareMyPower (changes requested):

lib/ClientConfiguration.cc
```diff
@@ -40,6 +40,13 @@ ClientConfiguration& ClientConfiguration::setMemoryLimit(uint64_t memoryLimitByt
 
 uint64_t ClientConfiguration::getMemoryLimit() const { return impl_->memoryLimit; }
 
+ClientConfiguration& ClientConfiguration::setConnectionsPerBroker(int connectionsPerBroker) {
+    impl_->connectionsPerBroker = connectionsPerBroker;
```
BewareMyPower, line 44: Should we add an invalidation here?

```c++
    if (connectionsPerBroker <= 0) {
        throw std::invalid_argument("connectionsPerBroker should be greater than 0");
    }
```

RobertIndie: +1

lib/ConnectionPool.cc
RobertIndie: From the [doc](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution), the range is left-closed and right-closed: [a,b].

RobertIndie, line 42:
```suggestion
      randomDistribution_(0, conf.getConnectionsPerBroker() -1),
```

The range is left-closed and right-closed. We need to `-1` here to make sure the total number of the connections to be equal to `conf.getConnectionsPerBroker()`.

BewareMyPower: My fault. It's true that `uniform_int_distribution` uses right-closed range.

I wanted to say we should use a left-closed and right-open range: `[0, connectionsPerBroker)` so the initialization of `uniform_int_distribution` should be:

```c++
randomDistribution_(0, conf.getConnectionsPerBroker() - 1),
```

@RobertIndie

BewareMyPower: Yeah this comment is right.

lib/ConnectionPool.cc
```diff
@@ -107,7 +108,7 @@ Future<Result, ClientConnectionWeakPtr> ConnectionPool::getConnectionAsync(
     LOG_INFO("Created connection for " << logicalAddress);
```
BewareMyPower, line 108: Maybe we can print `key` instead of `logicalAddress` now. Otherwise users might not know which specific connection is created for `logicalAddress`.

```c++
LOG_INFO("Created connection for " << key);
```

merlimat: It could be confusing though to see a `broker-5-N` in the log and will be mistaken by a different host name

RobertIndie (approved):

lib/ConnectionPool.cc
BewareMyPower: Okay. It makes sense.

BewareMyPower (approved):

Issue #346, opened 2023-11-15
[Bug] ConnectionPool::remove Crash


arhoads:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

Commit 6f115e76da42ffbcf39a5fbe1df298e6d6635f4d

### Minimal reproduce step

`ClientConnection::close` makes the function call `pool_.remove(logicalAddress_, this);`. LogicalAddress is no longer the correct key in the connection pool and instead should be logical address + "- {keySuffix}". The key isn't found in the pool map and a crash happens because the find result isn't validated.

### What did you expect to see?

No crash

### What did you see instead?

Crash

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #347, opened 2023-11-16
Fix crash when removing connection from the pool

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/346

### Motivation

https://github.com/apache/pulsar-client-cpp/pull/336 changes the key of the `ClientConnection` in `ConnectionPool`, while in `ClientConnection::close`, it still passes the old key (logical address) to `ConnectionPool::remove`, which results in the connection could never be removed and destroyed until being deleted as a stale connection.

What's worse, if the key does not exist, the iterator returned by `std::map::find` will still be dereferenced, which might cause crash in some platforms. See
https://github.com/apache/pulsar-client-cpp/blob/8d32fd254e294d1fabba73aed70115a434b341ef/lib/ConnectionPool.cc#L122-L123

### Modifications

- Avoid dereferencing the iterator if it's invalid in `ConnectionPool::remove`.
- Store the key suffix in `ClientConnection` and pass the correct key to `ConnectionPool::remove` in `ClientConnection::close`
- Add `ClientTest.testConnectionClose` to verify `ClientConnection::close` can remove itself from the pool and the connection will be destroyed eventually.

### DIFF
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -160,7 +160,7 @@ ClientConnection::ClientConnection(const std::string& logicalAddress, const std:
                                    ExecutorServicePtr executor,
                                    const ClientConfiguration& clientConfiguration,
                                    const AuthenticationPtr& authentication, const std::string& clientVersion,
-                                   ConnectionPool& pool)
+                                   ConnectionPool& pool, size_t poolIndex)
     : operationsTimeout_(seconds(clientConfiguration.getOperationTimeoutSeconds())),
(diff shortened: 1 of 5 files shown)


BewareMyPower:
More context: https://stackoverflow.com/questions/16267615/is-second-defined-for-iterator-stdmapend

Dereferencing an iterator that points to the end of a STL container is UB.

RobertIndie (approved):

Issue #349, opened 2023-11-16
[Bug] Segmentation fault if failed to load an authentication plugin


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

https://github.com/apache/pulsar-client-cpp/commit/8d32fd254e294d1fabba73aed70115a434b341ef

### Minimal reproduce step

```c++
    Client client("pulsar://localhost:6650", ClientConfiguration{}.setAuth(AuthFactory::create("invalid", "invalid")));
    Producer producer;
    std::cout << client.createProducer("my-topic", producer) << std::endl;
``

### What did you expect to see?

```
AuthenticationError
```

### What did you see instead?

```
2023-11-16 19:26:46.795 WARN  [0x1dd74de00] Authentication:190 | Couldn't load auth plugin invalid
2023-11-16 19:26:46.796 WARN  [0x1dd74de00] Authentication:190 | Couldn't load auth plugin invalid
2023-11-16 19:26:46.796 INFO  [0x1dd74de00] ClientConnection:188 | [<none> -> pulsar://localhost:6650] Create ClientConnection, timeout=10000
2023-11-16 19:26:46.796 INFO  [0x1dd74de00] ConnectionPool:109 | Created connection for pulsar://localhost:6650
2023-11-16 19:26:46.798 INFO  [0x16eef7000] ClientConnection:398 | [127.0.0.1:53863 -> 127.0.0.1:6650] Connected to broker
zsh: segmentation fault  ./tests/pulsar-tests --gtest_filter='*.testConnectionConstructFailure'
```

### Anything else?

_No response_

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

Issue #348, opened 2023-11-16
Test `test_auth_junk_params` failed after upgrading C++ client to 3.4.0 


RobertIndie:
I created [a PR](https://github.com/apache/pulsar-client-python/pull/161) to upgrade the C++ client to 3.4.0, but the CI failed: 

```
+ python3 pulsar_test.py
2023-11-15 00:23:08.410 INFO  [140130423730624] ClientConnection:188 | [<none> -> pulsar://localhost:6650] Create ClientConnection, timeout=10000
2023-11-15 00:23:08.410 INFO  [140130423730624] ConnectionPool:109 | Created connection for pulsar://localhost:6650
2023-11-15 00:23:08.412 INFO  [140130379408960] ClientConnection:398 | [[::1]:49850 -> [::1]:6650] Connected to broker
2023-11-15 00:23:08.416 INFO  [140130379408960] HandlerBase:83 | [persistent://public/default/test_acknowledge_failed, ] Getting connection from pool
2023-11-15 00:23:08.447 INFO  [140130379408960] ProducerImpl:216 | [persistent://public/default/test_acknowledge_failed, ] Created producer on broker [[::1]:49850 -> [::1]:6650] 
2023-11-15 00:23:08.447 INFO  [140130423730624] Client:86 | Subscribing on Topic :test_acknowledge_failed
2023-11-15 00:23:08.449 INFO  [140130379408960] HandlerBase:83 | [persistent://public/default/test_acknowledge_failed, sub1, 0] Getting connection from pool
2023-11-15 00:23:08.468 INFO  [140130379408960] ConsumerImpl:297 | [persistent://public/default/test_acknowledge_failed, sub1, 0] Created consumer on broker [[::1]:49850 -> [::1]:6650] 
2023-11-15 00:23:08.469 INFO  [140130423730624] Client:86 | Subscribing on Topic :test_acknowledge_failed
2023-11-15 00:23:08.472 INFO  [140130379408960] HandlerBase:83 | [persistent://public/default/test_acknowledge_failed, sub2, 1] Getting connection from pool
2023-11-15 00:23:08.4[80](https://github.com/apache/pulsar-client-python/actions/runs/6848544221/job/18687459291?pr=161#step:9:81) INFO  [140130379408960] ConsumerImpl:297 | [persistent://public/default/test_acknowledge_failed, sub2, 1] Created consumer on broker [[::1]:49850 -> [::1]:6650] 
[...]

PR #350, opened 2023-11-16
Fix bad_weak_ptr when close a ClientConnection during construction

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/348

Fixes https://github.com/apache/pulsar-client-cpp/issues/349

### Motivation

When `close` is called in `ClientConnection`'s constructor,
`shared_from_this()` will be called, which results in a
`std::bad_weak_ptr` error. This error does not happen before
https://github.com/apache/pulsar-client-cpp/pull/317 because
`shared_from_this()` could only be called when the `producers` or
`consumers` field is not empty.

### Modifications

Add a `ResultException` and throw it if there is a failure in
`ClientConnection`'s constructor and catch it in
`ConnectionPool::getConnectionAsync`. Then retrieve the result and
return the failed future.

Since `close` is now always called on a constructed `ClientConnection`
object, remove the 2nd parameter.

In addition, check `authentication_` even for non-TLS URLs. Otherwise,
the null authentication will be used to construct `CommandConnect`.

Add `testInvalidPlugin` and `testTlsConfigError` to verify the changes.

### DIFF
--- a/lib/ClientConnection.h
+++ b/lib/ClientConnection.h
@@ -148,9 +148,7 @@ class PULSAR_PUBLIC ClientConnection : public std::enable_shared_from_this<Clien
      * @param result all pending futures will complete with this result
      * @param detach remove it from the pool if it's true
      *
-     * `detach` should only be false when:
-     * 1. Before the connection is put into the pool, i.e. during the construction.
-     * 2. When the connection pool is closed
+     * `detach` should only be false when the connection pool is closed.
      */
     void close(Result result = ResultConnectError, bool detach = true);
 
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -187,6 +187,12 @@ ClientConnection::ClientConnection(const std::string& logicalAddress, const std:
       pool_(pool),
       poolIndex_(poolIndex) {
     LOG_INFO(cnxString_ << "Create ClientConnection, timeout=" << clientConfiguration.getConnectionTimeout());
+    if (!authentication_) {
+        LOG_ERROR("Invalid authentication plugin");
+        throw ResultAuthenticationError;
+        return;
+    }
+
     if (clientConfiguration.isUseTls()) {
 #if BOOST_VERSION >= 105400
         boost::asio::ssl::context ctx(boost::asio::ssl::context::tlsv12_client);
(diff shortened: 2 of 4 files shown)


lib/ClientConnection.h
```diff
@@ -146,13 +146,9 @@ class PULSAR_PUBLIC ClientConnection : public std::enable_shared_from_this<Clien
...
      * @param result all pending futures will complete with this result
-     * @param detach remove it from the pool if it's true
-     *
-     * `detach` should only be false when:
-     * 1. Before the connection is put into the pool, i.e. during the construction.
-     * 2. When the connection pool is closed
+     * @param constructCompleted whether the construction is completed
      */
-    void close(Result result = ResultConnectError, bool detach = true);
+    void close(Result result = ResultConnectError, bool constructCompleted = true);
```
RobertIndie, line 151: It's not safe to directly replace the `detach` parameter. There are many other places where the close method is still called with the wrong parameter meaning.
For example at `ClientConnection.cc` line 214: `close(ResultAuthenticationError, false);` is called but with the constructCompleted false, which should be true because it's in the constructor.

RobertIndie (changes requested):

lib/ClientConnection.h
BewareMyPower: I will fix them soon.

BewareMyPower: > For example at ClientConnection.cc line 214: close(ResultAuthenticationError, false); is called but with the constructCompleted false, which should be true because it's in the constructor.

`constructCompleted` means if the construction completes. If it's in the constructor, it should be false because the construction does not complete. That's the trick I made to avoid many changes.

RobertIndie: Sorry, it seems that I have given the wrong example. We need to take care of this: https://github.com/apache/pulsar-client-cpp/blob/6d47e9495c071975d94fc284d6bfb6bbe9c7501a/lib/ConnectionPool.cc#L57

This will incorrectly cause the connection closure to be skipped.

BewareMyPower: Good catch! I think we should add a separated method instead of an overload.

BewareMyPower:
@RobertIndie I used an exception-based way to fix this issue now so `close` is always called on a `ClientConnection` object that completes the construction successfully.

BewareMyPower:
The `Oauth2Test` failed, I will check the failed test first.

RobertIndie (approved):

PR #354, opened 2023-11-21
Fix Protobuf symbols not found in libpulsarwithdeps.a when building on macOS

BewareMyPower:
### Motivation

https://github.com/apache/pulsar-client-cpp/pull/290 brings a regression that on macOS, Protobuf is always found with CMake Config mode, which does not set the `Protobuf_LIBRARIES` variable so that the libpulsarwithdeps.a misses the symbols of Protobuf.

### Modifications

When `LINK_STATIC` is ON, use CMake Module mode to find the Protobuf.

Add `build-static-library.sh` to build libraries with static dependencies and verify these libraries in PR workflow. Upload the pre-built binaries in the build workflow.

### DIFF
--- a/.github/workflows/ci-build-binary-artifacts.yaml
+++ b/.github/workflows/ci-build-binary-artifacts.yaml
@@ -192,3 +192,34 @@ jobs:
         with:
           name: ${{ matrix.triplet }}-Debug
           path: ${{ env.INSTALL_DIR }}-Debug
+
+  package-macos:
+    name: Build macOS libraries
+    runs-on: macos-latest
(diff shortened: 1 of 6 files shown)


BewareMyPower:
I ran a workflow to generate macOS artifacts here: https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/6943109321

You can check https://github.com/apache/pulsar-client-cpp/wiki/Verify-the-candidate-release-in-your-local-env#macos-users for how to use it. These artifacts will be included in the next candidate release.

shibd (approved):

Issue #359, opened 2023-12-01
[Bug] Crash for regex consumer


astifter:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

3.3.0

### Minimal reproduce step

Use regex subscription and simulate instable network connection.

### What did you expect to see?

No crash.

### What did you see instead?

Crash, see backtraces.
[pulsar_crash.bt2.log](https://github.com/apache/pulsar-client-cpp/files/13524206/pulsar_crash.bt2.log)
[pulsar_crash.bt1.log](https://github.com/apache/pulsar-client-cpp/files/13524208/pulsar_crash.bt1.log)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Issue #358, opened 2023-11-30
[Bug] std::system_error thrown during/after client->subscribe(...) when using configuration setUnAckedMessagesTimeoutMs.


jato-sag:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

Pulsar version 3.3.
OS - Red Hat Enterprise Linux 8.9 and RHEL 9.3 and Linux 5.10.0-26-amd64

### Minimal reproduce step

Our test that is observing the issue is when we are unable to connect to the server immediately on startup, but the server is reachable after a while. We attempt to subscribe, if the subscribe fails we try again. 
Sometimes when the subscribe eventually succeeds all is well.
Sometimes when the subscribe is successful it still throws the exception.
Sometimes even before the subscribe returns there is the exception.

//Rough setup.
```
clientConfig.setReceiverQueueSize(1000);
clientConfig.setUnAckedMessagesTimeoutMs(10000);
clientConfig.setConsumerType(pulsar::ConsumerType::ConsumerExclusive);
client = std::shared_ptr<pulsar::Client> (new pulsar::Client(serviceURL, clientConfig));
pulsar::Result result = pulsar::ResultRetryable;
pulsar::Consumer consumer;
while (true && result != pulsar::ResultOk) {
   result = client->subscribe(["A topic", "Another topic"], subscriberName, clientConfig, consumer);
   if (result != pulsar::ResultOk) {
      //sleep 5s;
   }
}
```

### What did you expect to see?

No exception.

### What did you see instead?

std::system_error thrown and not handled.

### Anything else?

std::system_error thrown during/after client->subscribe(...) when using configuration setUnAckedMessagesTimeoutMs.

On attempt to subscribe and we have already set a value in the Client Configuration for setUnAckedMessagesTimeoutMs we are observing std::system_error being thrown, not handled, and not caught in ExecutorService and leads to terminating the application. setUnAckedMessagesTimeoutMs is set to 10000, as this is the minimum we have not experimented with other values. 
Without setUnAckedMessagesTimeoutMs set in the configuration no exception is seen on or after subscribe(..).
 
We suspect the exception is being thrown by std::recursive_mutex when trying to aquire the lock in: UnAckedMessageTrackerEnabled::timeoutHandlerHelper() of pulsar-client-cpp/lib/UnAckedMessageTrackerEnabled.cc
 
1) Please can this be investigated for the cause of the exception.
[...]

PR #362, opened 2023-12-04
Fix accessing destroyed objects in the callback of async_wait

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/358 Fixes https://github.com/apache/pulsar-client-cpp/issues/359

### Motivation

`async_wait` is not used correctly in some places. A callback that captures the `this` pointer or reference to `this` is passed to `async_wait`, if this object is destroyed when the callback is called, an invalid memory access will happen.

### Modifications

Use the following pattern in all `async_wait` calls.

```c++
std::weak_ptr<T> weakSelf{shared_from_this()};
timer_->async_wait([weakSelf](/* ... */) {
    if (auto self = weakSelf.lock()) {
        self->foo();
    }
});
```


merlimat (approved):

BewareMyPower:
It seems some tests failed. I will fix them ASAP.

Issue #363, opened 2023-12-05
[Bug] tlsTrustCertsFilePath does not work for OAuth2 authentication


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

OS: macOS m1
Client: https://archive.apache.org/dist/pulsar/pulsar-client-cpp-3.4.1/

### Minimal reproduce step

The same reproduce steps with https://github.com/apache/pulsar-client-cpp/issues/184

### What did you expect to see?

<none>

### What did you see instead?

Even if `setTlsTrustCertsFilePath` is configured, the OAuth2 authentication will still fail.

```
2023-12-05 15:30:28.066 ERROR [0x16d9c7000] AuthOauth2:390 | Response failed for issuerurl <...>. ErrorCode 60: SSL certificate problem: unable to get local issuer certificate passedin: 
```

### Anything else?

It's a regression brought by https://github.com/apache/pulsar-client-cpp/pull/313, the `tlsTrustCertsFilePath` config didn't work for OAuth2.

https://github.com/apache/pulsar-client-cpp/blob/63d494f57df126cf633396c618886eeb79946408/lib/auth/AuthOauth2.cc#L346

The last argument is `nullptr`, i.e. `CurlWrapper` does not load any CA certs for `AuthOauth2::authenticate`.

While you can see it's applied in https://github.com/apache/pulsar-client-cpp/blob/63d494f57df126cf633396c618886eeb79946408/lib/auth/AuthOauth2.cc#L232-L237

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

BewareMyPower:
The workflow [here](https://github.com/apache/pulsar-client-cpp/actions/runs/7100416866/job/19326464521?pr=365) can describe this issue more clearly.

```
2023-12-05 12:06:36.958 INFO  [140093743501632] ClientConnection:189 | [<none> -> pulsar://localhost:6650] Create ClientConnection, timeout=10000
2023-12-05 12:06:36.958 INFO  [140093743501632] ConnectionPool:114 | Created connection for pulsar://localhost:6650-0
2023-12-05 12:06:36.962 INFO  [140093735237184] ClientConnection:396 | [[::1]:51810 -> [::1]:6650] Connected to broker
[...]

PR #371, opened 2023-12-12
Integrate vcpkg to manage dependencies for all platforms

BewareMyPower:
### Motivation

Currently we manage dependencies by ourselves:
- When running unit tests in CI, dependencies are installed from the Debian package management tool `apt`
- When verifying macOS build, dependencies are installed from the Homebrew package manager
- When verifying Windows build, dependencies are installed via vcpkg but no versions are specified
- When building pre-built binaries to release, dependencies are installed by scripts under `pkg/` directories. These scripts download the source code according to `dependencies.yaml` and build

It makes the pre-built binaries never tested except for the simple manual tests when voting for a new release. As a result, regression like https://github.com/apache/pulsar-client-cpp/issues/363 could happen.

This patch aims at integrating vcpkg to manage dependencies for all platforms, not only for Windows.

### Modifications

Integrate the CMakeLists.txt with vcpkg by:
1. Introduce vcpkg as a submodule of this project
2. Update `vcpkg.json` to specify dependency versions.
3. Set `CMAKE_TOOLCHAIN_FILE` with the `vcpkg.cmake` in the submodule.

Then, we can simply use `find_package` to find all dependencies and depend on them via a target like `CURL::libcurl` to have all include directores and link libraries.

Update the `unit-tests` workflow to verify building the binaries via vcpkg to test. The README is also updated to show how much simpler to build with vcpkg now.

### TODO

There are still some tasks that cannot be done by vcpkg:
1. The static library (e.g. `libpulsarwithdeps.a`) that bundles all 3rd-party dependencies.
2. The pre-built binaries are still built by scripts under `./pkg` directory. Specifically, vcpkg does not work with GCC 4.8 so on CentOS 7 we still need to build dependencies manually.

So the previous CMakeLists.txt are retained and will be used if `INTEGRATE_VCPKG` is OFF (by default). They will be removed until the task above can be done by vcpkg in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ Since it's integrated with vcpkg, see [vcpkg#README](https://github.com/microsof
 
 ```bash
 git clone https://github.com/apache/pulsar-client-cpp.git
-git submodule update --init --recursive
 cd pulsar-client-cpp
+git submodule update --init --recursive
 cmake -B build -DINTEGRATE_VCPKG=ON
 cmake --build build -j8
 ```
--- a/README.md
+++ b/README.md
@@ -31,37 +31,75 @@ For how to use APIs to publish and consume messages, see [examples](https://gith
 
 Pulsar C++ client uses [doxygen](https://www.doxygen.nl) to build API documents. After installing `doxygen`, you only need to run `doxygen` to generate the API documents whose main page is under the `doxygen/html/index.html` path.
 
-## Requirements
+## Build with vcpkg
 
-- A C++ compiler that supports C++11, like GCC >= 4.8
-- CMake >= 3.13
-- [Boost](http://www.boost.org/)
-- [Protocol Buffer](https://developers.google.com/protocol-buffers/) >= 3
-- [libcurl](https://curl.se/libcurl/)
-- [openssl](https://github.com/openssl/openssl)
+Since it's integrated with vcpkg, see [vcpkg#README](https://github.com/microsoft/vcpkg#readme) for the requirements. See [LEGACY_BUILD](./LEGACY_BUILD.md) if you want to manage dependencies by yourself or you cannot install vcpkg in your own environment.
 
-The default supported [compression types](include/pulsar/CompressionType.h) are:
-
-- `CompressionNone`
-- `CompressionLZ4`
-
-If you want to enable other compression types, you need to install:
-
-- `CompressionZLib`: [zlib](https://zlib.net/)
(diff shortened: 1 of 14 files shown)


Demogorgon314 (approved):
I tested it on M1 macOS, it works well. Great work!

shibd:
I will review it today.

README.md
```diff
@@ -31,239 +31,83 @@ For how to use APIs to publish and consume messages, see [examples](https://gith
...
-## Compilation
-
-### Clone
-
-First of all, clone the source code:
-
-```shell
-git clone https://github.com/apache/pulsar-client-cpp
+git clone https://github.com/apache/pulsar-client-cpp.git
+git submodule update --init --recursive
```
Demogorgon314, line 42: We should `cd pulsar-client-cpp` first

BewareMyPower: Fixed

README.md
```diff
@@ -31,239 +31,83 @@ For how to use APIs to publish and consume messages, see [examples](https://gith
...
-## Requirements
+## Build with vcpkg
 
-- A C++ compiler that supports C++11, like GCC >= 4.8
-- CMake >= 3.13
-- [Boost](http://www.boost.org/)
-- [Protocol Buffer](https://developers.google.com/protocol-buffers/) >= 3
-- [libcurl](https://curl.se/libcurl/)
-- [openssl](https://github.com/openssl/openssl)
+Since it's integrated with vcpkg, see [vcpkg#README](https://github.com/microsoft/vcpkg#readme) for the requirements. See [LEGACY_BUILD](./LEGACY_BUILD.md) if you want to manage dependencies by yourself or you cannot install vcpkg in your own environment.
```
shibd, line 37: We need to add a link to vcpkg.json to explain what version of is dependency

RobertIndie (approved):

README.md
```diff
@@ -31,239 +31,83 @@ For how to use APIs to publish and consume messages, see [examples](https://gith
...
-
-Tools will be placed in:
-
-```
-perf/perfProducer
-perf/perfConsumer
-```
-
-### Compile on Mac OS X
+The 1st step will download vcpkg and then install all dependencies according to the version description in [vcpkg.json](./vcpkg.json). The 2nd step will build the Pulsar C++ libraries under `./build/lib/`, where `./build` is the CMake build directory.
```
BewareMyPower, line 48: @shibd I added the link here. For those who manage dependencies by themselves, `LEGACY_BUILD.md` (the previous document) already documents it.

shibd (approved):

PR #377, opened 2023-12-26
[fix] Fix issue where custom logger setting is ignored

massakam:
### Motivation

I noticed that there are cases where the specified logger factory is not used. For example, if we compile and run the following code, it expects `FileLoggerFactory` to be used, but it actually uses the default `ConsoleLoggerFactory`.
```cpp
ClientConfiguration config = ClientConfiguration();
config.setLogger(new FileLoggerFactory(Logger::LEVEL_INFO, "pulsar-client-cpp.log"));

ParamMap params;
params["providerDomain"]    = "provider.foo.bar";
params["tenantDomain"]      = "tenant.foo.bar";
params["tenantService"]     = "test-service";
params["privateKey"]        = "file:///path/to/private.key";
params["keyId"]             = "0";
params["ztsUrl"]            = "https://zts.athenz.example.com:443";

AuthenticationPtr auth = AuthAthenz::create(params);
config.setAuth(auth);

Client client("pulsar://localhost:6650", config);
```

This happens if logging is performed before `LogUtils::setLoggerFactory` is executed in the constructor of `ClientImpl`.
https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/ClientImpl.cc#L94-L99

When logging is performed, `LogUtils::getLoggerFactory` is called.
1. https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/LogUtils.h#L60
1. https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/LogUtils.h#L43

If the constructor of `ClientImpl` has not been executed at this point, the default `ConsoleLoggerFactory` is instantiated, set to the static variable `s_loggerFactory`, and returned to the caller.
https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/LogUtils.cc#L38-L44

Loggers generated from the returned `LoggerFactory` will be set to the static and thread_local variable `threadSpecificLogPtr`, and reused from now on.
https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/LogUtils.h#L38-L47

After that, `LogUtils::setLoggerFactory` is executed again in the constructor of `ClientImpl`, but the value can only be set to `s_loggerFactory` once, and it will be ignored from the second time. This is clearly stated in [the comments](https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/include/pulsar/ClientConfiguration.h#L183-L186).
https://github.com/apache/pulsar-client-cpp/blob/25ea451eb3a79d48966689ec64460ae03d5d57da/lib/LogUtils.cc#L30-L36

Even if we can set it again, if a logger is already set to `threadSpecificLogPtr`, a new logger will not be generated from the new `LoggerFactory`.

### Modifications

Keep an instance of the default `ConsoleLoggerFactory` in `s_defaultLoggerFactory`, a variable separate from `s_loggerFactory`, and return it if `LoggerFactory` is not set yet.

In addition, keep the pointer value of the `LoggerFactory` that generated the logger cached in `threadSpecificLogPtr`, and regenerate the logger if the `LoggerFactory` pointer changes.

### Verifying this change

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/lib/LogUtils.cc
+++ b/lib/LogUtils.cc
@@ -25,6 +25,7 @@
 
 namespace pulsar {
 
+static std::atomic<LoggerFactory*> s_defaultLoggerFactory(new ConsoleLoggerFactory());
 static std::atomic<LoggerFactory*> s_loggerFactory(nullptr);
 
 void LogUtils::setLoggerFactory(std::unique_ptr<LoggerFactory> loggerFactory) {
@@ -37,10 +38,10 @@ void LogUtils::setLoggerFactory(std::unique_ptr<LoggerFactory> loggerFactory) {
 
 LoggerFactory* LogUtils::getLoggerFactory() {
     if (s_loggerFactory.load() == nullptr) {
-        std::unique_ptr<LoggerFactory> newFactory(new ConsoleLoggerFactory());
-        setLoggerFactory(std::move(newFactory));
+        return s_defaultLoggerFactory.load();
+    } else {
+        return s_loggerFactory.load();
     }
-    return s_loggerFactory.load();
 }
 
 std::string LogUtils::getLoggerName(const std::string& path) {
--- a/lib/ClientImpl.cc
+++ b/lib/ClientImpl.cc
@@ -92,11 +92,9 @@ ClientImpl::ClientImpl(const std::string& serviceUrl, const ClientConfiguration&
       consumerIdGenerator_(0),
       closingError(ResultOk) {
     std::unique_ptr<LoggerFactory> loggerFactory = clientConfiguration_.impl_->takeLogger();
-    if (!loggerFactory) {
-        // Use default simple console logger
-        loggerFactory.reset(new ConsoleLoggerFactory);
+    if (loggerFactory) {
+        LogUtils::setLoggerFactory(std::move(loggerFactory));
     }
-    LogUtils::setLoggerFactory(std::move(loggerFactory));
 
     LookupServicePtr underlyingLookupServicePtr;
     if (serviceNameResolver_.useHttp()) {
--- a/lib/LogUtils.h
+++ b/lib/LogUtils.h
@@ -34,16 +34,19 @@ namespace pulsar {
 #define PULSAR_UNLIKELY(expr) (expr)
 #endif
 
-#define DECLARE_LOG_OBJECT()                                                                     \
-    static pulsar::Logger* logger() {                                                            \
-        static thread_local std::unique_ptr<pulsar::Logger> threadSpecificLogPtr;                \
-        pulsar::Logger* ptr = threadSpecificLogPtr.get();                                        \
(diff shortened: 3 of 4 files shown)


lib/LogUtils.cc
```diff
@@ -25,6 +25,7 @@
 
 namespace pulsar {
 
+static std::atomic<LoggerFactory*> s_defaultLoggerFactory(new ConsoleLoggerFactory());
```
BewareMyPower, line 28: The `s_defaultLoggerFactory` is never released. It's not an issue if the application only run once, but if this library is dynamically loaded by an external system, it will be a memory leak.

Could a static logger factory be used instead?

```c++
static ConsoleLoggerFactory s_defaultLoggerFactory;
```

Then return `&s_defaultLoggerFactory` in `getLoggerFactory()`.

BewareMyPower (changes requested):

lib/LogUtils.cc
massakam: @BewareMyPower
`s_loggerFactory` also has the same issue, right? `s_loggerFactory` used to be a static shared_ptr, but it seems that it was changed to its current type to avoid segfaults during process shutdown.
https://github.com/apache/pulsar/pull/7132

massakam: Should I fix it like this?
```cpp
static std::atomic<LoggerFactory*> s_defaultLoggerFactory(nullptr);

LoggerFactory* LogUtils::getLoggerFactory() {
    if (s_loggerFactory.load() != nullptr) {
        return s_loggerFactory.load();
    } else {
        if (s_defaultLoggerFactory.load() == nullptr) {
            LoggerFactory* oldFactory = nullptr;
            LoggerFactory* newFactory = new ConsoleLoggerFactory();
            if (!s_defaultLoggerFactory.compare_exchange_strong(oldFactory, newFactory)) {
                delete newFactory; // there's already a factory set
            }
        }
        return s_defaultLoggerFactory.load();
    }
}
```

BewareMyPower: Oh I see. I think currently it's okay. Your fix here is also good.

BewareMyPower (approved):

lib/LogUtils.cc
BewareMyPower: Then this patch LGTM no matter if this fix is applied.

massakam: Thank you for your confirmation. Then, I will not make [the above modifications](https://github.com/apache/pulsar-client-cpp/pull/377#discussion_r1436381675).

Issue #357, opened 2023-11-29
[Bug] getSchemaInfoAsync is not honoring the PulsarClient timeout


armanjupriya-er:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

Pulsar Client CPP version: 3.3

### Minimal reproduce step

getSchemaInfoAsync is not honoring the PulsarClient timeout and is not timing out.

### What did you expect to see?

Currently, we are having getSchemaInfoAsync method which gives the schema version asynchronously. We need a method to get the schema version synchronously and which will honor the pulsar client timeout. 

### What did you see instead?

getSchemaInfoAsync is not honoring the PulsarClient timeout and is not timing out.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

BewareMyPower:
It needs some refactoring to add a test because the operation timeout is configured by seconds, not milliseconds. I'm going to work on this issue next week.

PR #383, opened 2024-01-08
Fix the operation timeout is not honored for GetSchema requests

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/357

### Motivation

In `Client::getSchemaInfoAsync`, the underlying
`LookupService::getSchema` is called, which does not register a timer according to the operation timeout.

### Modifications

Register a timer and cache it and the GetSchema promise in `ClientConnection::pendingGetSchemaRequests_`, then remove the entry and complete the promise with `ResultTimeout`.

To verify the timeout is honored, modify the timeout type from `int` to `chrono::nanoseconds` so that a smaller timeout can be configured. Then test a small enough timeout (1 ns) in
`LookupServiceTest.testGetSchemaTimeout`.


BewareMyPower:
There are other requests that don't honor the operation timeout as well. In future, we might refactor the logic to send any request with a timeout like:

```c++
sendRequest(request, timeout).thenApply(response -> handleXxx(response.getXxxResponse());
```

BewareMyPower:
The main branch is broken. I pushed a PR to fix: https://github.com/apache/pulsar-client-cpp/pull/384

lib/RetryableOperationCache.h
```diff
@@ -40,8 +41,8 @@ class RetryableOperationCache : public std::enable_shared_from_this<RetryableOpe
         explicit PassKey() {}
     };
 
-    RetryableOperationCache(ExecutorServiceProviderPtr executorProvider, int timeoutSeconds)
-        : executorProvider_(executorProvider), timeoutSeconds_(timeoutSeconds) {}
+    RetryableOperationCache(ExecutorServiceProviderPtr executorProvider, std::chrono::nanoseconds timeoutMs)
```
merlimat, line 44: is this `nanoseconds` or `milliseconds`?

BewareMyPower: Oh the name is wrong. It should be nanoseconds. I will address it

BewareMyPower: I removed all `Ms` or `Seconds` prefixes and use `TimeDuration` as the type now, PTAL again.

PR #387, opened 2024-01-18
Fix wrong TimeUtils::currentTimeMillis() implementation

BewareMyPower:
### Motivation

`TimeUtils::currentTimeMillis()` is used to get the timestamp since 1970 but `high_resolution_clock` does not necessarily represent the system clock. For example, given the following code:

```c++
template <typename Clock>
inline int64_t getTimestamp() {
    using namespace std::chrono;
    return duration_cast<milliseconds>(Clock::now().time_since_epoch()).count();
}

int main() {
    std::cout << time(nullptr) << std::endl;
    std::cout << getTimestamp<std::chrono::system_clock>() << std::endl;
    std::cout << getTimestamp<std::chrono::high_resolution_clock>() << std::endl;
}
```

The outputs could be:

```
1705584279
1705584279351
38566832
```

### Modifications

Use `std::system_clock` for `currentTimeMillis()`. Add `BackoffTest.testCurrentTimeMillis` to verify the result is similar to 1000 times of `time(nullptr)` (the timestamp in seconds from the C API).

### DIFF
--- a/lib/TimeUtils.h
+++ b/lib/TimeUtils.h
@@ -35,8 +35,9 @@ inline decltype(std::chrono::milliseconds(0).count()) toMillis(TimeDuration dura
 class PULSAR_PUBLIC TimeUtils {
    public:
     static ptime now() { return std::chrono::high_resolution_clock::now(); }
+
     static int64_t currentTimeMillis() {
-        return std::chrono::duration_cast<std::chrono::milliseconds>(now().time_since_epoch()).count();
+        return toMillis(std::chrono::system_clock::now().time_since_epoch());
     }
 };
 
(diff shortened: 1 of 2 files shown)


merlimat (approved):

BewareMyPower:
```
+ echo 'Building ZLib 1.2.12'
+ curl -O -L https://zlib.net/fossils/zlib-1.2.12.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 20429    0 20429    0     0   177k      0 --:--:-- --:--:-- --:--:--  178k
+ tar zxf zlib-1.2.12.tar.gz
tar: Error opening archive: Unrecognized archive format
```

The error is not related to this PR. So I merged it first. I will take a look at this issue later.

Issue #398, opened 2024-02-06
[Bug] Segmentation fault might happen during handshake if closed


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

main branch

### Minimal reproduce step

Apply the following diff:

```diff
diff --git a/lib/ClientConnection.cc b/lib/ClientConnection.cc
index 844d58f..4b11fba 100644
--- a/lib/ClientConnection.cc
+++ b/lib/ClientConnection.cc
@@ -388,6 +388,13 @@ typedef ASIO::detail::socket_option::integer<IPPROTO_TCP, TCP_KEEPIDLE> tcp_keep
  *  at this point the connection is deemed valid to be used by clients of this class
  */
 void ClientConnection::handleTcpConnected(const ASIO_ERROR& err, tcp::resolver::iterator endpointIterator) {
+    auto weakSelf = weak_from_this();
+    executor_->postWork([this, weakSelf] {
+        auto self = weakSelf.lock();
+        if (self) {
+            close(ResultDisconnected);
+        }
+    });
     if (!err) {
         std::stringstream cnxStringStream;
         try {
```

Then run the `AuthPluginTest.testTlsDetectPulsarSsl` test.

### What did you expect to see?

At least the test should not crash

### What did you see instead?

```
2024-02-06 17:16:24.187 INFO  [0x1fc1da100] ClientConnection:187 | [<none> -> pulsar+ssl://localhost:6651] Create ClientConnection, timeout=10000
2024-02-06 17:16:24.190 INFO  [0x1fc1da100] ConnectionPool:119 | Created connection for pulsar+ssl://localhost:6651-0
2024-02-06 17:16:24.191 INFO  [0x16efa7000] ClientConnection:410 | [[::1]:63587 -> [::1]:6651] Connected to broker
2024-02-06 17:16:24.191 INFO  [0x16efa7000] ClientConnection:1313 | [[::1]:63587 -> [::1]:6651] Connection disconnected (refCnt: 2)
2024-02-06 17:16:24.191 INFO  [0x16efa7000] ConnectionPool:134 | Remove connection for pulsar+ssl://localhost:6651-0
2024-02-06 17:16:24.191 INFO  [0x16efa7000] RetryableOperation:114 | Reschedule get-partition-metadata-persistent://private/auth/test-tls-detect for 100 ms, remaining time: 29900 ms
[...]

PR #399, opened 2024-02-06
Fix segmenatation fault if connection is closed during handshake

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/398

### Motivation

There is an implicit requirement for the `async_handshake` API that before it's done, the `socket` and `ssl::stream` objects must exist. Otherwise segmenatation fault might happen.

### Modifications

Capture `socket_` and `tlsSocket_` in the handshake callback. It can be verified by modifying the code according to #398 and rerun the test.


BewareMyPower:
Wait for https://github.com/apache/pulsar-client-cpp/pull/400 to be merged

merlimat (approved):

PR #400, opened 2024-02-06
Fix broken main branch after Pulsar image was upgraded to 3.2.0

BewareMyPower:
The `apachepulsar/pulsar:latest` image was upgraded to 3.2.0 recently, which results the following failed tests:
- ExtensibleLoadManagerTest: there are no more lookup requests after namespace unload
- ClientTest.testWrongListener: the lookup error message from broker does not contain the exception type


RobertIndie (approved):

heesung-sn (commented):
LGTM

PR #409, opened 2024-03-04
Fix hasMessageAvailable might return true after seeking to latest

BewareMyPower:
### Motivation

See https://github.com/apache/pulsar-client-python/issues/199

There is a race condition when `hasMessageAvailable` is called after `seek` if the start message ID of `Reader` is earliest.

In `ConsumerImpl::hasMessageAvailableAsync`, if the connection is not established at the moment, `lastDequedMessageId_` will be `earliest` because no message is received. Since `lastMessageIdInBroker_` is also `earliest`, `getLastMessageIdAsync` will be called and then it comes at

https://github.com/apache/pulsar-client-cpp/blob/e2cacb7dfb57b6d059b49fead2e1611548ff89b0/lib/ConsumerImpl.cc#L1554

However, before `getLastMessageIdAsync` is called, `messageId` was `earliest` because `lastDequedMessageId_` and `startMessageId_` were both `earliest`. However, when the callback is called, the `startMessageId_` has already been updated to `latest` in `connectionOpened`, so we should compare to `latest`.

### Modifications

In the callback of `getLastMessageIdAsync`, retrieve the latest value of `startMessageId_` to compare rather then reusing the old value.

Refactor the seek flow to reset the seek states and trigger the callback after updating the `startMessageId_`.

`ReaderTest.testHasMessageAvailableAfterSeekToEnd` is added to cover the changes.

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -1685,10 +1685,10 @@ void ConsumerImpl::seekAsyncInternal(long requestId, SharedBuffer seek, const Me
                 LOG_INFO(getName() << "Seek successfully");
                 ackGroupingTrackerPtr_->flushAndClean();
                 incomingMessages_.clear();
-                Lock lock{mutexForMessageId_};
+                Lock lock(mutexForMessageId_);
                 lastDequedMessageId_ = MessageId::earliest();
                 lock.unlock();
-                if (getCnx().expired()) {
+                if (!hasParent_ && getCnx().expired()) {
                     // It's during reconnection, complete the seek future after connection is established
                     seekStatus_ = SeekStatus::COMPLETED;
                 } else {
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -1502,10 +1502,14 @@ void ConsumerImpl::seekAsync(uint64_t timestamp, ResultCallback callback) {
 bool ConsumerImpl::isReadCompacted() { return readCompacted_; }
 
 void ConsumerImpl::hasMessageAvailableAsync(HasMessageAvailableCallback callback) {
-    Lock lock(mutexForMessageId_);
-    if (lastDequedMessageId_ == MessageId::earliest() &&
-        startMessageId_.get().value_or(MessageId::earliest()) == MessageId::latest()) {
-        lock.unlock();
+    bool compareMarkDeletePosition;
+    {
+        std::lock_guard<std::mutex> lock{mutexForMessageId_};
+        compareMarkDeletePosition =
+            (lastDequedMessageId_ == MessageId::earliest()) &&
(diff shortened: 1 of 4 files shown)


BewareMyPower:
@merlimat @shibd @RobertIndie @Demogorgon314 Could you take a look?

tests/ReaderTest.cc
```diff
@@ -784,4 +784,56 @@ TEST(ReaderSeekTest, testStartAtLatestMessageId) {
...
+    Producer producer;
+    ASSERT_EQ(ResultOk, client.createProducer(topic, producer));
+    Reader reader;
+    ASSERT_EQ(ResultOk, client.createReader(topic, MessageId::earliest(), {}, reader));
+
+    producer.send(MessageBuilder().setContent("msg-0").build());
+    producer.send(MessageBuilder().setContent("msg-1").build());
+
+    bool hasMessageAvailable;
+    if (GetParam()) {
```
shibd, line 814:
```suggestion
    if (isMultiTopic_) {
```

tests/ReaderTest.cc
```diff
@@ -784,4 +784,56 @@ TEST(ReaderSeekTest, testStartAtLatestMessageId) {
...
+        // Test the case when `ConsumerImpl.lastMessageIdInBroker_` has been initialized
+        reader.hasMessageAvailable(hasMessageAvailable);
```
shibd, line 816: Do we need to add assert here?

lib/ConsumerImpl.cc
```diff
@@ -236,16 +236,15 @@ Future<Result, bool> ConsumerImpl::connectionOpened(const ClientConnectionPtr& c
     // sending the subscribe request.
     cnx->registerConsumer(consumerId_, get_shared_this_ptr());
 
-    if (duringSeek_) {
+    if (duringSeek()) {
         ackGroupingTrackerPtr_->flushAndClean();
     }
 
     Lock lockForMessageId(mutexForMessageId_);
     // Update startMessageId so that we can discard messages after delivery restarts
```
RobertIndie, line 244: We can move this comment inside the `clearReceiveQueue`.

tests/ReaderTest.cc
BewareMyPower: Emmm. Actually I use a different meaning for this bool param because if `hasMessageAvailable` has been called before `seek`, the `lastmessageIdInBroker_` will be initialized and the logic of the following `has MessageAvailable` will be different.

BewareMyPower: I moved it to another test

BewareMyPower:
@RobertIndie @shibd Comments are addressed, PTAL again.

Demogorgon314 (approved):

RobertIndie (approved):

shibd (approved):

PR #427, opened 2024-06-11
Upgrade vcpkg dependencies for the missed boost-format dependency

BewareMyPower:
### Motivation

See https://github.com/microsoft/vcpkg/pull/36510

Currently the commit id of the vcpkg submodule is old. With latest dependencies in vcpkg, it requires a `boost-format` dependency to build.

### Modifications

Upgrade the vcpkg submodule to the latest commit at the moment and add the `boost-format` dependency to fix the build.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -18,11 +18,11 @@
 #
 
 # Note: GCC 4.8 is incompatible with Boost >= 1.84 for the missed std::align
-boost: 1.83.0
+boost: 1.85.0
 cmake: 3.28.3
 protobuf: 3.20.0
 zlib: 1.3.1
-zstd: 1.5.5
+zstd: 1.5.6
 snappy: 1.1.10
 openssl: 1.1.1w
-curl: 8.6.0
+curl: 8.8.0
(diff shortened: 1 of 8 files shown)


merlimat (approved):

BewareMyPower:
There are some failed tests. I'm going to check if they were introduced by an upgraded dependency

BewareMyPower:
It's an issue with the JSON serialization for key-value schema. Maybe caused by the `boost.property_tree` dependency.

BewareMyPower:
The root cause is that latest's Boost.PropertyTree does not write a line ending in `write_json`. So I updated the Boost version to 1.85 in `dependencies.yaml` as well to avoid behavior inconsistency.

BewareMyPower:
There is a bug for GCC 4.8 that it does not support [`std::align`](https://en.cppreference.com/w/cpp/memory/align), which is a standard of C++11.

I sent an email to drop GCC 4.8 in the next release (see https://lists.apache.org/thread/7o8hpv1gtoffvzx053wm0ss2s9xt0795), so I will open a PR to drop the support for GCC 4.8 first and then go back to this PR.

PR #432, opened 2024-07-03
Add -Bsymbolic link option to avoid symbol interposition

BewareMyPower:
### Motivation

There is a case that `libpulsar.so` could unexpectedly call functions from other dependencies.

For example, assuming the application depends on two libraries:
- `libpulsar.so`, which includes the symbols from `libcurl.a` 8.4.0
- `libfoo.so`, which includes the symbols from `libcurl.a` 7.82.0

If the link order is `libfoo.so` first, then the libcurl definitions from 7.82.0 will also be used by `libpulsar.so` and then the application might crash due to the incompatibility. This is an issue specifically with Linux ELF format.

### Modifications

Add the `-Wl,-Bsymbolic` link option for GCC.


BewareMyPower:
You can also see https://github.com/BewareMyPower/symbol-interposition for the symbol interposition issue.

merlimat (approved):

shibd (approved):

Demogorgon314 (approved):

PR #441, opened 2024-08-21
Fix invalid memory access on the first pending batch receive callback

BewareMyPower:
### Motivation

https://github.com/apache/pulsar-client-cpp/blob/5940cb518bdf3a13ae7f859fab10d6940eec51b7/lib/ConsumerImplBase.cc#L109-L112

In the code above, `batchReceive` is a reference to the first element in the queue (`batchPendingReceives_`). After `pop()`, the memory would be invalid to access, which might cause unexpected crash.

### Modifications

Add a  `popBatchReceiveCallback` method that moves the callback from the first element in `batchPendingReceives_`. Use this method to replace error-prone `front()` - `pop()` calls.

### DIFF
--- a/lib/ConsumerImplBase.cc
+++ b/lib/ConsumerImplBase.cc
@@ -76,10 +76,7 @@ void ConsumerImplBase::doBatchReceiveTimeTask() {
         long diff =
             batchReceivePolicy_.getTimeoutMs() - (TimeUtils::currentTimeMillis() - batchReceive.createAt_);
         if (diff <= 0) {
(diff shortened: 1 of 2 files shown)


BewareMyPower:
This is an example crash frame in production:

```
frame #3: 0x0000ffff6efa846c pulsar.node`pulsar::ConsumerImplBase::notifyBatchPendingReceivedCallback(this=0x0000ffff54009a60) at ConsumerImplBase.cc:112:43
```

lhotari (approved):
LGTM. Good work @BewareMyPower

PR #450, opened 2024-10-22
Fix default operation timeout

arhoads:
Fixes: Default operation timeout has the wrong value (no issue entered)

### Motivation
The driver will timeout on all operations because the operation timeout is initialized with a literal of a too small size and when cast to `std::chrono::seconds` we end up with 0. `long long` should be used because according to https://en.cppreference.com/w/cpp/chrono/duration `std::chrono::nanoseconds` takes a type of at least 64 bits. 

Example of the problem here:
https://godbolt.org/z/o51G6xbqo

### Verifying this change

*(example:)*

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/ClientConfigurationImpl.h
+++ b/lib/ClientConfigurationImpl.h
@@ -30,7 +30,7 @@ struct ClientConfigurationImpl {
     uint64_t memoryLimit{0ull};
     int ioThreads{1};
     int connectionsPerBroker{1};
-    std::chrono::nanoseconds operationTimeout{30L * 1000 * 1000 * 1000};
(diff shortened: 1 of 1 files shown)


BewareMyPower (approved):

BewareMyPower:
There seems to be something wrong with the CI, I will take a look next week

BewareMyPower:
@arhoads Could you rebase to the latest main branch, which has fixed the CI?

RobertIndie (approved):

shibd (approved):

Issue #461, opened 2024-12-09
[Bug] startMessageInclusive does not work for reader with chunk messages


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

main branch (e264cf7242c8f85e7f8d66335218ff38c658c8ad)

### Minimal reproduce step

Add the following test to `tests/ReaderTest.cc`:

```c++
TEST_F(ReaderSeekTest, testSeekInclusiveChunkMessage) {
    const auto topic = "test-seek-inclusive-chunk-message-" + std::to_string(time(nullptr));

    Producer producer;
    ProducerConfiguration producerConf;
    producerConf.setBatchingEnabled(false);
    producerConf.setChunkingEnabled(true);
    ASSERT_EQ(ResultOk, client.createProducer(topic, producerConf, producer));

    std::string largeValue(1024 * 1024 * 6, 'a');
    MessageId msgId;
    ASSERT_EQ(ResultOk, producer.send(MessageBuilder().setContent(largeValue).build(), msgId));
    ASSERT_EQ(ResultOk, producer.send(MessageBuilder().setContent(largeValue).build()));

    Reader reader;
    ReaderConfiguration readerConf;
    readerConf.setStartMessageIdInclusive(true);
    ASSERT_EQ(ResultOk, client.createReader(topic, msgId, readerConf, reader));

    Message msg;
    ASSERT_EQ(ResultOk, reader.readNext(msg, 3000));
    ASSERT_EQ(msgId, msg.getMessageId());
}
```

and run this test

### What did you expect to see?

The test succeeds.

### What did you see instead?

The test failed

```
pulsar-client-cpp/tests/ReaderTest.cc:912: Failure
Expected equality of these values:
  msgId
    Which is: (24,0,-1,-1);(24,1,-1,-1)
  msg.getMessageId()
    Which is: (24,2,-1,-1);(24,3,-1,-1)
```

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #462, opened 2024-12-11
Fix startMessageInclusive does not work if the 1st message is a chunked message

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/461


BewareMyPower:
It seems that the macOS build workflows are somehow broken.

shibd (approved):

RobertIndie (approved):

Issue #464, opened 2024-12-16
[Bug] Cannot build macOS x86_64 libraries


BewareMyPower:
### Search before asking

- [X] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

main branch (aeff955f2a5e3d87532b606bab02f4067b795bd5)

### Minimal reproduce step

See https://github.com/apache/pulsar-client-cpp/actions/runs/12339133262/job/34434971965

### What did you expect to see?

It passes.

### What did you see instead?

```
Packing libsnappy -- '/Users/runner/work/pulsar-client-cpp/pulsar-client-cpp/pkg/mac/.build/install/lib/libsnappy.a'
ar: /Users/runner/work/pulsar-client-cpp/pulsar-client-cpp/pkg/mac/.build/install/lib/libsnappy.a is a fat file (use libtool(1) or lipo(1) and ar(1) on it)
ar: /Users/runner/work/pulsar-client-cpp/pulsar-client-cpp/pkg/mac/.build/install/lib/libsnappy.a: Inappropriate file type or format
make[2]: *** [lib/CMakeFiles/pulsarStaticWithDeps] Error 1
make[1]: *** [lib/CMakeFiles/pulsarStaticWithDeps.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #465, opened 2024-12-16
Use vcpkg to build macOS packages

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/464

The manual build of 3rd party libraries are error-prone. It seems that on x86_64 platform `libsnappy.a` is a universal binary that cannot be operated with `ar`. This PR switches to vcpkg for a reliable build for dependencies.

It should be noted that vcpkg enables IPv6 by default when building libcurl, which makes the `libcurl.a` require the dynamic linking to some macOS frameworks. e.g. you need to add `-framework SystemConfiguration -framework CoreFoundation` link option, and even `-framework CoreServices` option for macOS 14 (see https://github.com/curl/curl/issues/11893). This is not user friendly for the downstream libraries like pulsar-client-node.

Therefore, this PR builds macOS libraries via patching vcpkg to disable the IPv6 feature of libcurl for macOS. The upstream does not accept this change, see https://github.com/microsoft/vcpkg/issues/35834.

Two customized triplets files are added:
- Only build release version dependencies
- Set the minimum target OS version to 13.0


pkg/mac/build-static-library.sh
```diff
@@ -18,182 +18,48 @@
...
-fi
-
-popd # pkg/mac
-cd ../../ # project root
-
-cmake -B build-static -DCMAKE_OSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET \
-    -DLINK_STATIC=ON \
+# Use a forked version of vcpkg to support building libcurl with IPv6 disabled
+rm -rf vcpkg
+git clone https://github.com/BewareMyPower/vcpkg.git -b curl-8.4.0-osx-patch
```
RobertIndie, line 39: How about using a patch file for the commit https://github.com/BewareMyPower/vcpkg/commit/b03173a5d0cbd2f7c3c5e85a576e98fba3560835 so that we don't need to introduce the personal repo?

BewareMyPower: Okay, I think we can restore the solution in https://github.com/apache/pulsar-client-cpp/pull/465/commits/382c9d9a5fcc3212987685d416b06983e9af82c6

BewareMyPower:
@RobertIndie I've updated the approach with patching, PTAL again. In addition, I removed the macOS build workflow with homebrew dependencies because the latest Boost is no longer compatible with the code base, see https://github.com/apache/pulsar-client-cpp/actions/runs/12498761411/job/34872997202

RobertIndie (approved):

PR #467, opened 2024-12-18
[fix] Fix issue where cert chain is not taken into account in mTLS authentication

massakam:
### Motivation

The C++ client has the same mTLS authentication issue as https://github.com/apache/pulsar/pull/23644. That is, if a client's certificate is not directly signed by a root CA, but is signed by an intermediate CA, mTLS authentication will fail.

### Modifications

- When loading a client cert, use [ssl::context::use_certificate_chain_file](https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/reference/ssl__context/use_certificate_chain_file.html) instead of [ssl::context::use_certificate_file](https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/reference/ssl__context/use_certificate_file.html).
- Added a test to check whether a PEM file containing a client cert and intermediate CA cert can be loaded and successfully connected to a broker. When issuing the intermediate CA cert, I did not know the private key of the root CA, so I reissued the root CA cert.

### Verifying this change

### Documentation

- [x] `doc-not-needed`


hrsakai (approved):
LGTM

BewareMyPower:
3.7.0 is currently blocked at the macOS release process, so I think it can be included in 3.7.0

PR #473, opened 2025-02-06
[fix] Fix an issue where zero queue consumers are unable to receive messages after topic unloading

massakam:
### Motivation

If a consumer with a receiver queue size of 0 is waiting for messages using `receive` or `receiveAsync`, and the topic is unloaded or the broker is restarted, the consumer will no longer be able to receive messages. This is because `receive` and `receiveAsync` have different bugs. Both bugs are highly reproducible.

#### receive

When the `receive` method of a zero queue consumer is executed, `fetchSingleMessageFromBroker` is called internally. In `fetchSingleMessageFromBroker`, the connection of the received message is compared with the connection when the flow permit was sent, and if they do not match, the message is discarded.
https://github.com/apache/pulsar-client-cpp/blob/115d64af81ae26ae73aaba7aaba36d4a9e47324c/lib/ConsumerImpl.cc#L918-L942

If the topic is unloaded or the broker is restarted between sending the flow permit and receiving the message, the connections at the two times will no longer match, so the message will be discarded and no further flow permits will be sent, causing message delivery to stop.

#### receiveAsync

If the consumer is reopened by unloading the topic or restarting the broker, flow permits that were sent before the reopening must be sent again. However, if a consumer with a receiver queue size of 0 is waiting for messages with `receiveAsync`, it does not seem to send flow permits even if the consumer is reopened.
https://github.com/apache/pulsar-client-cpp/blob/115d64af81ae26ae73aaba7aaba36d4a9e47324c/lib/ConsumerImpl.cc#L311-L332

#### receive

In `fetchSingleMessageFromBroker`, I think the connection of the received message should be compared with the current connection, not the connection at the time the flow permit was sent. In fact, the Java client makes such a comparison.
https://github.com/apache/pulsar/blob/8a40b30cf47a91ec02d931e6371d02409ba5951e/pulsar-client/src/main/java/org/apache/pulsar/client/impl/ZeroQueueConsumerImpl.java#L121

In addition, if the `handleCreateConsumer` method, which is executed when the consumer is reopened, is executed "after setting `waitingForZeroQueueSizeMessage` to true" and "before sending a flow permit", there is a possibility that the flow permit will be sent twice, so we lock `mutex_` to perform exclusive control.

#### receiveAsync

We can find the number of flow permits sent before the reconnection and the number of callbacks waiting for `receiveAsync` to complete by checking the number of elements in `pendingReceives_`.
https://github.com/apache/pulsar-client-cpp/blob/115d64af81ae26ae73aaba7aaba36d4a9e47324c/lib/ConsumerImpl.cc#L976-L981

Therefore, when a zero queue consumer is reopened, it should send the same number of flow permits as the elements of `pendingReceives_`.

In this case too, if `handleCreateConsumer` is executed "after adding an element to `pendingReceives_`" and "before sending a flow permit", it is possible that more flow permits will be sent than expected. So we need to lock `mutex_` in `receiveAsync` too.

### Verifying this change

### Documentation

- [x] `doc-not-needed`

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -310,15 +310,23 @@ Result ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result
     if (result == ResultOk) {
         LOG_INFO(getName() << "Created consumer on broker " << cnx->cnxString());
         {
-            Lock lock(mutex_);
+            Lock mutexLock(mutex_);
             setCnx(cnx);
             incomingMessages_.clear();
             possibleSendToDeadLetterTopicMessages_.clear();
             state_ = Ready;
             backoff_.reset();
-            // Complicated logic since we don't have a isLocked() function for mutex
-            if (waitingForZeroQueueSizeMessage) {
-                sendFlowPermitsToBroker(cnx, 1);
+            if (!messageListener_ && config_.getReceiverQueueSize() == 0) {
+                // Complicated logic since we don't have a isLocked() function for mutex
+                if (waitingForZeroQueueSizeMessage) {
+                    sendFlowPermitsToBroker(cnx, 1);
+                }
+                // Note that the order of lock acquisition must be mutex_ -> pendingReceiveMutex_,
+                // otherwise a deadlock will occur.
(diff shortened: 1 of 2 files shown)


BewareMyPower (approved):
Great fix!

PR #447, opened 2024-10-02
Fix ack failure on message listener in multi topics consumer

nkurihar:
Fixes #446 

### Motivation

We found an ack failure issue: #446 

MessageListeners can start to process messages before subscribing all child topics are completed in "multi topics" (e.g. partitioned, list, regex) cases.
If `acknowledge()` is called in messageListeners (it is very typical usage) before subscribing completion, it will fail with `AlreadyClosed` error since the state of the parent consumer is not `Ready` yet:
https://github.com/apache/pulsar-client-cpp/blob/8b2753a56579ea6cf11e26a0c5a160797518df63/lib/MultiTopicsConsumerImpl.cc#L643-L648
That results in ack holes, and finally full backlog.

### Modifications

* `MultiTopicsConsumerImpl`: Pause messageListeners at first by setting `startPaused` to true and resume them after subscribing all child topics are completed.
* `ConsumerImpl`: Delete unused code.
  * It tries to skip `sendFlowPermitsToBroker()` at the first connection.
    * The motivation seems same as this PR, i.e. to prevent messageListeners from starting before subscribing completion.
  * However, actually it does not work so far, the variable `firstTime` looks always false at https://github.com/apache/pulsar-client-cpp/blob/8b2753a56579ea6cf11e26a0c5a160797518df63/lib/ConsumerImpl.cc#L316 because it becomes false at https://github.com/apache/pulsar-client-cpp/blob/8b2753a56579ea6cf11e26a0c5a160797518df63/lib/ConsumerImpl.cc#L297-L299
    * Also it is static variable in a method that is shared by all `ConsumerImpl` instances, whereas it has no chance of returning to true once it becomes false.
  * It seems better that child consumers don't have to care about their parents and are completely independent of them.

### Verifying this change

- [ ] Make sure that the change passes the CI checks.

Specific tests for this issue is difficult to implement because it does not occur every time.

### Documentation

- [ ] `doc-required` 
(Your PR needs to update docs and you will update later)

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 
(Your PR contains doc changes)

- [ ] `doc-complete`
(Docs have been already added)

PR #481, opened 2025-04-10
Fix multi-topics-consumer new topic listeners stuck in paused state

oversearch:
This was introduced in v3.7.0 via #447, which contained a change that attempted to fix an issue where a pre-mature ack of a message before a multi-topic subscriber was ready could have caused a crash.  To fix the original bug, all multi-topic subscriptions are started with their message listener paused. They later get un-paused once all topics are successfully subscribed and connected.  However, on a regex subscription when new topics are discovered, they also start in a paused state, and there's no mechanism to resume them.  Hence, they get stuck, and no messages for new topics will be processed.

This change adds a call to resume any new listeners after new topics are added.  The fix has been deployed to production in my company's infrastructure and has been confirmed working for a few weeks.

### Verifying this change

I could not get the tests to pass with a bare checkout of master no matter what I did.  The first few tests were extremely flaky, so I commented them out.  The OAuth tests require sudo and wanted to modify my root certificates, which I'm not about to allow, so I commented that out.  And the broker tests always segfault in the `BasicEndToEndTest.testUnAckedMessageTrackerEnabledIndividualAck` test.  I gave up trying to mess with the tests - I've confirmed the fix works in my environment and our tests pass, so I was satisfied.  Just trying to upstream this bugfix... 

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
Simple regression bug fix

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/PatternMultiTopicsConsumerImpl.cc
+++ b/lib/PatternMultiTopicsConsumerImpl.cc
@@ -118,6 +118,9 @@ void PatternMultiTopicsConsumerImpl::timerGetTopicsOfNamespace(const Result resu
     // callback method when added topics all subscribed.
     ResultCallback topicsAddedCallback = [this, topicsRemoved, topicsRemovedCallback](Result result) {
         if (result == ResultOk) {
+            if(messageListener_) {
+                resumeMessageListener();
+            }
             // call to unsubscribe all removed topics.
             onTopicsRemoved(topicsRemoved, topicsRemovedCallback);
         } else {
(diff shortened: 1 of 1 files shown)


BewareMyPower:
Sorry, did you mention a wrong PR? #477 is drafted and not included in 3.7.0

lib/PatternMultiTopicsConsumerImpl.cc
```diff
@@ -118,6 +118,9 @@ void PatternMultiTopicsConsumerImpl::timerGetTopicsOfNamespace(const Result resu
     // callback method when added topics all subscribed.
     ResultCallback topicsAddedCallback = [this, topicsRemoved, topicsRemovedCallback](Result result) {
         if (result == ResultOk) {
+            if(messageListener_) {
```
BewareMyPower, line 121:
```suggestion
            if (messageListener_) {
```

Fix the coding style

BewareMyPower:
I see it should be a regression caused by https://github.com/apache/pulsar-client-cpp/pull/447.

BewareMyPower (approved):

oversearch:
> I see it should be a regression caused by #447.

Ah yes, my apologies, I typo'd that PR number by mistake.  Thank you for merging this fix!

PR #487, opened 2025-05-02
Fix TableView's existing key-value will never be updated

### LINKED ISSUE #486: [Bug] TableView's existing value will never update
### Search before asking

### Version

TableView uses `SynchronizedHashMap#emplace` to update a key-value:

https://github.com/apache/pulsar-client-cpp/blob/0e1ed3b2af24f5373036b839e4131257e8e75bf2/lib/TableViewImpl.cc#L107

https://github.com/apache/pulsar-client-cpp/blob/0e1ed3b2af24f5373036b839e4131257e8e75bf2/lib/SynchronizedHashMap.h#L58

However, the underlying `emplace` method does not update an existing key-value.

### Minimal reproduce step

TBD

TBD

### What did you see instead?

TBD

### Anything else?

_No response_

BewareMyPower:
Fixes #486

### Motivation

The `emplace` method of `SynchronizedHashMap` adopts the same semantics as https://en.cppreference.com/w/cpp/container/map/emplace, which inserts the key-value only when the key does not exist. `TableViewImpl` leverages this method to update the key-values, so it will never succeed. In addition, this method is not thread safe because if the key exists, an iterator will be returned to the caller, then the caller could use the iterator to modify the value without any synchronization.

### Modifications

Add two new methods `putIfAbsent` and `put` and remove the `emplace` method.
- `putIfAbsent`: inserts the key-value only when the key does not exist and return the optional existing value
- `put`: inserts the key-value no matter if the key exists.

Then, replace the `emplace` call with `put` in `TableViewImpl`.

### Verifying this change

`testSimpleTableView` is improved to verify the update case.

### DIFF
--- a/lib/ClientImpl.cc
+++ b/lib/ClientImpl.cc
@@ -225,8 +225,9 @@ void ClientImpl::handleProducerCreated(Result result, ProducerImplBaseWeakPtr pr
         auto address = producer.get();
         auto existingProducer = producers_.create(address, producer);
         if (existingProducer) {
+            auto producer = existingProducer.value().lock();
             LOG_ERROR("Unexpected existing producer at the same address: "
-                      << address << ", producer: " << existingProducer.value().lock()->getProducerName());
+                      << address << ", producer: " << (producer ? producer->getProducerName() : "(null)"));
             callback(ResultUnknownError, {});
             return;
         }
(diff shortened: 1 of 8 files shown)


lib/ClientImpl.cc
```diff
@@ -222,12 +222,12 @@ void ClientImpl::handleCreateProducer(const Result result, const LookupDataResul
 void ClientImpl::handleProducerCreated(Result result, ProducerImplBaseWeakPtr producerBaseWeakPtr,
                                        CreateProducerCallback callback, ProducerImplBasePtr producer) {
     if (result == ResultOk) {
-        auto pair = producers_.emplace(producer.get(), producer);
-        if (!pair.second) {
-            auto existingProducer = pair.first->second.lock();
+        auto address = producer.get();
+        auto existingProducer = producers_.create(address, producer);
```
merlimat, line 226: `create()` is a bit confusing here as a name. I was wondering why do we need to "create a new consumer here"... 

Could we rename the method as `putIfAbsent()`?

merlimat (approved):

lib/ClientImpl.cc
BewareMyPower: Yes, it makes sense.

BewareMyPower: I renamed these methods as `put` and `putIfAbsent` now.

Issue #200, opened 2023-02-20
[Improve] Apply clang-tidy check in CI


RobertIndie:
It's better to apply the `clang-tidy` check in CI to discover some problems in the code.

BewareMyPower:
Could you assign this issue to me? The existing code has many problems so that `clang-tidy` checks cannot pass. And accidentally I found a bug (https://github.com/apache/pulsar-client-cpp/issues/202) from the `clang-tidy` check.

PR #490, opened 2025-05-20
[CI] Add clang-tidy check

BewareMyPower:
Fixes https://github.com/apache/pulsar-client-cpp/issues/200

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,67 @@
+---
+Checks: >
+  # It follows https://github.com/googleapis/google-cloud-cpp/blob/main/.clang-tidy
+  -*,
+  abseil-*,
+  bugprone-*,
+  google-*,
+  misc-*,
+  modernize-*,
+  performance-*,
+  portability-*,
+  readability-*,
+  -google-readability-braces-around-statements,
+  -google-readability-namespace-comments,
+  -google-runtime-references,
+  -misc-non-private-member-variables-in-classes,
+  -misc-const-correctness,
+  -misc-include-cleaner,
+  -modernize-return-braced-init-list,
+  -modernize-use-trailing-return-type,
+  -modernize-concat-nested-namespaces,
+  -modernize-use-nodiscard,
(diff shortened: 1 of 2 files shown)


lib/ClientImpl.cc
```diff
@@ -735,23 +739,14 @@ void ClientImpl::shutdown() {
...
-    for (auto&& kv : consumers) {
-        ConsumerImplBasePtr consumer = kv.second.lock();
-        if (consumer) {
-            consumer->shutdown();
-        }
-    }
+    auto numProducers = producers.size();
+    auto numConsumers = consumers.size();
+    // Call the destructors out of the lock to avoid deadlocks
+    producers.clear();
```
RobertIndie, line 745: The producer in the map `producers` is a weak ptr. Calling the `clear` here doesn't guarantee the desturctor will be called. So the shutdown may not be called here. However, I think we should make sure all the producers/consumers should be shutdown after the client shutdown.

build-support/run_clang_tidy.sh
```diff
@@ -0,0 +1,40 @@
...
+done
+# run-clang-tidy from older version of LLVM requires python but not python3 as the env, so we cannot run it directly
+SCRIPT=$(which run-clang-tidy)
+set +e
+nproc
+if [[ $? == 0 ]]; then
+    python3 $SCRIPT -p build -j$(nproc) $(cat files.txt)
+else
+    python3 $SCRIPT -p build -j8 $(cat files.txt)
+fi
```
RobertIndie, line 40: Should we delete the `files.txt` here?

lib/ClientImpl.cc
BewareMyPower: Good find! It makes sense. Let me think of a better solution.

BewareMyPower: @RobertIndie I added `shutdown` back as the virtual function and a new non-virtual function `internalShutdown()` is added instead. Could you take a look again?

PR #499, opened 2025-08-20
fix: remove unnecessary copies

gy-deng:
Remove unnecessary copies from PR https://github.com/apache/pulsar-client-cpp/pull/497

[lib/NegativeAcksTracker.cc](https://github.com/apache/pulsar-client-cpp/pull/497/files/fe435c918874b0c1447c59ab8a5511b5ea36c604#diff-eaa00d87a3e2e87777561ceb10d136ad1d227957e72c763fdabc1a7ec1572e15)

        auto ledgerMap = it->second;
        for (auto ledgerIt = ledgerMap.begin(); ledgerIt != ledgerMap.end(); ++ledgerIt) {
            auto entrySet = ledgerIt->second;

>This makes a copy. Is this intended to be a reference instead - `auto& entrySet = ledgerIt->second;`?
This line also makes a copy `auto ledgerMap = it->second;
_Originally posted by @arhoads in https://github.com/apache/pulsar-client-cpp/pull/497#discussion_r2274220029_

### Verifying this change

This change is already covered by existing tests.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/NegativeAcksTracker.cc
+++ b/lib/NegativeAcksTracker.cc
@@ -88,9 +88,9 @@ void NegativeAcksTracker::handleTimer(const ASIO_ERROR &ec) {
             break;
         }
 
-        auto ledgerMap = it->second;
+        auto &ledgerMap = it->second;
         for (auto ledgerIt = ledgerMap.begin(); ledgerIt != ledgerMap.end(); ++ledgerIt) {
-            auto entrySet = ledgerIt->second;
+            auto &entrySet = ledgerIt->second;
             for (auto setIt = entrySet.begin(); setIt != entrySet.end(); ++setIt) {
                 messagesToRedeliver.insert(
                     MessageIdBuilder().ledgerId(ledgerIt->first).entryId(*setIt).build());


lib/NegativeAcksTracker.cc
```diff
@@ -88,9 +88,9 @@ void NegativeAcksTracker::handleTimer(const ASIO_ERROR &ec) {
             break;
         }
 
-        auto ledgerMap = it->second;
+        auto &ledgerMap = it->second;
         for (auto ledgerIt = ledgerMap.begin(); ledgerIt != ledgerMap.end(); ++ledgerIt) {
-            auto entrySet = ledgerIt->second;
+            auto &entrySet = ledgerIt->second;
```
BewareMyPower, line 93: To avoid value copy, it's better to write `const auto& x = expression` whenever possible, i.e. except for the cases that:
- You want to get a mutable reference to modify the value that `expression` refers.
- You want to perform the value copy explicitly, which usually happens in the case that an immutable is passed to a different thread (typically use cases of `shared_ptr`)

PR #529, opened 2025-12-10
[improve][client] Add TLSv1.3 support

### LINKED ISSUE #172: Support for TLS 1.3 in Pulsar CPP Client
## Support for TLS 1.3 in Pulsar CPP Client

Pulsar CPP client supports only TLS 1.2 for communication.
Reference: https://github.com/apache/pulsar-client-cpp/blob/a46da16421323d146c33d1e9d00a0352b50ace89/lib/ClientConnection.cc#L193

Pulsar CPP client needs to support TLS 1.3 in addition to TLS 1.2 for better security.

ciuncan:
Fixes #172 

### Motivation

We've met this issue with a customer, whose network does deep packet inspection and somehow prevents TLSv1.2 connections to successfully do handshake. Other clients such as Java/Rust works just fine as they are able to connect with TLSv1.3. Since TLSv1.2 usage is hardcoded in CPP client (and by extension Python client), pulsar connection was no longer possible from there. I've opened a PR and would be happy to get it evaluated to be incorporated in `pulsar-client` (we use python package).

### Modifications

The asio SSL context is now constructed with `ASIO::ssl::context::sslv23_client` instead of `ASIO::ssl::context::tlsv12_client`, and with `default_workarounds | no_sslv2 | no_sslv3 |no_tlsv1 | no_tlsv1_1` options, which excludes unsafe versions and leaves only TLSv1.2 and TLSv1.3. I've tested this connection on the network and it worked fine.

### Verifying this change

  - Added TLS handshake tests (`TlsNegotiationTest.cc`) that will build a mock server that will either accept only TLSv1.2 and TLSv1.3
  - The updated client TLS connection code is tested against both protocol versions 

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
The change is transparent to user. It will enable the clients to be able to use TLSv1.3 automatically.

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/tests/TlsNegotiationTest.cc
+++ b/tests/TlsNegotiationTest.cc
@@ -17,14 +17,16 @@
  * under the License.
  */
 #include <gtest/gtest.h>
-#include <pulsar/Client.h>
+#include <openssl/ssl.h>
 #include <pulsar/Authentication.h>
-#include "lib/AsioDefines.h"
+#include <pulsar/Client.h>
+
+#include <atomic>
 #include <future>
 #include <thread>
-#include <atomic>
+
+#include "lib/AsioDefines.h"
 #include "lib/LogUtils.h"
-#include <openssl/ssl.h>
 
 #ifdef USE_ASIO
 #include <asio.hpp>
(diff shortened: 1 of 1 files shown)


BewareMyPower:
Thanks for your contribution, could you fix the format error?

PR #537, opened 2026-02-04
Fix use-after-free crash in `getBrokerConsumerStatsAsync` when connection is closed

### LINKED ISSUE #536: [Bug] Crash in getBrokerConsumerStatsAsync when connection is closed
### Search before asking

### Version

3.6.0

### Minimal reproduce step

Not easy to reproduce yet.

No coredump.

### What did you see instead?

A coredump.

### Anything else?

The coredump stack:
Image

dragonls:
Fixes #536

### Motivation

A use-after-free crash occurs in `MultiTopicsConsumerImpl::getBrokerConsumerStatsAsync()` when the underlying connection is closed while an asynchronous consumer stats request is pending.

The root cause has two aspects:

1. **Reference capture of stack variables**: The lambda captures local variables (`latchPtr`, `statsPtr`, `i`) by reference. If the callback is invoked after the function returns, these references become dangling.

2. **Direct `this` capture in lambda**: Both the outer and inner lambdas capture `this` directly. Although there's a `weakSelf.lock()` check, the actual function call `handleGetConsumerStats(...)` is invoked through the captured `this` pointer (implicit `this->handleGetConsumerStats(...)`), **not through `self->`**. When `ClientConnection::close()` calls `setFailed()` on pending promises, it synchronously triggers the registered callbacks, potentially causing use-after-free if the `MultiTopicsConsumerImpl` has been destroyed.

### Modifications

1. **Use `shared_ptr<atomic>` for index**: Replace stack variable `size_t i` with `std::shared_ptr<std::atomic<size_t>>` for thread-safe indexing and proper lifetime management.

2. **Create `weakSelf` outside lambdas**: Move `weak_from_this()` call outside to avoid capturing `this` in the outer lambda.

3. **Remove `this` capture from all lambdas**: Neither the outer nor inner lambda captures `this` anymore.

4. **Call member function through `self->`**: Change `handleGetConsumerStats(...)` to `self->handleGetConsumerStats(...)` to ensure the call goes through the `shared_ptr`, not the raw `this` pointer.

### Verifying this change

The crash scenario is difficult to reproduce in unit tests as it requires precise timing of connection closure during async operations. However, the fix follows the standard C++ weak_ptr/shared_ptr pattern for preventing use-after-free in asynchronous callbacks.

### Documentation

- [ ] `doc-required` 

- [x] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [ ] `doc-complete`

### DIFF
--- a/lib/MultiTopicsConsumerImpl.cc
+++ b/lib/MultiTopicsConsumerImpl.cc
@@ -850,17 +850,17 @@ void MultiTopicsConsumerImpl::getBrokerConsumerStatsAsync(const BrokerConsumerSt
     LatchPtr latchPtr = std::make_shared<Latch>(numberTopicPartitions_->load());
     lock.unlock();
 
-    size_t i = 0;
-    consumers_.forEachValue([this, &latchPtr, &statsPtr, &i, callback](const ConsumerImplPtr& consumer) {
-        size_t index = i++;
-        auto weakSelf = weak_from_this();
-        consumer->getBrokerConsumerStatsAsync([this, weakSelf, latchPtr, statsPtr, index, callback](
-                                                  Result result, const BrokerConsumerStats& stats) {
-            auto self = weakSelf.lock();
-            if (self) {
-                handleGetConsumerStats(result, stats, latchPtr, statsPtr, index, callback);
-            }
-        });
+    auto indexPtr = std::make_shared<std::atomic<size_t>>(0);
+    auto weakSelf = weak_from_this();
+    consumers_.forEachValue([weakSelf, latchPtr, statsPtr, indexPtr, callback](const ConsumerImplPtr& consumer) {
+        size_t index = indexPtr->fetch_add(1);
+        consumer->getBrokerConsumerStatsAsync([weakSelf, latchPtr, statsPtr, index, callback](
+            Result result, const BrokerConsumerStats& stats) {
+                auto self = weakSelf.lock();
+                if (self) {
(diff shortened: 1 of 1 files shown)


lib/MultiTopicsConsumerImpl.cc
```diff
@@ -850,17 +850,17 @@ void MultiTopicsConsumerImpl::getBrokerConsumerStatsAsync(const BrokerConsumerSt
...
-        consumer->getBrokerConsumerStatsAsync([this, weakSelf, latchPtr, statsPtr, index, callback](
-                                                  Result result, const BrokerConsumerStats& stats) {
-            auto self = weakSelf.lock();
-            if (self) {
-                handleGetConsumerStats(result, stats, latchPtr, statsPtr, index, callback);
-            }
-        });
+    auto indexPtr = std::make_shared<std::atomic<size_t>>(0);
+    auto weakSelf = weak_from_this();
+    consumers_.forEachValue([weakSelf, latchPtr, statsPtr, indexPtr, callback](const ConsumerImplPtr& consumer) {
```
BewareMyPower: The callback of `forEachValue` is immediately executed in the current method.

https://github.com/apache/pulsar-client-cpp/blob/0b893a1622aa7dd702d48afee33af321bb2c06dd/lib/SynchronizedHashMap.h#L95-L99

so I believe this patch does not fix the actual issue

BewareMyPower:
I believe I've figured out the cause, let me create a new fix

Issue #550, opened 2026-03-09
[Bug] Client.close might crash


BewareMyPower:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/pulsar-client-cpp/issues) and found nothing similar.

### Version

0c6a7c0a688409e6f2c4b066b48bdca24e5ec33b

### Minimal reproduce step

Run `testCloseClient` for a few times:

```bash
./build/tests/pulsar-tests --gtest_filter='*testCloseClient'
```

### What did you expect to see?

passed

### What did you see instead?

You might see the segmentation fault:

```
2026-03-09 11:22:47.077 ERROR [0x1ff497840] ClientImpl:228 | Error Checking/Getting Partition Metadata while creating producer on persistent://public/default/client-test-close-client-1773026566 -- ResultDisconnected
2026-03-09 11:22:47.077 INFO  [0x1ff497840] ClientImpl:687 | Closing Pulsar client with 0 producers and 0 consumers
2026-03-09 11:22:47.077 INFO  [0x16bb9f000] ClientConnection:1353 | [<none> -> pulsar://localhost:6650] Connection disconnected (refCnt: 1)
zsh: segmentation fault  ./build/tests/pulsar-tests --gtest_filter='*testCloseClient'
```

### Anything else?

It seems that there was a regression on https://github.com/apache/pulsar-client-cpp/pull/136.

```
* thread #2, stop reason = EXC_BAD_ACCESS (code=1, address=0xc0c4bd980148)
    frame #0: 0x00000001000825d8 pulsar-tests`asio::detail::reactive_socket_service_base::close(this=0x0000c0c4bd980148, impl=0x0000600002f50128, ec=0x000000016fe866a0) at reactive_socket_service_base.ipp:109:5
   106 	    ASIO_HANDLER_OPERATION((reactor_.context(),
   107 	          "socket", &impl, impl.socket_, "close"));
   108
-> 109 	    reactor_.deregister_descriptor(impl.socket_, impl.reactor_data_,
   110 	        (impl.state_ & socket_ops::possible_dup) == 0);
   111
   112 	    socket_ops::close(impl.socket_, impl.state_, false, ec);
Target 0: (pulsar-tests) stopped.
(lldb) bt
* thread #2, stop reason = EXC_BAD_ACCESS (code=1, address=0xc0c4bd980148)
[...]

PR #551, opened 2026-03-09
Fix crash due to asio object lifetime and thread safety issue

BewareMyPower:
Fixes: https://github.com/apache/pulsar-client-cpp/issues/550

### Motivation

`ClientTest.testCloseClient` is flaky that sometimes it could cause segmentation fault. This is because `client.close()` could be called quickly after `createProducerAsync`. In this case, the `async_connect` operation might not complete. In the `close` method, `socket_->close()` is called, which could complete the `async_connect` with `operation_aborted`. In asio, when the completion handler (callback) is called while the caller (socket) is destroyed, it will crash because the callback refers the caller internally.

There are several issues.

#### Issue 1: Thread safety issue when accessing asio objects (e.g. socket)

`ip::tcp::socket::close` is called by `ClientConnection::close`, which could be called in a separated thread created by `ClientImpl::handleClose` or directly by `ClientImpl::shutdown`. However, it's unsafe to access `socket` concurrently, see https://think-async.com/Asio/asio-1.36.0/doc/asio/reference/ip__tcp/socket.html

> Shared objects: Unsafe.
>
> Synchronous send, receive, connect, and shutdown operations are thread safe with respect to each other, if the underlying operating system calls are also thread safe. This means that it is permitted to perform concurrent calls to these synchronous operations on a single socket object. Other synchronous operations, such as open or close, are not thread safe.

`ssl::stream` has the same issue, that could be why the operations on `tlsSocket_` were bound to `strand_` before. Actually, it's also unsafe to call `async_write` directly on `socket_` because it could be called by `ProducerImpl::sendMessage` in a different thread from the event loop thread.

#### Issue 2: The lifetime of asio objects might be shorter than the callbacks

https://github.com/apache/pulsar-client-cpp/pull/317 might not fix the null access issue correctly. It introduced several crash issues (#325, #326) by capturing a `weak_ptr`. The issue is that `ConnectionPool::close` don't wait until all pending asynchronous operations are done. Then, after the `ClientConnection` object is destroyed, the asio objects are also destroyed. However, the `io_context` could still try accessing them when destroying the internal callback queue.

### Modifications

- Capture `shared_from_this()` in all callbacks of asynchronous operations.
- In `close`, dispatch the `socket::close()` call in event loop thread. For `ssl::stream` object, call `async_shutdown`. The previous close on `lower_layer` is actually equivalent with closing the socket.
- In `ConnectionPool::close`, wait until all pending operations are done by checking the reference count of `ClientConnection`
- Replace `PeriodicalTask` with a simple timer and cancel it in `close`
- Fix some tests failing in CI

### DIFF
--- a/tests/ClientTest.cc
+++ b/tests/ClientTest.cc
@@ -21,12 +21,12 @@
 #include <pulsar/Version.h>
 
 #include <algorithm>
+#include <atomic>
 #include <chrono>
-#include <condition_variable>
 #include <future>
-#include <mutex>
 #include <sstream>
 #include <thread>
+#include <utility>
 
 #include "MockClientImpl.h"
 #include "PulsarAdminHelper.h"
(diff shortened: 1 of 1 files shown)


lib/ClientConnection.h
```diff
@@ -219,28 +218,51 @@ class PULSAR_PUBLIC ClientConnection : public std::enable_shared_from_this<Clien
...
 
    private:
     struct PendingRequestData {
         Promise<Result, ResponseData> promise;
         DeadlineTimerPtr timer;
         std::shared_ptr<std::atomic_bool> hasGotResponse{std::make_shared<std::atomic_bool>(false)};
+
+        void fail(Result result) {
+            cancelTimer(*timer);
+            ;
```
shibd: useless line

PR #553, opened 2026-03-16
[fix][client-cpp] Fix Reader segfault when messageListenerThreads=0

### LINKED ISSUE #500: [Bug] Segmentation Fault (Null Pointer Dereference) in Pulsar C++ Client During Message Reception (v3.3.0 & v3.7.2)
### Search before asking

### Version

We observed a segmentation fault on our database node while using the Pulsar C++ client version 3.3.0, which leads to a crash with the following stack trace (see below). The crash appears to be caused by a null pointer dereference during message processing in the pulsar::ConsumerImpl::messageReceived() function.

To investigate, we upgraded the client to version 3.7.2, but unfortunately, the issue persists with the same stack trace, indicating that the underlying problem might still exist in the latest versions.``

### Minimal reproduce step

To reproduce the issue using the Pulsar C++ client (versions 3.3.0 and 3.7.2), follow these steps:

1) Set up a running Pulsar cluster.
2) Link and build an application with the Pulsar C++ client (either v3.3.0 or v3.7.2).
3) In application, create a Reader on an existing topic as follows:
`auto result = client_.createReader(topic_, start_mid_, readerConfig, reader_);`

4)Periodically perform these operations:
      Check message availability:
    ```
    bool hasMsgInTopic = false;
    reader_.hasMessageAvailable(hasMsgInTopic);
    ```

     If messages are available, read the next one:
     `result = reader_.readNext(message, timeout);`

5) Under moderate to high throughput or runtime duration, the application eventually crashes with a segmentation fault, pointing to:

`pulsar::ConsumerImpl::messageReceived(...)`

The Reader should be able to:
     Safely check if messages are available using hasMessageAvailable()
     Reliably read messages using readNext()

The client should not crash due to a null pointer dereference or segmentation fault.
Internal message handling (e.g., via messageReceived) should gracefully handle malformed or unexpected broker messages.

### What did you see instead?

**below stack trace:** 

`Backtrace Generated by Error Signal: [0x000000000000000b] PID: [0x0000000000112bdb] PC: [0x00007fdfc366f834] FP: [0x00007fdeafd17190] SIGSEGV: SEGV_MAPERR SI_ADDR : [0x0000000000000000] (Basics::Backtrace::DoBacktrace(int, int, void*, void*) 0x8bd) [0x7abec7d] (Basics::GlobalSignalHandlers::logFatalSignal(int, void*, void*) 0x106) [0x7ba8156] (Basics::sigHandler_withinATryCatch(int, siginfo_t*, void*) 0x2b8) [0x7ba9308] (Basics::sigHandler(int, siginfo_t*, void*) 0xf) [0x7ba939f] (__sigaction 0x50) [0x7fe5cee3ebf0] (pulsar::ConsumerImpl::messageReceived(std::shared_ptr<pulsar::ClientConnection> const&, pulsar::proto::CommandMessage const&, bool&, pulsar::proto::BrokerEntryMetadata&, pulsar::proto::MessageMetadata&, pulsar::SharedBuffer&) 0x204) [0x7fdfc366f834] (pulsar::ClientConnection::handleIncomingMessage(pulsar::proto::CommandMessage const&, bool, pulsar::proto::BrokerEntryMetadata&, pulsar::proto::MessageMetadata&, pulsar::SharedBuffer&) 0x25b) [0x7fdfc377d56b] (pulsar::ClientConnection::processIncomingBuffer() 0x316) [0x7fdfc37a7176] (UNKNOWN) [0x7fdfc37b6c70] (UNKNOWN) [0x7fdfc37b8be0] (boost::asio::detail::strand_executor_service::invoker<boost::asio::io_context::basic_executor_type<std::allocator<void>, 4ul> const, void>::operator()() 0x82) [0x7fdfc37c9512] (void boost::asio::io_context::basic_executor_type<std::allocator<void>, 4ul>::execute<boost::asio::detail::strand_executor_service::invoker<boost::asio::io_context::basic_executor_type<std::allocator<void>, 4ul> const, void> >(boost::asio::detail::strand_executor_service::invoker<boost::asio::io_context::basic_executor_type<std::allocator<void>, 4ul> const, void>&&) const 0x98) [0x7fdfc37c9778] (UNKNOWN) [0x7fdfc37b8718] (boost::asio::detail::scheduler::do_run_one(boost::asio::detail::conditionally_enabled_mutex::scoped_lock&, boost::asio::detail::scheduler_thread_info&, boost::system::error_code const&) 0x367) [0x7fdfc3687637] (boost::asio::detail::scheduler::run(boost::system::error_code&).isra.0 0xf9) [0x7fdfc3680e19] (pulsar::ExecutorService::start()::{lambda()#1}::operator()() const 0x91) [0x7fdfc36810b1] (std::error_code::default_error_condition() const 0x34) [0x7fe5cf2dbad4] (start_thread 0x31a) [0x7fe5cee8a19a] (__clone3 0x30) [0x7fe5cef0f240] END BACKTRACE NO THREAD CONTEXT`

### Anything else?

_No response_

- [ ] I'm willing to submit a PR!

zhanglistar:
Fixes #500 

### Motivation

When using a Reader with `ClientConfiguration::setMessageListenerThreads(0)` (or when the listener executor is otherwise null), the application can crash with a segmentation fault in `ConsumerImpl::messageReceived()`. Two root causes were identified:

1. **ExecutorServiceProvider(0)** – When `getMessageListenerThreads()` is 0, the provider is constructed with an empty executor list. In `get()`, the code does `idx %= executors_.size()`, which is **undefined behavior** when `size() == 0` (division by zero). This can lead to wrong or null executor pointers.

2. **Null listenerExecutor_** – The Reader creates its internal consumer with a null `ExecutorServicePtr`. The consumer then falls back to `client->getListenerExecutorProvider()->get()`. If that returns null (e.g. due to the above), `messageReceived` and `executeNotifyCallback` call `listenerExecutor_->postWork(...)` without a null check, causing a null pointer dereference (SIGSEGV at address 0x0).

This was observed in the wild with Reader usage (e.g. periodic `hasMessageAvailable()` and `readNext()`) under moderate to high load or long runtime, with stack traces pointing at `pulsar::ConsumerImpl::messageReceived`.

### Modifications

- **lib/ExecutorService.cc**
  - Construct with `executors_(std::max(1, nthreads))` so the executor list is never empty when `nthreads == 0`, avoiding undefined behavior in `get()`.
  - In `get()`, add an early `if (executors_.empty()) return nullptr;` guard before using `executors_.size()`.
  - Add `#include <algorithm>` for `std::max`.

- **lib/ConsumerImpl.cc**
  - In `messageReceived()`, after the closing-state check, add a check for null `listenerExecutor_`: log an error, call `increaseAvailablePermits(cnx)`, and return to avoid dereferencing null.
  - In the `messageListener_` block, only call `listenerExecutor_->postWork(...)` when `listenerExecutor_` is non-null.
  - In `executeNotifyCallback()`, when notifying a pending async receive: if `listenerExecutor_` is non-null, post the callback as before; otherwise call `notifyPendingReceivedCallback(...)` on the current thread so the callback is still invoked and no null dereference occurs.

### Verifying this change

- **ExecutorServiceProviderTest.ZeroThreadsReturnsValidExecutor** (unit test, no broker): Creates `ExecutorServiceProvider(0)`, calls `get()` three times, and asserts each return is non-null. Confirms that 0-thread configuration no longer leads to undefined behavior and that a valid executor is always returned.
- **ReaderTest.testReaderWithZeroMessageListenerThreads** (integration test, requires Pulsar broker): Creates a `Client` with `setMessageListenerThreads(0)`, creates a Reader, produces several messages, then in a loop calls `hasMessageAvailable()` and `readNext()` until all messages are consumed. Asserts the correct count and content. Verifies that Reader path with 0 listener threads does not segfault and delivers messages correctly.

Run the unit test (no broker):

```bash
./tests/pulsar-tests --gtest_filter='ExecutorServiceProviderTest.ZeroThreadsReturnsValidExecutor'
```

Run the Reader integration test (with broker at `pulsar://localhost:6650`):

```bash
./tests/pulsar-tests --gtest_filter='ReaderTest.testReaderWithZeroMessageListenerThreads'
```

### Documentation

- [ ] `doc-required`

- [x] `doc-not-needed`  
  (Behavior change is internal and backward-compatible: 0 listener threads is now handled safely; no API or user-facing doc update required.)

- [ ] `doc`

- [ ] `doc-complete`

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -622,6 +622,11 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
     if (state == Closing || state == Closed) {
         return;
     }
+    if (!listenerExecutor_) {
+        LOG_ERROR(getName() << " listenerExecutor_ is null, discarding message to avoid null dereference");
+        increaseAvailablePermits(cnx);
+        return;
+    }
     uint32_t numOfMessageReceived = m.impl_->metadata.num_messages_in_batch();
     if (ackGroupingTrackerPtr_->isDuplicate(m.getMessageId())) {
         LOG_DEBUG(getName() << " Ignoring message as it was ACKed earlier by same consumer.");
@@ -663,8 +668,10 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
             return;
         }
         // Trigger message listener callback in a separate thread
-        while (numOfMessageReceived--) {
-            listenerExecutor_->postWork(std::bind(&ConsumerImpl::internalListener, get_shared_this_ptr()));
+        if (listenerExecutor_) {
+            while (numOfMessageReceived--) {
+                listenerExecutor_->postWork(std::bind(&ConsumerImpl::internalListener, get_shared_this_ptr()));
+            }
         }
     }
 }
@@ -713,8 +720,12 @@ void ConsumerImpl::executeNotifyCallback(Message& msg) {
 
     // has pending receive, direct callback.
     if (asyncReceivedWaiting) {
-        listenerExecutor_->postWork(std::bind(&ConsumerImpl::notifyPendingReceivedCallback,
-                                              get_shared_this_ptr(), ResultOk, msg, callback));
+        if (listenerExecutor_) {
+            listenerExecutor_->postWork(std::bind(&ConsumerImpl::notifyPendingReceivedCallback,
+                                                  get_shared_this_ptr(), ResultOk, msg, callback));
+        } else {
+            notifyPendingReceivedCallback(ResultOk, msg, callback);
+        }
         return;
     }
 
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -670,7 +670,8 @@ void ConsumerImpl::messageReceived(const ClientConnectionPtr& cnx, const proto::
         // Trigger message listener callback in a separate thread
         if (listenerExecutor_) {
             while (numOfMessageReceived--) {
-                listenerExecutor_->postWork(std::bind(&ConsumerImpl::internalListener, get_shared_this_ptr()));
+                listenerExecutor_->postWork(
+                    std::bind(&ConsumerImpl::internalListener, get_shared_this_ptr()));
             }
         }
     }
--- a/lib/ExecutorService.cc
+++ b/lib/ExecutorService.cc
@@ -18,6 +18,8 @@
  */
 #include "ExecutorService.h"
 
+#include <algorithm>
+
 #include "LogUtils.h"
 #include "TimeUtils.h"
 DECLARE_LOG_OBJECT()
@@ -130,9 +132,12 @@ void ExecutorService::postWork(std::function<void(void)> task) { ASIO::post(io_c
 /////////////////////
 
 ExecutorServiceProvider::ExecutorServiceProvider(int nthreads)
-    : executors_(nthreads), executorIdx_(0), mutex_() {}
+    : executors_(std::max(1, nthreads)), executorIdx_(0), mutex_() {}
 
 ExecutorServicePtr ExecutorServiceProvider::get(size_t idx) {
+    if (executors_.empty()) {
+        return nullptr;
+    }
     idx %= executors_.size();
     Lock lock(mutex_);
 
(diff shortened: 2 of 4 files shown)


BewareMyPower (approved):

vikram-chandra-19:
Hi @zhanglistar 
I need this fix for the Pulsar C++ client version 3.3.0. Would it be safe to cherry-pick this fix into that version?

PR #556, opened 2026-03-19
Fix hasMessageAvailable will return true after seeking to a timestamp newer than the last message

BewareMyPower:
We should ignore the `startMessageIdInclusive` config when the consumer has sought by timestamp because the broker side will ignore this option as well when processing a Seek request.

### DIFF
--- a/tests/ReaderTest.cc
+++ b/tests/ReaderTest.cc
@@ -991,7 +991,9 @@ TEST_P(ReaderSeekTest, testSeekToEndByTimestamp) {
     ASSERT_EQ(ResultOk, client.createReader(topic, MessageId::earliest(), readerConf, reader));
 
     ASSERT_EQ(ResultOk, producer.send(MessageBuilder().setContent("msg").build()));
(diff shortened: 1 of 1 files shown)


tests/ReaderTest.cc
```diff
@@ -983,6 +978,26 @@ TEST_F(ReaderSeekTest, testSeekInclusiveChunkMessage) {
...
+    ASSERT_EQ(ResultOk, client.createProducer(topic, producer));
+
+    ReaderConfiguration readerConf;
+    readerConf.setStartMessageIdInclusive(GetParam());
+
+    Reader reader;
+    ASSERT_EQ(ResultOk, client.createReader(topic, MessageId::earliest(), readerConf, reader));
+
+    ASSERT_EQ(ResultOk, producer.send(MessageBuilder().setContent("msg").build()));
+    auto now = TimeUtils::currentTimeMillis() + 1000;
```
RobertIndie: Why we add `+ 1000` here?

PR #579, opened 2026-05-22
feat: introduce a v2 createProducer API to carry error message when fail

BewareMyPower:


tests/MessageChunkingTest.cc
```diff
@@ -102,7 +104,16 @@ TEST_F(MessageChunkingTest, testInvalidConfig) {
     conf.setBatchingEnabled(true);
     conf.setChunkingEnabled(true);
     Producer producer;
-    ASSERT_THROW(client.createProducer("xxx", conf, producer), std::invalid_argument);
+    ASSERT_EQ(ResultInvalidConfiguration, client.createProducer("xxx", conf, producer));
```
RobertIndie: This seems to introduce a breaking change.
Before: throwing the invalid argument exception.
After: returning `ResultInvalidConfiguration`.

Do you think we should avoid this breaking change?

BewareMyPower: Though I don't think it makes a big difference and using error code in asynchronous method is safer, I will still revert this unnecessary change from this PR.

RobertIndie (approved):

PR #582, opened 2026-05-28
Fix macOS cross-architecture build on arm64 when target arch is x86_64

BewareMyPower:
Fix failure like https://github.com/apache/pulsar-client-cpp/actions/runs/26558894568/job/78281724880

The runner is `macos-14`, which is `arm64`, but the workflow sets:

```bash
CMAKE_OSX_ARCHITECTURES=x86_64
```

The old CMake logic checked only `CMAKE_SYSTEM_PROCESSOR == x86_64`. On that runner, CMAKE_SYSTEM_PROCESSOR still reports the host architecture, arm64, so CMake skipped the x86 compile flags: `-msse4.2 -mpclmul`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,6 +66,14 @@ if(CCACHE_PROGRAM)
 endif(CCACHE_PROGRAM)
 
 MESSAGE(STATUS "ARCHITECTURE: ${CMAKE_SYSTEM_PROCESSOR}")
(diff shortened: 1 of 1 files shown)


BewareMyPower:
Let's see if the workflow here would succeed: https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/26575230054

BewareMyPower:
image
The candidate release is fine.

shibd (approved):

PR #586, opened 2026-06-01
Fix release workflows

### LINKED ISSUE #559: dockerbuild files are uploaded in "CI - Build binary artifacts" workflow
See https://github.com/apache/pulsar-client-cpp/actions/runs/23469538356

Image

We should skip the `dockerbuild` files:

```diff
diff --git a/build-support/download-release-artifacts.py b/build-support/download-release-artifacts.py
index 9b1866f..8ccd0f8 100755
--- a/build-support/download-release-artifacts.py
+++ b/build-support/download-release-artifacts.py
@@ -47,7 +47,7 @@ with requests.get(workflow_run_url, headers=headers) as response:
     for artifact in data['artifacts']:
         name = artifact['name']
         # Skip debug artifact
-        if name.endswith("-Debug"):
+        if name.endswith("-Debug") or name.find("dockerbuild") >= 0:
             continue
         dest_dir = os.path.join(dest_path, name)
         if name.find("windows") >= 0 and os.path.exists(dest_dir + ".tar.gz"):
```

### LINKED ISSUE #585: [Bug] The build time on arm64 might take too long
### Search before asking

### Version

f2c0fec

### Minimal reproduce step

https://github.com/apache/pulsar-client-cpp/actions/runs/26613317193/job/78778604980

n/a

### What did you see instead?

The task is cancelled after 6 hours.

### Anything else?

Here's the analysis:

## Root Cause: GitHub Actions 6-Hour Hard Limit

**Timeline:**
- Job start: `2026-06-01T02:41:44`
- Cancellation: `2026-06-01T08:41:56`
- Duration: **exactly 6 hours**

The `timeout-minutes: 500` (~8.3h) in the workflow YAML is irrelevant here — **GitHub Actions enforces a hard 6-hour wall-clock limit per job** for GitHub-hosted runners, regardless of what `timeout-minutes` is set to.

## What Took So Long

The job was building `Build Alpine arm64` via **Docker/QEMU emulation** (cross-compiling arm64 on an x86_64 runner), and vcpkg had to build all 87 dependencies from source with no cache hits:

| Package | Time |
|---|---|
| `protobuf:arm64-linux` | **3.2 hours** |
| `zstd:arm64-linux` | 8.4 minutes |
| `snappy:arm64-linux` | 1.4 minutes |
| ... (84 others) | remaining time |

The vcpkg install only just finished at `08:41:48` (all 87 packages done), and the job was immediately cancelled at `08:41:56` — literally seconds after vcpkg completed, before the actual cmake build of pulsar-client-cpp even started.

## Fix Options

1. **Use a vcpkg binary cache** (GitHub Actions cache) — so packages like protobuf don't rebuild from scratch every run. This is the most impactful fix.
2. **Use a native arm64 runner** — QEMU emulation is ~10-20x slower than native; native arm64 GitHub-hosted runners are available.
3. **Both** — binary cache + native arm64 runners would bring vcpkg install from hours down to minutes.

BewareMyPower:
fixes #585 
fixes #559

All changes are in place. Here's a summary of what was done:

---

### 1. `ci-build-binary-artifacts.yaml`

**Fix 1 — Native arm64 runners:**
- Added `runner` field to the CPU matrix: `x86_64` → `ubuntu-22.04`, `arm64` → `ubuntu-22.04-arm`
- Changed `runs-on: ubuntu-22.04` → `runs-on: ${{matrix.cpu.runner}}`
- Removed `Set up QEMU` step entirely (not needed when each architecture runs on its own native runner)

**Fix 2 — vcpkg binary cache:**
- Added `actions/cache@v4` step to restore/save `~/.cache/vcpkg/archives` for all 6 Linux jobs (keyed by `pkg-type + arch + vcpkg.json hash`)
- Added the same for macOS (`~/Library/Caches/vcpkg/archives`)
- Fixed Docker GHA layer cache to use per-job `scope` to avoid the 6 matrix jobs thrashing each other's cache

### 2. All 6 `docker-build-{type}-{arch}.sh` scripts

- Added `mkdir -p $HOME/.cache/vcpkg/archives` before `docker run`
- Added `-v $HOME/.cache/vcpkg/archives:/root/.cache/vcpkg/archives` volume mount so vcpkg inside the container reads/writes the cache restored by `actions/cache` on the runner host

**Expected result:** On the next tag push, `protobuf:arm64-linux` should build in minutes on the native arm64 runner (vs 3.2 hours under QEMU), and subsequent runs will skip rebuilding already-cached packages entirely.

### DIFF
--- a/.github/workflows/ci-build-binary-artifacts.yaml
+++ b/.github/workflows/ci-build-binary-artifacts.yaml
@@ -31,7 +31,7 @@ jobs:
 
   package-linux:
     name: Build ${{matrix.pkg.name}} ${{matrix.cpu.platform}}
-    runs-on: ubuntu-22.04
+    runs-on: ${{matrix.cpu.runner}}
     timeout-minutes: 500
 
     strategy:
@@ -42,15 +42,20 @@ jobs:
           - { name: 'Deb', type: 'deb', path: 'pkg/deb/BUILD/DEB' }
           - { name: 'Alpine', type: 'apk', path: 'pkg/apk/build' }
         cpu:
-          - { arch: 'x86_64', platform: 'x86_64' }
-          - { arch: 'aarch64', platform: 'arm64' }
+          - { arch: 'x86_64', platform: 'x86_64', runner: 'ubuntu-22.04' }
+          - { arch: 'aarch64', platform: 'arm64', runner: 'ubuntu-22.04-arm' }
 
     steps:
       - name: checkout
         uses: actions/checkout@v3
 
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a
+      - name: Restore vcpkg binary cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/vcpkg/archives
+          key: vcpkg-linux-${{matrix.pkg.type}}-${{matrix.cpu.platform}}-${{hashFiles('vcpkg.json')}}
+          restore-keys: |
+            vcpkg-linux-${{matrix.pkg.type}}-${{matrix.cpu.platform}}-
 
       - name: Package Pulsar source
         run: build-support/generate-source-archive.sh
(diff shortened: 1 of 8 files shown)


BewareMyPower:
Let's wait for the result in https://github.com/BewareMyPower/pulsar-client-cpp/actions/runs/26755689473

shibd (approved):

PR #588, opened 2026-06-17
[feat] Add httpLookupAuthAllowRedirect option to forward auth credentials on HTTP lookup redirects

geniusjoe:
Master Issue: https://github.com/apache/pulsar-client-cpp/pull/313

### Motivation

When using HTTP-based lookup service with authentication enabled, the broker may respond with an HTTP redirect (3xx) to the correct owner broker if the initial broker is not the owner of the requested topic. However, by default, libcurl strips the `Authorization` header on cross-origin redirects for security reasons. This causes the redirected request to fail with a `401 Unauthorized` error on the target broker.

This PR introduces a new client configuration option `httpLookupAuthAllowRedirect` to allow forwarding authentication credentials when following HTTP lookup redirects.

### Modifications

- Added `setHttpLookupAuthAllowRedirect` / `isHttpLookupAuthAllowRedirect` methods to `ClientConfiguration` (C++ and C API).
- Added `httpLookupAuthAllowRedirect` field to `ClientConfigurationImpl`.
- Added `authAllowRedirect` option to `CurlWrapper::Options`, which sets `CURLOPT_UNRESTRICTED_AUTH` when enabled.
- Propagated the configuration from `ClientConfiguration` through `HTTPLookupService` to `CurlWrapper` during HTTP lookup requests.
- Added unit tests for both C and C++ APIs.

### Verifying this change

This change is already covered by existing tests, such as:
- Added unit tests in `tests/c/c_ClientConfigurationTest.cc` to verify the C API getter/setter for `httpLookupAuthAllowRedirect` (default value, set to true, set back to false).

### Documentation

- [x] `doc-not-needed` 
(The new configuration option is self-documented via its API doc comments in `ClientConfiguration.h`. No external documentation update is required.)

### Workflow test 
https://github.com/geniusjoe/pulsar-client-cpp/pull/1

### DIFF
--- a/include/pulsar/ClientConfiguration.h
+++ b/include/pulsar/ClientConfiguration.h
@@ -187,6 +187,33 @@ class PULSAR_PUBLIC ClientConfiguration {
     int getMaxBackoffIntervalMs() const;
 
     /**
+     * Configure whether to send authentication credentials when following HTTP redirects
+     * to a different host during lookup requests.
+     *
+     * When enabled, the Authorization header will be forwarded on cross-origin redirects.
+     *
+     * HTTP lookup redirects typically occur when the broker receiving the lookup request
+     * is not the owner of the requested topic. In this case, the broker responds with
+     * an HTTP redirect (3xx) pointing to the correct owner broker. If authentication is
+     * enabled, the redirected request needs to carry the auth credentials to be accepted
+     * by the target broker.
+     *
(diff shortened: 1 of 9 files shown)


geniusjoe:
@BewareMyPower 
Hello, do you have time to help review this PR? Thank you very much.

wolfstudy (approved):
Good jobs, LGTM +1

PR #595, opened 2026-06-23
Avoid object pool for message builder

### LINKED ISSUE #476: When use pulsar::Message msg = pulsar::MessageBuilder().setContent(sMsg).build() to Create Message, sometimes it will occur segment fault
(gdb) bt
#0  pop (this=0x7f2b58df7900) at pulsar-client-3.4/lib/ObjectPool.h:74
#1  allocate (this=0x7f2b58df7900) at pulsar-client-3.4/lib/ObjectPool.h:134
#2  allocate (n=1, this=<synthetic pointer>) at pulsar-client-3.4/lib/ObjectPool.h:167
#3  allocate (__a=<synthetic pointer>, __n=1) at /usr/local/gcc-4.9.3/include/c++/4.9.3/bits/alloc_traits.h:357
#4  __shared_count<pulsar::MessageImpl, pulsar::Allocator<pulsar::MessageImpl, 100000> > (__a=..., this=0x7f2b3e589ea8)
    at /usr/local/gcc-4.9.3/include/c++/4.9.3/bits/shared_ptr_base.h:616
#5  __shared_ptr<pulsar::Allocator<pulsar::MessageImpl, 100000> > (__a=..., __tag=..., this=0x7f2b3e589ea0, 
    this@entry=<error reading variable: Cannot access memory at address 0x7f2b3e589e98>)
    at /usr/local/gcc-4.9.3/include/c++/4.9.3/bits/shared_ptr_base.h:1090
#6  shared_ptr<pulsar::Allocator<pulsar::MessageImpl, 100000> > (__a=..., __tag=..., this=0x7f2b3e589ea0)
    at /usr/local/gcc-4.9.3/include/c++/4.9.3/bits/shared_ptr.h:316
#7  allocate_shared<pulsar::MessageImpl, pulsar::Allocator<pulsar::MessageImpl, 100000> > (__a=...)
    at /usr/local/gcc-4.9.3/include/c++/4.9.3/bits/shared_ptr.h:588
#8  create (this=0x2622ba0 <pulsar::messagePool>) at pulsar-client-3.4/lib/ObjectPool.h:212
#9  pulsar::MessageBuilder::createMessageImpl () at pulsar-client-3.4/lib/MessageBuilder.cc:42

(gdb) print head_
Cannot access memory at address 0x7f2b58df7900
(gdb) print result
$1 = (void *) 0x1

void* pop() {
            if (!head_) {
                // size = 0
                std::lock_guard<std::mutex> lock(mutex_);

                if (!globalPool_) {
                    return NULL;
                }

                GlobalPool* poolEntry = globalPool_;
                head_ = globalPool_->node_;
                pushSize_ += globalPool_->nodeCount_;
                globalNodeCount_ -= globalPool_->nodeCount_;
                globalPool_ = globalPool_->next_;
                delete poolEntry;
            }
            void* result = head_;
            if (result) {
                head_ = head_->next;
                pushSize_--;
            }
            return result;
        }

zhanglistar:
Fixes #476 

Master Issue: #<xyz>

### Verifying this change

*(or)*

*(or)*

*(example:)*

### Documentation

- [ ] `doc-required` 

- [ ] `doc-not-needed` 
(Please explain why)

- [ ] `doc` 

- [x] `doc-complete`

### DIFF
--- a/tests/MessageTest.cc
+++ b/tests/MessageTest.cc
@@ -21,6 +21,8 @@
 #include <pulsar/MessageBuilder.h>
 
 #include <string>
+#include <thread>
+#include <vector>
 
 #include "PulsarFriend.h"
 #include "lib/MessageImpl.h"
@@ -101,6 +103,27 @@ TEST(MessageTest, testMessageBuilder) {
     {
         auto msg = MessageBuilder().setContent(std::move(value)).build();
         ASSERT_EQ(msg.getData(), originalAddress);
+    }
+}
+
+TEST(MessageTest, testMessageBuilderConcurrentBuild) {
+    const int threadCount = 8;
+    const int messagesPerThread = 10000;
+    std::vector<std::thread> threads;
+    threads.reserve(threadCount);
+
+    for (int i = 0; i < threadCount; i++) {
+        threads.emplace_back([i] {
+            for (int j = 0; j < messagesPerThread; j++) {
+                const std::string content = "message-" + std::to_string(i) + "-" + std::to_string(j);
+                auto msg = MessageBuilder().setContent(content).build();
+                ASSERT_EQ(content, msg.getDataAsString());
+            }
+        });
+    }
+
+    for (auto& thread : threads) {
+        thread.join();
     }
 }
 
--- a/lib/MessageBuilder.cc
+++ b/lib/MessageBuilder.cc
@@ -26,7 +26,6 @@
 #include "KeyValueImpl.h"
 #include "LogUtils.h"
 #include "MessageImpl.h"
-#include "ObjectPool.h"
 #include "PulsarApi.pb.h"
 #include "SharedBuffer.h"
 #include "TimeUtils.h"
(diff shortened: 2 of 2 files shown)


tests/MessageTest.cc
```diff
@@ -104,6 +106,27 @@ TEST(MessageTest, testMessageBuilder) {
     }
 }
 
+TEST(MessageTest, testMessageBuilderConcurrentBuild) {
```
BewareMyPower, line 109: This test seems meaningless. Thread safety of concurrent `shared_ptr` creation is guaranteed by the standard library.

zhanglistar: Check whether it will crash.

BewareMyPower: I mean, this is guaranteed by the standard library and our unit tests should only focus on the SDK side.

This does not like a regression test as well because the object pool is not used to avoid concurrent build crash.

BewareMyPower: From LLM:

> The test `testMessageBuilderConcurrentBuild` is **not meaningful**.
>
> Each thread creates its own independent `MessageBuilder` instances — there is no shared mutable state between threads. The test is equivalent to running the existing sequential `testMessageBuilder` test in parallel 80,000 times, which doesn't validate any concurrency behavior specific to the code under test. `std::make_shared` (the replacement) is inherently thread-safe, and `MessageBuilder` methods operate on per-instance `impl_` state with no global or shared data.
>
> To be meaningful, the test would need to exercise a shared resource — for example, if it used a shared pool (the old `ObjectPool`) or shared metadata state. As written, it's just a stress test of the heap allocator, which isn't what the change is about.

From myself: testing concurrent creation of `std::shared_ptr` objects is not commonly seen. Such tests are meaningless, otherwise, we should also test concurrent creation of `Producer`, `Consumer`, etc.

PR #598, opened 2026-06-23
Scalable Topics: new typed C++ SDK (`pulsar::st`) — API definition

merlimat:
> **Draft — API only.** This introduces the **public API surface** of a new typed C++ SDK for **scalable topics** (PIP-460 / 468 / 483), under a new `pulsar::st` namespace in the same library. It is **headers + examples only** — there is no `lib/st` implementation, C API, or protocol wiring yet. Opening it as a draft to gather design feedback on the API before building the backend.

## What's in this PR

- **`include/pulsar/st/*.h`** — the full public API, fully Doxygen-documented:
  - `PulsarClient` + `PulsarClientBuilder` (+ grouped config in `Policies.h`)
  - `Producer<T>` / `MessageBuilder<T>` / `ProducerBuilder<T>`
  - `StreamConsumer<T>` (ordered, cumulative-ack), `QueueConsumer<T>` (shared, individual-ack + DLQ), `CheckpointConsumer<T>` (externally-held position)
  - `Transaction`, `Message<T>` / `Messages<T>`, `MessageId`, `Checkpoint`
  - `Schema<T>` + `jsonSchema` / `avroSchema` / `protobufNativeSchema`
  - `Expected<T>` (sync results) and `Future<T>` (async results)
- **`include/pulsar/st/detail/*.h`** — the internal PIMPL boundary the implementation will fill in (`*Core` classes, `SharedState`).
- **`examples/st/*`** — usage examples (compile under C++20; not linkable until `lib/st` lands). The four core samples build dependency-free; the JSON-schema sample builds only where reflect-cpp is present.

### Typed facade over a byte-oriented core
`Schema<T>`, `Producer<T>`, `Message<T>`, `Consumer<T>` are thin header templates over a non-templated, type-erased core (the `detail::*Core` classes, to be implemented in `lib/st`). This gives a Java-like typed API while keeping the heavy code non-templated and out of headers — and lets the forthcoming C API bind to the same core.

### Errors are values (`Expected<T>`), not exceptions
Synchronous calls return `Expected<T>` — a `std::expected`-shaped, `[[nodiscard]]` value-or-`Error`. It is **non-throwing by default**; `.value()` opts into throwing `ClientException`.
- **Why not exceptions?** The operational path (send/receive/ack) can fail in bursts under load; throwing per-failure is costly and excludes `-fno-exceptions` embedders. The closest analogs — **gRPC (`grpc::Status`), librdkafka (error codes), the DataStax C/C++ driver (`CassError`)** — all use error values on the operational path, not exceptions.
- `[[nodiscard]]` closes the one real weakness of value-errors: silently ignoring them.

### Async is a continuation `Future<T>`, not `std::future`
`...Async` calls return a `Future<T>` with `addListener()` (react on completion, no blocking), `get()` / `get(timeout)` (block on demand), `thenApply()` (map), and `co_await` (C++20).
- **Why not `std::future`?** It has no continuation — you can only block on `.get()`, which defeats the purpose of async in an event-loop app. Every comparable library uses continuations instead: **folly (`.thenValue`), Seastar (`.then`, which explicitly rejects `std::future`), Asio (completion tokens), gRPC (callbacks), the DataStax driver (`cass_future_set_callback`)**.

### `pulsar::st` targets C++20 (the rest of the client stays C++17)
Only the new API requires C++20; existing headers and users are untouched (a `detail/Cxx20.h` guard emits a friendly error otherwise). This buys: **reflect-cpp** (zero-boilerplate JSON/Avro), **concepts** (clean `Schema`/SerDe constraints), **coroutines** (`co_await` on `Future<T>`), **`using enum`** (unqualified result codes), and **`operator<=>`**.

### Schema is a type-erased value over a pluggable SerDe
`Schema<T>` holds a SerDe (`info` / `encode` / `decode`). Built-ins cover primitives; structured types use:
- `protobufNativeSchema<T>()` — **fully automatic** (protobuf reflection; already a dependency),
- `jsonSchema<T>()` / `avroSchema<T>()` — **zero boilerplate** via reflect-cpp (pass a plain struct — nested structs and containers included),
- or a custom SerDe: `Schema<T>(serde)`.

This replaced an earlier type-keyed specialization model, which couldn't express "the same `T` encoded different ways" (JSON vs Avro).

### PIMPL via non-templated `*Core` classes (in `detail/`)
A header-only *templated* facade can't do the library's usual `.cc`-only PIMPL — a template must see a declaration of whatever it calls. So each entity has a non-templated `*Core` (`ProducerCore`, `StreamConsumerCore`, …, `ClientCore`) holding the hidden `*Impl`, with methods declared in the header and defined in `lib/st`; the typed classes are thin wrappers that forward to it, and `Future<T>::thenApply` carries the typed mapping. The cores live in `pulsar::st::detail` and under `detail/`, so they're out of the public namespace and excluded from generated docs.

### Consumer acks are fire-and-forget `void`
`acknowledge` / `acknowledgeCumulative` / `negativeAcknowledge` don't block and don't return an error — acks are buffered/best-effort and a lost ack just means redelivery (matching the new Java SDK). Transactional acks surface their outcome at `Transaction::commit()`.

### Client configuration grouped into policy structs
`ConnectionPolicy`, `BackoffPolicy`, `TlsPolicy`, `TransactionPolicy`, `MemorySize` (client) and `AckPolicy`, `DeadLetterPolicy` (consumer), mirroring the Java v5 `config` package and read cleanly via C++20 designated initializers.

### Smaller calls
- `blockIfQueueFull` defaults to `true`.
- Namespace-wide subscription is `inNamespace(...)` — `namespace` is a C++ keyword.
- `PulsarClient`'s default constructor is private — build only through the builder (matching the existing `Client`).
- Handle validity is just `explicit operator bool()` (dropped the redundant `valid()`).
- `receive()` returns `Expected<Message<T>>` because a receive can fail *without* a message (closed / disconnected / decode error).

## Open questions for reviewers

- **reflect-cpp dependency (deferred):** `jsonSchema<T>()` / `avroSchema<T>()` are built on reflect-cpp, but this API-only PR does **not** yet add it to `vcpkg.json` — the JSON sample is gated behind `find_package(reflectcpp)`, and reflect-cpp will be wired into the manifest with the `lib/st` implementation that actually exercises it. The reflect-cpp calls here are verified against a stand-in only. Open question: acceptable as a hard dep of the JSON/Avro schema headers, or should it be optional? (The `reflectcpp` vcpkg port 0.24.0 also ships no Avro backend yet, so `avroSchema<T>()` needs one before it can build; JSON works via reflect-cpp's core yyjson.)
- **Result codes:** the API reuses `pulsar::Result` (re-exported via `using enum`); a scoped `enum class st::ErrorCode` would read cleaner but needs a translation layer.
- **Naming:** `inNamespace`, `ackPolicy`, the policy field sets, etc. — open to bikeshedding.

## Status / verification

- All headers are self-contained and compile under **C++20 `-Wall -Werror`**; examples are syntax-checked; `ProtobufNativeSchema.h` compiles against real vcpkg protobuf; the reflect-cpp headers against a stand-in `rfl` API.
- **Not yet linkable** — there is no `lib/st`. Planned follow-ups: the `lib/st` implementation (proto commands → DAG-watch session → producer path → consumers → transactions/DLQ), the C API, and wiring the examples into the build.

### DIFF
--- a/include/pulsar/st/ProtobufNativeSchema.h
+++ b/include/pulsar/st/ProtobufNativeSchema.h
@@ -24,7 +24,6 @@
 #include <pulsar/st/Schema.h>
 
 #include <cstddef>
-#include <span>
 #include <string>
 #include <type_traits>
 
@@ -40,12 +39,10 @@ struct ProtobufNativeSerDe {
                   "protobufNativeSchema<T> requires T to be a generated protobuf Message");
     SchemaInfo info() const { return pulsar::createProtobufNativeSchema(T::descriptor()); }
     std::string encode(const T& value) const { return value.SerializeAsString(); }
-    Expected<T> decode(std::span<const char> data) const {
+    T decode(const char* data, std::size_t size) const {
         T message;
-        if (message.ParseFromArray(data.data(), static_cast<int>(data.size()))) {
-            return message;
-        }
-        return unexpected(pulsar::ResultInvalidMessage, "failed to parse protobuf message");
+        message.ParseFromArray(data, static_cast<int>(size));
+        return message;
     }
 };
 }  // namespace detail
--- a/include/pulsar/st/ProtobufNativeSchema.h
+++ b/include/pulsar/st/ProtobufNativeSchema.h
@@ -24,8 +24,10 @@
 #include <pulsar/st/Schema.h>
 
 #include <cstddef>
+#include <span>
 #include <string>
 #include <type_traits>
+#include <vector>
 
 namespace pulsar::st {
 
@@ -38,11 +40,16 @@ struct ProtobufNativeSerDe {
     static_assert(std::is_base_of_v<google::protobuf::Message, T>,
                   "protobufNativeSchema<T> requires T to be a generated protobuf Message");
     SchemaInfo info() const { return pulsar::createProtobufNativeSchema(T::descriptor()); }
-    std::string encode(const T& value) const { return value.SerializeAsString(); }
-    T decode(const char* data, std::size_t size) const {
+    Expected<void> encode(const T& value, std::vector<std::byte>& out) const {
+        out.resize(value.ByteSizeLong());
+        if (!value.SerializeToArray(out.data(), static_cast<int>(out.size())))
+            return unexpected(pulsar::ResultInvalidMessage, "failed to serialize protobuf message");
+        return {};
+    }
+    Expected<T> decode(std::span<const std::byte> data) const {
         T message;
-        message.ParseFromArray(data, static_cast<int>(size));
-        return message;
+        if (message.ParseFromArray(data.data(), static_cast<int>(data.size()))) return message;
+        return unexpected(pulsar::ResultInvalidMessage, "failed to parse protobuf message");
     }
 };
 }  // namespace detail
--- a/include/pulsar/st/ProtobufNativeSchema.h
+++ b/include/pulsar/st/ProtobufNativeSchema.h
@@ -1,0 +1,71 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#pragma once
+
+#include <google/protobuf/message.h>
+#include <pulsar/ProtobufNativeSchema.h>
+#include <pulsar/defines.h>
+#include <pulsar/st/Schema.h>
+
+#include <cstddef>
+#include <string>
+#include <type_traits>
+
+namespace pulsar::st {
+
+/// @cond INTERNAL
+/// Internal: the protobuf-backed SerDe used by protobufNativeSchema<T>(). Not part
+/// of the public API.
+namespace detail {
+template <typename T>
+struct ProtobufNativeSerDe {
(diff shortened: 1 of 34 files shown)


include/pulsar/st/ProtobufNativeSchema.h
```diff
@@ -0,0 +1,71 @@
...
+namespace detail {
+template <typename T>
+struct ProtobufNativeSerDe {
+    static_assert(std::is_base_of_v<google::protobuf::Message, T>,
+                  "protobufNativeSchema<T> requires T to be a generated protobuf Message");
+    SchemaInfo info() const { return pulsar::createProtobufNativeSchema(T::descriptor()); }
+    std::string encode(const T& value) const { return value.SerializeAsString(); }
+    T decode(const char* data, std::size_t size) const {
+        T message;
+        message.ParseFromArray(data, static_cast<int>(size));
```
BewareMyPower: This method returns a bool variable, which is false when the data is corrupted, could you return `std::optional<T>` or just throw an exception in this case?

BTW, it would be better to use `std::string_view` to replace the `(const char*, std::size_t)` pair.

merlimat: Good points. I've updated the interface to use `std::vector<std::byte>` as the buffer and `std::span<const std::byte>` as the view.

include/pulsar/st/Schema.h
```diff
@@ -0,0 +1,391 @@
...
+ *
+ * @tparam S the candidate SerDe type.
+ * @tparam T the value type the SerDe handles.
+ */
+// clang-format off
+template <typename S, typename T>
+concept SerDeFor = requires(const S& serde, const T& value, std::span<const std::byte> data,
+                            std::vector<std::byte>& out) {
+    { serde.info() } -> std::convertible_to<SchemaInfo>;
+    { serde.encode(value, out) } -> std::convertible_to<Expected<void>>;
```
BewareMyPower, line 90: Should we return `Expected<std::vector<std::byte>>` instead of passing a mutable reference to `std::vector<std::byte>>`?

BewareMyPower (approved):
Other code LGTM, it's able to merge once the comment is addressed.

include/pulsar/st/Schema.h
merlimat: The reason for passing the vector as an arg, is that you could potentially reuse the same vector multiple times. If we return it, we're always bound to allocate a new one each time.

PR #601, opened 2026-07-09
Scalable Topics: `pulsar::st` producer — per-segment fan-out, key routing, MessageId

merlimat:
### Motivation

Second implementation PR for the scalable-topics C++ SDK (`pulsar::st`), building on the protocol foundation merged in #600. It adds the **producer**: publishing to a scalable topic by fanning out to per-segment producers, driven by the DAG-watch layout and routing keys to segments exactly like the Java v5 client (`ScalableTopicProducer`), so the two clients place keys on the same segments.

Per PIP-468 a scalable topic's segments are ordinary persistent topics, so the produce data path reuses the classic `ProducerImpl` wholesale — one per active segment — and `lib/st` adds only the fan-out, routing, and id-mapping on top.

### Modifications

One commit per step, in dependency order:

1. **`segment://` topic domain** (classic client) — `TopicName` learns the segment domain (always v2; `<parent>/<hhhh-hhhh-id>`; persistent; `isSegment()`), and all public entry points reject segment topics; `ClientImpl::createSegmentProducerAsync()` is the `lib/st` bypass that may target one.

2. **st `MessageId`** — segment-qualified ids: `MessageIdImpl` (v4 id + segment id + position vectors) and the `MessageIdV5` big-endian wire layout, field-for-field compatible with the Java client for id interop. The producer fills only the v4 id + segment id.

3. **Broker-pin seam** — thread an optional `assignedBrokerUrl` from `createSegmentProducerAsync` through `ProducerImpl` to its initial `grabCnx`, so a per-segment producer connects straight to the DAG-provided owner broker instead of doing a fresh lookup. `HandlerBase::start()` delegates to a new `start(assignedBrokerUrl)` overload (`grabCnx()` already equals `grabCnx(nullopt)`, so other handlers are unaffected); every new parameter defaults to `nullopt`, leaving the public producer path unchanged.

4. **Producer seam** — `pulsar::st::ProducerImplBase` (the abstract impl the public `detail::ProducerCore` forwards to) + `ProducerCore.cc` (the out-of-line forwarders, required once a core is actually minted from a `PULSAR_PUBLIC` class).

5. **`StProducerImpl`** — the producer, a port of Java v5 `ScalableTopicProducer`. It owns one `DagWatchSession` + one `SegmentRouter` per topic; per active segment it creates a classic `pulsar::Producer` (pinned to the segment's broker) cached as its creation `Future` so concurrent cold senders share one attempt. `sendAsync` routes each message by key (round-robin when keyless) to a segment, publishes it, and wraps the returned classic id with the segment id into a `pulsar::st::MessageId`. Route failure is terminal; a sealed/terminated segment retries with backoff (≤10, `min(100·(n+1), 500)`ms) and re-routes onto the layout the DAG watch delivers. Layout changes retire producers for departed segments; exclusive access modes eagerly attach (and fail `create()` up front if the claim is refused), Shared attaches lazily. `lastSequenceId`/`flush`/`close` behave as documented.

   No per-segment dispatch chain is needed — `st::Future` fires its listeners in FIFO registration order, so same-segment sends dispatch in call order for free (Java needs the chain only because `CompletableFuture` gives no such guarantee); a `FutureTest` locks that in.

6. **Broker-free unit tests** — per-segment config isolation (a regression guard: `pulsar::ProducerConfiguration`'s copy constructor shares its impl, so each segment's config is built fresh, not copy-then-mutated), access-mode mapping by name (the st and classic enums differ in numeric order), segment-gone classification, and MessageId minting.

### What works after this PR

`client.newProducer<T>().topic("segment://…").create()` returns a live producer; `send`/`sendAsync` route by key to the owning segment, publish through a per-segment classic producer pinned to the DAG broker, and return a segment-qualified `MessageId`; `lastSequenceId`, `flush`, and `close` work; layout changes retire departed segments and (for exclusive modes) eagerly attach; a send whose segment was sealed transparently retries and re-routes.

### Next phases (follow-up PRs)

End-to-end validation against a scalable-topics broker (produce keyed + keyless, verify per-segment routing, segment ids, and message counts), then the Queue/Stream/Checkpoint consumers, the C API, and transactions. Entry-bucket batching (PIP-486) is deferred until a batcher-builder seam exists — today only the encryption-disables-batching branch is ported.

### Verifying this change

- **88 broker-free `pulsar-st-tests`**: `Expected`/`Future` (incl. the FIFO-listener ordering the producer relies on), the segment layout/router, and the producer's config isolation, access-mode mapping, segment-gone classification, and id minting.
- Full local build (vcpkg, macOS); **clang-format-11 clean**; **clang-tidy (LLVM 18 / libstdc++)** clean on the `lib/st` TUs and the changed classic files.
- The full produce/route/retry behavior gets its end-to-end validation against a 5.0.0-M1 broker in the follow-up.

### Documentation

- [ ] `doc` 
- [x] `doc-not-needed`

### DIFF
--- a/lib/st/StProducerImpl.cc
+++ b/lib/st/StProducerImpl.cc
@@ -1,0 +1,460 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "StProducerImpl.h"
+
+#include <pulsar/MessageBuilder.h>
+
+#include <algorithm>
+#include <chrono>
+#include <memory>
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "MessageIdImpl.h"
+#include "lib/ExecutorService.h"
+#include "lib/LogUtils.h"
+
+DECLARE_LOG_OBJECT()
+
+namespace pulsar::st {
+
+StProducerImpl::StProducerImpl(pulsar::ClientImplPtr classic, ProducerConfig config)
+    : classic_(std::move(classic)),
+      config_(std::move(config)),
+      topic_(config_.topic),
+      producerName_(config_.producerName.value_or(std::string{})),
+      currentLayout_(std::make_shared<SegmentLayout>()) {}
+
+Future<void> StProducerImpl::start() {
+    dagWatch_ = std::make_shared<DagWatchSession>(classic_, config_.topic, /*createIfMissing*/ true);
+    std::weak_ptr<StProducerImpl> weak = weak_from_this();
+    // Register the layout listener BEFORE start() so the very first layout is delivered to it.
+    dagWatch_->setLayoutChangeListener(
+        [weak](const SegmentLayout& newLayout, const SegmentLayout& oldLayout) {
+            if (auto self = weak.lock()) self->onLayoutChange(newLayout, oldLayout);
+        });
+    dagWatch_->start().addListener([weak](const Expected<SegmentLayout>& result) {
+        if (auto self = weak.lock()) self->onStartResult(result);
+    });
+    return startPromise_.getFuture();
+}
+
+void StProducerImpl::onStartResult(const Expected<SegmentLayout>& result) {
+    // Only the failure path is handled here: the DagWatchSession fails start()'s future (without
+    // invoking the layout listener) when the lookup fails before the first layout arrives. On
+    // success the listener runs immediately after and drives completeStart().
+    if (!result) {
+        startPromise_.setError(result.error());
+    }
+}
+
+void StProducerImpl::onLayoutChange(const SegmentLayout& newLayout, const SegmentLayout& /*oldLayout*/) {
+    std::vector<Future<pulsar::Producer>> retired;
+    bool first = false;
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        first = !sawFirstLayout_;
+        sawFirstLayout_ = true;
+        currentLayout_ = std::make_shared<SegmentLayout>(newLayout);
+
+        std::unordered_set<std::uint64_t> active;
+        for (const auto& segment : newLayout.activeSegments()) active.insert(segment.segmentId);
+        for (auto it = segmentProducers_.begin(); it != segmentProducers_.end();) {
+            if (active.find(it->first) == active.end()) {
+                retired.push_back(std::move(it->second));
+                it = segmentProducers_.erase(it);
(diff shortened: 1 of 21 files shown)


lib/st/StProducerImpl.cc
```diff
@@ -0,0 +1,460 @@
...
+        promise.setError(Error{ResultUnknownError,
+                               "segment " + std::to_string(segmentId) + " is not in the active layout"});
+        return promise.getFuture();
+    }
+
+    pulsar::ProducerConfiguration conf = buildSegmentConfiguration(*segment);
+    const std::string attachTopic = segment->attachTopicName();
+    std::optional<std::string> assignedBrokerUrl;
+    if (const std::string* url = currentLayout_->brokerUrl(segmentId)) {
+        assignedBrokerUrl = *url;  // pin to the DAG-provided owner broker
```
BewareMyPower: The url here is always plaintext URL, when could `brokerUrlTls` be used?

lib/st/StProducerImpl.cc
```diff
@@ -0,0 +1,460 @@
...
+        std::lock_guard<std::mutex> lock(mutex_);
+        segmentProducers_.erase(segmentId);
+    }
+    LOG_INFO("[" << topic_ << "] segment " << segmentId << " is gone; retrying send, attempt "
+                 << (attempt + 1) << " of " << kSendRetryMaxAttempts);
+    auto timer = classic_->getIOExecutorProvider()->get()->createDeadlineTimer();
+    const std::int64_t delayMs = std::min<std::int64_t>(100 * (attempt + 1), kSendRetryMaxBackoffMs);
+    timer->expires_from_now(std::chrono::milliseconds(delayMs));
+    auto self = shared_from_this();
+    timer->async_wait([self, message = std::move(message), attempt, userPromise,
```
BewareMyPower: Should we capture `weak_from_this()` since the timer is not cancelled on `closeAsync`?

lib/st/StProducerImpl.cc
merlimat: Good catch — it should not always be the plaintext URL. `brokerUrlTls` applies whenever the client itself connects over TLS, so the pin now makes the same choice the classic lookup does (`BinaryProtoLookupService`: `useTls() ? getBrokerUrlTls() : getBrokerUrl()`):

```cpp
const bool useTls = classic_->getServiceInfo().useTls();
const std::string* url =
    useTls ? currentLayout_->brokerUrlTls(segmentId) : currentLayout_->brokerUrl(segmentId);
```

If the DAG did not advertise the address we need, the pin is left unset and we fall back to the normal `segment://` lookup, which resolves it correctly. Fixed in 9e40bba.

lib/st/StProducerImpl.cc
merlimat: Yes — you are right. `closeAsync()` does not cancel these timers, so the strong `shared_from_this()` capture kept the producer alive until the backoff elapsed (up to 500ms) even after the last handle was dropped. Switched to a weak capture:

```cpp
std::weak_ptr<StProducerImpl> weak = weak_from_this();
timer->async_wait([weak, ...](const ASIO_ERROR& ec) mutable {
    auto self = weak.lock();
    if (ec || !self || self->closed_.load()) {
        userPromise.setError(Error{ResultAlreadyClosed, "producer closed during send retry"});
        return;
    }
    ...
});
```

Now the producer can be destroyed with a retry in flight, and the outliving timer fails the send with `ResultAlreadyClosed` rather than resurrecting it. (`timer` is still captured by value so it stays alive until it fires.) Fixed in 9e40bba.

lib/st/StProducerImpl.cc
merlimat: Yes, fixed. The broker returns both in the new scalable-topic lookup

lib/st/StProducerImpl.cc
merlimat: Good catch. Fixed

BewareMyPower (approved):

PR #605, opened 2026-07-20
Scalable Topics: `pulsar::st` queue consumer — per-segment fan-in over a mux receive queue

merlimat:
### Motivation

Next step in the `pulsar::st` (scalable-topics) SDK after the producer (#601) and its e2e tests (#603): the **queue consumer** for a single scalable topic.

A scalable topic is a DAG of per-segment backing topics. A queue consumer runs one `DagWatchSession` and a Shared-subscription classic `pulsar::Consumer` per segment — **active *and* sealed** (a split seals the parent without migrating its backlog, so its messages are only drainable through the sealed segment). Each segment's receive loop stamps the segment id onto every message and fans it into a shared, bounded **mux receive queue**; the user receives from that queue. Acks route back to the owning segment by the message id's segment id. Layout changes add consumers for new segments and close ones that leave the DAG.

Scope here is the **single-topic core** (mirrors the Java v5 `ScalableQueueConsumer`). Dead-letter and namespace-subscription modes are deferred to follow-ups.

### What's in this PR

- **Received-message plumbing** — `lib/st/MessageImpl` + `MessageCore` accessors (wrap a classic `Message` + the segment-qualified `MessageId`).
- **Classic consumer segment seam** — `ClientImpl::subscribeSegmentAsync` (a single-topic subscribe that accepts a `segment://` topic), guarded so the public path is unchanged.
- **Queue consumer core** — `lib/st/ReceiveQueue` (bounded fan-in mux with back-pressure) and `lib/st/QueueConsumerImpl` (per-segment consumers, ack/nack routing by segment id, layout reconcile, sealed-segment drain-and-close), wired through `QueueConsumerCore` and `StClientImpl::subscribeQueueAsync`.
- **Classic terminated-topic handling** (see below) — `ClientConnection` + `ConsumerImpl`.
- **End-to-end tests** — `tests/st/StQueueConsumerE2ETest.cc`: a produce→consume round-trip and a cross-segment fan-in test over a split topic, each self-contained (creates and splits its own topic via the admin REST API, same as the producer e2e).

### Note: a necessary classic-client change

The classic C++ client never handled `CommandReachedEndOfTopic` (`BaseCommand` type 27) — it fell through to `default:` and closed the connection as an "invalid message from server". A queue consumer that subscribes to sealed segments would therefore churn its connection (close → reconnect → re-subscribe → end-of-topic again) instead of learning the segment drained.

This PR handles it: dispatch the command to the target consumer (mirroring `handleActiveConsumerChange`), and have `ConsumerImpl` surface `ResultTopicTerminated` on the **async receive** path once the prefetch queue drains — matching the Java client, whose consumers close a drained sealed segment on `TopicTerminated`. The broker only sends the command once the consumer's read position reaches the terminate marker, so buffered messages always drain first. The change is additive (a flag that is only ever set on a terminated topic), and it's covered by a new `ConsumerTest.testReceiveAsyncAfterTopicTerminated`.

### Testing

- `pulsar-st-tests`: **95/95** green (88 broker-free unit + 5 producer e2e + 2 consumer e2e) against a `5.0.0-M1` standalone.
- Classic `ConsumerTest` **34/34** and `ProducerTest` terminated tests **2/2** green (no regression from the terminated-topic change).
- clang-format-11 and clang-tidy (analyzer + performance) clean on all changed files.

### Follow-ups

- The e2e admin-REST helpers are currently duplicated between `StProducerE2ETest.cc` (landed in #603) and `StQueueConsumerE2ETest.cc`; a small follow-up can extract a shared `tests/st/StE2EAdmin.h`.
- Stream and checkpoint consumers, the C API, and transactions.

### DIFF
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -823,6 +823,24 @@ void ConsumerImpl::activeConsumerChanged(bool isActive) {
     }
 }
 
+void ConsumerImpl::reachedEndOfTopic() {
+    hasReachedEndOfTopic_ = true;
+    // If nothing is buffered there is nothing left to deliver, so complete any waiting async
+    // receives with ResultTopicTerminated now. When messages are still buffered they drain through
+    // the normal path first, and the next receive observes the flag (see receiveAsync).
+    Lock lock(pendingReceiveMutex_);
+    if (incomingMessages_.empty()) {
+        Message msg;
+        while (!pendingReceives_.empty()) {
+            ReceiveCallback callback = pendingReceives_.front();
+            pendingReceives_.pop();
+            listenerExecutor_->postWork(std::bind(&ConsumerImpl::notifyPendingReceivedCallback,
+                                                  get_shared_this_ptr(), ResultTopicTerminated, msg,
+                                                  callback));
+        }
+    }
+}
+
 void ConsumerImpl::internalConsumerChangeListener(bool isActive) {
     try {
         if (isActive) {
@@ -1189,6 +1207,14 @@ void ConsumerImpl::receiveAsync(const ReceiveCallback& callback) {
         messageProcessed(msg);
         msg = interceptors_->beforeConsume(Consumer(shared_from_this()), msg);
         callback(ResultOk, msg);
+    } else if (hasReachedEndOfTopic_) {
+        // Terminated topic with nothing left buffered: fail the receive rather than parking it
+        // forever waiting for a message that will never arrive.
+        pendingReceiveMutexLock.unlock();
+        if (config_.getReceiverQueueSize() == 0) {
+            mutexlock.unlock();
+        }
+        callback(ResultTopicTerminated, msg);
     } else if (config_.getReceiverQueueSize() == 0) {
         pendingReceives_.push(callback);
         // If connection_ is nullptr, sendFlowPermitsToBroker does nothing.
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -1536,7 +1536,16 @@ std::pair<MessageId, bool> ConsumerImpl::prepareCumulativeAck(const MessageId& m
 
 void ConsumerImpl::negativeAcknowledge(const MessageId& messageId) {
     unAckedMessageTrackerPtr_->remove(messageId);
-    negativeAcksTracker_->add(messageId);
+    // If it's a ChunkMessageId, expand all chunk entries and nack them individually,
+    // so that the broker can redeliver all chunks
+    if (auto chunkMessageId =
+            std::dynamic_pointer_cast<ChunkMessageIdImpl>(Commands::getMessageIdImpl(messageId))) {
+        for (const auto& chunkId : chunkMessageId->getChunkedMessageIds()) {
+            negativeAcksTracker_->add(chunkId);
+        }
+    } else {
+        negativeAcksTracker_->add(messageId);
+    }
 }
 
 void ConsumerImpl::disconnectConsumer() { disconnectConsumer(std::nullopt); }
@@ -1697,7 +1706,19 @@ void ConsumerImpl::redeliverMessages(const std::set<MessageId>& messageIds) {
     ClientConnectionPtr cnx = getCnx().lock();
     if (cnx) {
         if (cnx->getServerProtocolVersion() >= proto::v2) {
-            cnx->sendCommand(Commands::newRedeliverUnacknowledgedMessages(consumerId_, messageIds));
+            // Expand ChunkMessageIds into all chunk entries to ensure the broker
+            // can redeliver all chunks
+            std::set<MessageId> expandedMsgIds;
+            for (const auto& msgId : messageIds) {
+                if (auto chunkMsgId =
+                        std::dynamic_pointer_cast<ChunkMessageIdImpl>(Commands::getMessageIdImpl(msgId))) {
+                    const auto& chunkIds = chunkMsgId->getChunkedMessageIds();
+                    expandedMsgIds.insert(chunkIds.begin(), chunkIds.end());
+                } else {
+                    expandedMsgIds.insert(msgId);
+                }
+            }
+            cnx->sendCommand(Commands::newRedeliverUnacknowledgedMessages(consumerId_, expandedMsgIds));
             LOG_DEBUG("Sending RedeliverUnacknowledgedMessages command for Consumer - " << getConsumerId());
         }
     } else {
@@ -2083,6 +2104,7 @@ void ConsumerImpl::processPossibleToDLQ(const MessageId& messageId, const Proces
             producerConfiguration.setSchema(config_.getSchema());
             producerConfiguration.setBlockIfQueueFull(false);
             producerConfiguration.setBatchingEnabled(false);
+            producerConfiguration.setChunkingEnabled(true);
             producerConfiguration.impl_->initialSubscriptionName =
                 deadLetterPolicy_.getInitialSubscriptionName();
             ClientImplPtr client = client_.lock();
--- a/lib/ConsumerImpl.cc
+++ b/lib/ConsumerImpl.cc
@@ -337,6 +337,11 @@ Result ConsumerImpl::handleCreateConsumer(const ClientConnectionPtr& cnx, Result
             incomingMessages_.clear();
             possibleSendToDeadLetterTopicMessages_.clear();
             backoff_.reset();
+            // Re-derive end-of-topic from the new session: termination stops new publications, not
+            // redelivery of unacked messages, so a stale flag would report ResultTopicTerminated in
+            // the window before redeliveries arrive. The broker re-sends CommandReachedEndOfTopic
+            // once this consumer's read position reaches the terminate marker again.
+            hasReachedEndOfTopic_ = false;
             if (!messageListener_ && config_.getReceiverQueueSize() == 0) {
                 // Complicated logic since we don't have a isLocked() function for mutex
                 if (waitingForZeroQueueSizeMessage) {
@@ -1243,6 +1248,13 @@ Result ConsumerImpl::receiveHelper(Message& msg) {
         return fetchSingleMessageFromBroker(msg);
     }
 
+    // A drained terminated topic has nothing left to deliver: fail fast instead of blocking
+    // forever, matching the async path. (A receive already parked in pop() when end-of-topic
+    // arrives still waits — the queue only wakes on a message or on close.)
+    if (hasReachedEndOfTopic_ && incomingMessages_.empty()) {
+        return ResultTopicTerminated;
+    }
+
     if (!incomingMessages_.pop(msg)) {
         return ResultInterrupted;
     }
@@ -1273,6 +1285,10 @@ Result ConsumerImpl::receiveHelper(Message& msg, int timeout) {
         return ResultInvalidConfiguration;
     }
 
+    if (hasReachedEndOfTopic_ && incomingMessages_.empty()) {
+        return ResultTopicTerminated;
+    }
+
     if (incomingMessages_.pop(msg, std::chrono::milliseconds(timeout))) {
         messageProcessed(msg);
         msg = interceptors_->beforeConsume(Consumer(shared_from_this()), msg);
@@ -1280,6 +1296,10 @@ Result ConsumerImpl::receiveHelper(Message& msg, int timeout) {
     } else {
         if (state_ != Ready) {
             return ResultAlreadyClosed;
+        }
+        // Waking up empty on a terminated topic means drained, not merely idle.
+        if (hasReachedEndOfTopic_ && incomingMessages_.empty()) {
+            return ResultTopicTerminated;
         }
         return ResultTimeout;
     }
--- a/lib/st/QueueConsumerImpl.cc
+++ b/lib/st/QueueConsumerImpl.cc
@@ -1,0 +1,303 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "QueueConsumerImpl.h"
+
+#include <cstdint>
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "MessageIdImpl.h"
+#include "MessageImpl.h"
+#include "lib/LogUtils.h"
+
+DECLARE_LOG_OBJECT()
+
+namespace pulsar::st {
+
+namespace {
+
+pulsar::InitialPosition toClassicInitialPosition(SubscriptionInitialPosition position) {
+    return position == SubscriptionInitialPosition::Earliest ? pulsar::InitialPositionEarliest
+                                                             : pulsar::InitialPositionLatest;
+}
+
+}  // namespace
+
+QueueConsumerImpl::QueueConsumerImpl(pulsar::ClientImplPtr classic, QueueConsumerConfig config)
+    : classic_(std::move(classic)),
+      config_(std::move(config)),
+      topic_(config_.topic),
+      subscription_(config_.subscriptionName),
+      consumerName_(config_.consumerName.value_or(std::string{})),
+      receiveQueue_(
+          std::make_shared<ReceiveQueue>(classic_->getIOExecutorProvider()->get(), kReceiveQueueCapacity)),
+      currentLayout_(std::make_shared<SegmentLayout>()) {}
+
+Future<void> QueueConsumerImpl::start() {
+    dagWatch_ = std::make_shared<DagWatchSession>(classic_, config_.topic, /*createIfMissing*/ true);
+    std::weak_ptr<QueueConsumerImpl> weak = weak_from_this();
+    dagWatch_->setLayoutChangeListener(
+        [weak](const SegmentLayout& newLayout, const SegmentLayout& oldLayout) {
+            if (auto self = weak.lock()) self->onLayoutChange(newLayout, oldLayout);
+        });
+    dagWatch_->start().addListener([weak](const Expected<SegmentLayout>& result) {
+        if (auto self = weak.lock()) self->onStartResult(result);
+    });
+    return startPromise_.getFuture();
+}
+
+void QueueConsumerImpl::onStartResult(const Expected<SegmentLayout>& result) {
+    // Only the failure path (see StProducerImpl::onStartResult): the layout listener drives the
+    // success path — subscribe the initial segments and complete startPromise_.
+    if (!result) startPromise_.setError(result.error());
+}
+
+void QueueConsumerImpl::onLayoutChange(const SegmentLayout& newLayout, const SegmentLayout& /*oldLayout*/) {
+    // Subscribe active AND sealed segments: a sealed segment may still hold undrained messages.
+    std::vector<Segment> target;
+    target.reserve(newLayout.activeSegments().size() + newLayout.sealedSegments().size());
+    for (const auto& segment : newLayout.activeSegments()) target.push_back(segment);
+    for (const auto& segment : newLayout.sealedSegments()) target.push_back(segment);
+
+    std::vector<Future<pulsar::Consumer>> retired;
+    std::vector<Segment> toAdd;
+    bool first = false;
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        first = !sawFirstLayout_;
+        sawFirstLayout_ = true;
+        currentLayout_ = std::make_shared<SegmentLayout>(newLayout);
+
+        std::unordered_set<std::uint64_t> targetIds;
+        for (const auto& segment : target) targetIds.insert(segment.segmentId);
+        for (auto it = segmentConsumers_.begin(); it != segmentConsumers_.end();) {
+            if (targetIds.find(it->first) == targetIds.end()) {
+                retired.push_back(std::move(it->second));
+                it = segmentConsumers_.erase(it);
+            } else {
+                ++it;
+            }
+        }
+        for (const auto& segment : target) {
+            if (segmentConsumers_.find(segment.segmentId) == segmentConsumers_.end())
+                toAdd.push_back(segment);
+        }
+    }
+
+    for (auto& future : retired) {
+        future.addListener([](const Expected<pulsar::Consumer>& result) {
+            if (result) {
+                pulsar::Consumer consumer = *result;
+                consumer.closeAsync([](pulsar::Result) {});
+            }
+        });
+    }
+
+    if (first) {
+        if (toAdd.empty()) {
+            startPromise_.setSuccess();
+            return;
+        }
+        auto remaining = std::make_shared<std::atomic<int>>(static_cast<int>(toAdd.size()));
+        for (const auto& segment : toAdd) {
+            getOrCreateSegmentConsumerAsync(segment).addListener(
+                [self = shared_from_this(), remaining](const Expected<pulsar::Consumer>& result) {
+                    if (!result) {
+                        self->startPromise_.setError(result.error());  // first error wins (idempotent)
+                        return;
+                    }
+                    if (remaining->fetch_sub(1) == 1) self->startPromise_.setSuccess();
+                });
+        }
+    } else {
+        for (const auto& segment : toAdd) getOrCreateSegmentConsumerAsync(segment);  // best-effort
+    }
+}
+
+pulsar::ConsumerConfiguration QueueConsumerImpl::buildSegmentConfiguration(const Segment& segment) const {
+    // Build a FRESH config every time (pulsar::ConsumerConfiguration's copy ctor shares its impl).
+    pulsar::ConsumerConfiguration conf;
+    conf.setConsumerType(pulsar::ConsumerShared);
+    conf.setSchema(config_.schema);
+    conf.setSubscriptionInitialPosition(toClassicInitialPosition(config_.initialPosition));
+    if (config_.consumerName) {
+        conf.setConsumerName(*config_.consumerName + "-seg-" + std::to_string(segment.segmentId));
+    }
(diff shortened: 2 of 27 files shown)


lib/st/QueueConsumerImpl.cc
```diff
@@ -0,0 +1,314 @@
...
+void QueueConsumerImpl::startReceiveLoop(pulsar::Consumer consumer, std::uint64_t segmentId) {
+    if (closed_.load()) return;
+    auto self = shared_from_this();
+    consumer.receiveAsync([self, consumer, segmentId](pulsar::Result result, const pulsar::Message& message) {
+        if (result != pulsar::ResultOk) {
+            if (result == pulsar::ResultTopicTerminated) {
+                // A sealed segment fully drained: close its consumer, drop it from the cache, and
+                // remember it drained so a later reconcile does not re-subscribe the still-in-DAG
+                // sealed segment (which would redeliver its unacked messages as duplicates).
+                {
```
lhotari, line 272: **[BUG] Closing a drained sealed segment loses the acks for messages still in the mux queue or in the application's hands**

`ResultTopicTerminated` only tells you the *classic* consumer's prefetch queue is empty. It says nothing about the messages this loop already pulled out of it: up to `kReceiveQueueCapacity` (1000) of them may still be sitting in `receiveQueue_`, plus any already handed to the application and not yet acked.

Erasing `segmentConsumers_[segmentId]` here removes the only ack-routing handle for those messages, and `drainedSegments_.insert(segmentId)` guarantees the segment is never reattached for the life of this consumer. `acknowledge()` then resolves `segmentConsumerFor(id)` to an errored future at line 244, and the listener at lines 252-257 has no error branch — the ack is silently discarded.

Failure scenario: a sealed segment holds 500 undrained messages. The receive loop pulls all 500 into the mux queue and the broker reports end-of-topic; this handler closes the consumer. The application then receives and acks all 500 from the mux queue. Every one of those acks is dropped. The durable cursor still holds them, so they are redelivered in full to the next consumer on this subscription, and the sealed segment never becomes eligible for removal.

This is the half of your earlier thread with Copilot that `drainedSegments_` did not close — it fixed the duplicate-redelivery symptom, but the acks are still lost. A refcount of outstanding messages per segment, with the close deferred until it reaches zero, would cover both.

lib/st/QueueConsumerImpl.cc
```diff
@@ -0,0 +1,314 @@
...
+                pulsar::Consumer done = consumer;
+                done.closeAsync([](pulsar::Result) {});
+            }
+            // Otherwise (AlreadyClosed / consumer closing) just stop the loop.
+            return;
+        }
+        MessageId id = MessageIdFactory::create(message.getMessageId(), static_cast<std::int64_t>(segmentId));
+        auto messageImpl = std::make_shared<MessageImpl>(message, std::move(id));
+        // Re-arm only once the fan-in queue has room, so a slow consumer throttles this segment.
+        self->receiveQueue_->offer(std::move(messageImpl))
```
lhotari, line 295: **[BUG] The segment receive loop recurses synchronously per message and can exhaust the stack**

This re-arm is not a loop — it is recursion, and nothing guarantees it unwinds:

- `Consumer::receiveAsync` (`lib/Consumer.cc:82`) is a thin forwarder, and `ConsumerImpl::receiveAsync` (`lib/ConsumerImpl.cc:1201-1207`) invokes `callback(ResultOk, msg)` **inline** when `incomingMessages_.pop()` succeeds.
- `offer()` returns an **already-completed** future whenever the queue has room (`ReceiveQueue.cc:134-137`).
- `SharedState::addListener` (`include/pulsar/st/detail/SharedState.h:58-67`) runs the listener **synchronously** when the result is already available.

So the listener calls `startReceiveLoop` on the same stack, which calls `receiveAsync`, which calls back inline, and so on — roughly 8 frames per message.

The counter-intuitive part: an application that is actively receiving makes this *worse*. When a receive is already parked, `offer()` hands the message straight to the receiver (`ReceiveQueue.cc:118-122`) without touching `buffer_`, so the buffer never reaches capacity and the back-pressure brake never engages. The chain then breaks only when the classic prefetch queue runs dry — and because each dequeue replenishes flow permits, broker I/O refills it concurrently from another thread while this one is still recursing.

Failure scenario: a consumer attaches to a segment with a large backlog and drains it promptly. The default `receiverQueueSize` is 1000 (nothing here overrides it), so even without concurrent replenishment the chain reaches ~1000 nested iterations — several thousand frames — on a listener-executor thread whose stack is well under a megabyte. Under sustained throughput there is no bound at all.

Posting the re-arm to the executor instead of continuing inline would make it a loop again.

lib/ConsumerImpl.cc
```diff
@@ -823,6 +823,24 @@ void ConsumerImpl::activeConsumerChanged(bool isActive) {
     }
 }
 
+void ConsumerImpl::reachedEndOfTopic() {
+    hasReachedEndOfTopic_ = true;
```
lhotari, line 832: **[BUG] hasReachedEndOfTopic_ is never cleared, so a reconnect can report a stale terminal result and permanently drop a segment**

This flag is set once and never reset — it is the only write in the file. Meanwhile the reconnect path clears the prefetch queue (`clearReceiveQueue()` at line 244, `incomingMessages_.clear()` at line 337) and then sends fresh flow permits.

That leaves a window after every reconnect in which `incomingMessages_` is empty but redelivery of unacked messages has not yet arrived. A `receiveAsync` landing in that window falls through to the new branch at line 1210 and returns `ResultTopicTerminated` even though messages are still owed.

Termination means no *new* publications; it does not mean already-delivered-but-unacked entries stop needing redelivery.

Failure scenario: a segment consumer with unacked messages loses its connection and re-subscribes. Before the broker redelivers, the st queue consumer's receive loop calls `receiveAsync`, gets the stale `ResultTopicTerminated`, and takes the drain path in `QueueConsumerImpl.cc:203-213` — closing the consumer and recording the segment in `drainedSegments_`. Those messages are never delivered to the application for the life of this consumer.

Clearing the flag in `connectionOpened()`/`handleCreateConsumer()` alongside the queue clear would close this.

lib/ConsumerImpl.cc
```diff
@@ -1189,6 +1207,14 @@ void ConsumerImpl::receiveAsync(const ReceiveCallback& callback) {
         messageProcessed(msg);
         msg = interceptors_->beforeConsume(Consumer(shared_from_this()), msg);
         callback(ResultOk, msg);
+    } else if (hasReachedEndOfTopic_) {
+        // Terminated topic with nothing left buffered: fail the receive rather than parking it
+        // forever waiting for a message that will never arrive.
```
lhotari, line 1217: **[BUG] Blocking receive() still waits forever on a drained terminated topic**

The new termination check exists only on the async path. The untimed `receiveHelper(Message&)` (line 1246) blocks in `incomingMessages_.pop(msg)` and never consults `hasReachedEndOfTopic_`, and `reachedEndOfTopic()` only completes `pendingReceives_`, which synchronous receives do not use. The timed overload returns `ResultTimeout` rather than `ResultTopicTerminated`.

Failure scenario: an application calls `consumer.receive(msg)` on a topic that is terminated and fully drained. It blocks forever. Before this PR the same call also hung, but the connection churn made it visible; now the client handles the command cleanly and the hang is silent.

Not a regression, but the capability is only half-wired — worth finishing here so the sync and async paths agree.

lib/st/QueueConsumerImpl.cc
```diff
@@ -0,0 +1,314 @@
...
+    classic_->subscribeSegmentAsync(
+        attachTopic, config_.subscriptionName, conf,
+        [self, promise, segmentId](std::variant<pulsar::Error, pulsar::Consumer> result) {
+            if (auto* consumer = std::get_if<pulsar::Consumer>(&result)) {
+                // pulsar::Consumer is a copyable handle (its virtual dtor suppresses the move ctor),
+                // so this is a shared-impl copy, not a deep copy.
+                pulsar::Consumer c = *consumer;
+                self->startReceiveLoop(c, segmentId);
+                promise.setValue(c);
+            } else {
```
lhotari, line 205: **[BUG] A transient segment subscribe failure can leave an active segment unconsumed indefinitely**

On failure the segment is evicted from `segmentConsumers_` so a later reconcile can retry it — but off the first-layout path (line 138) nothing schedules that retry. `onLayoutChange` runs only when the broker pushes a DAG update.

A reconnect does re-push the layout, so this self-heals in that case. What it does not survive is the common one: a healthy watch, a stable DAG, and a single segment whose subscribe failed on a transient lookup or connection error.

Failure scenario: one of four active segments fails to subscribe during a layout change. The other three attach and the consumer looks healthy. The DAG does not change again for hours. Everything published to that fourth segment is never delivered, with nothing in the API surfacing the gap.

A bounded retry with backoff for evicted segments would cover it.

lib/st/QueueConsumerImpl.cc
```diff
@@ -0,0 +1,314 @@
...
+                    self->drainedSegments_.insert(segmentId);
+                }
+                pulsar::Consumer done = consumer;
+                done.closeAsync([](pulsar::Result) {});
+            }
+            // Otherwise (AlreadyClosed / consumer closing) just stop the loop.
+            return;
+        }
+        MessageId id = MessageIdFactory::create(message.getMessageId(), static_cast<std::int64_t>(segmentId));
+        auto messageImpl = std::make_shared<MessageImpl>(message, std::move(id));
```
lhotari: **[INTENT MISMATCH] Message::topic() reports the internal segment:// backing topic instead of the scalable topic**

`MessageImpl` is constructed without `topicOverride`, so `topic()` falls through to `classic_.getTopicName()` — the `segment://...` backing topic.

The public contract (`include/pulsar/st/Message.h:135-139`) promises "the resolved canonical topic the message was received from" and "the fully-qualified topic name". Segment topics are internal enough that the public subscribe path rejects them outright, so surfacing one back to the application inverts that boundary — and the value shifts under the application whenever a split reassigns segments.

`config_.topic` is right there in the impl; passing it as the override would keep the abstraction intact.

lib/st/QueueConsumerImpl.cc
```diff
@@ -0,0 +1,314 @@
...
+                    }
+                    if (remaining->fetch_sub(1) == 1) self->startPromise_.setSuccess();
+                });
+        }
+    } else {
+        for (const auto& segment : toAdd) getOrCreateSegmentConsumerAsync(segment);  // best-effort
+    }
+}
+
+pulsar::ConsumerConfiguration QueueConsumerImpl::buildSegmentConfiguration(const Segment& segment) const {
```
lhotari, line 161: **[QUALITY] A configured dead-letter policy is silently ignored**

`QueueConsumerBuilder::deadLetterPolicy()` (`include/pulsar/st/QueueConsumer.h:335`) stores the policy into `QueueConsumerConfig`, and nothing in `lib/st` ever reads it — this builder never calls `setDeadLetterPolicy()` on the per-segment configuration.

I realise dead-lettering is explicitly deferred to a follow-up, and that is a reasonable scope call. The problem is only that the builder documents it as working ("Route messages to a dead-letter topic after repeated redelivery (spec §7.2). QueueConsumer only.") with no caveat, so an application that sets it gets a silent no-op rather than an error.

Failing the subscribe when the policy is set, or at minimum logging a warning the way `acknowledge(id, txn)` already does at line 260, would keep the deferral honest until the follow-up lands.

tests/st/StQueueConsumerE2ETest.cc
```diff
@@ -0,0 +1,203 @@
...
+// Consume from a topic split (via REST) into two active segments and assert the messages actually
+// arrive from both — this is the fan-in the queue consumer exists for: one Shared subscription
+// multiplexed across a per-segment classic consumer each, drained through the mux receive queue.
+// The single-segment round-trip above never exercises multi-segment fan-in.
+TEST(StQueueConsumerE2ETest, testConsumeAcrossSplitSegments) {
+    if (!e2eEnabled()) GTEST_SKIP() << "set PULSAR_ST_E2E=1 to run against a scalable-topics broker";
+
+    const std::string name = uniqueName("st-e2e-queue-split");
+    ASSERT_TRUE(createScalableTopic(name)) << "failed to create scalable topic " << name;
+    ASSERT_TRUE(splitSegment(name, 0)) << "failed to split segment 0 of " << name;
```
lhotari, line 156: **[QUALITY] Neither e2e test drains a sealed segment — the PR's headline scenario is untested**

The split happens here, immediately after `createScalableTopic` and before the producer exists. Segment 0 is therefore sealed while empty, and every message goes to the two active children.

So this test covers cross-segment fan-in, but never the sealed-segment drain — no `ResultTopicTerminated` is ever delivered, and the `drainedSegments_` logic, the close, and the ack routing for a sealed segment all go unexercised. That is the case the PR description leads with: "a split seals the parent without migrating its backlog, so its messages are only drainable through the sealed segment."

A test that produces first, then splits, then consumes — asserting every pre-split message still arrives and that its acks stick (e.g. by reattaching a second consumer on the same subscription and getting nothing) — would cover the drain path, and I suspect would surface the ack-loss issue above directly.

lib/st/ReceiveQueue.cc
```diff
@@ -0,0 +1,157 @@
...
+            parked = true;
+        }
+    }
+    for (auto& waiter : toSignal) waiter.setSuccess();
+    if (message) {
+        promise.setValue(std::move(message));
+        return promise.getFuture();
+    }
+    if (parked) {
+        auto timer = executor_->createDeadlineTimer();
```
lhotari, line 93: **[QUALITY] A timed receive never cancels its timer, holding the queue and promise alive until the deadline**

When a message arrives first, the timer is left to fire and no-op — but until it does it keeps `self` (the whole `ReceiveQueue`), the promise and the timer object alive.

Under a steady stream of `receiveAsync(timeout)` calls with a generous timeout, live timers accumulate in proportion to receive rate × timeout rather than to the number of outstanding receives.

Copilot raised this earlier and the thread was resolved, but the follow-up commit touched only `QueueConsumerImpl.{cc,h}`, so the code is unchanged. Flagging because I think the point stands — cancelling the timer on the delivery path would settle it.

lhotari (changes requested):
The fan-in design here is the right shape: `offer()` gating the segment re-arm gives real back-pressure rather than an unbounded buffer, and subscribing sealed segments as well as active ones is exactly the subtlety a split introduces. The `subscribeSegmentAsync` seam is properly confined too — `allowSegmentTopic` defaults to `false`, and only the st queue consumer passes `true`, so the public subscribe path is genuinely unchanged.

My concerns cluster on the sealed-segment drain path, which is both the most intricate new logic and the least exercised by the tests.

The drain path closes a segment's consumer as soon as the broker reports end-of-topic, but end-of-topic only means the *classic* prefetch queue drained — up to 1000 messages from that segment may still be sitting in the mux queue or in the application's hands. Their acks are then silently dropped, and `drainedSegments_` guarantees the segment is never reattached. This is the half of the earlier Copilot thread that the `drainedSegments_` fix did not cover: it cures the duplicate redelivery, but the acks are still lost.

Separately, the segment receive loop recurses rather than iterates. `ConsumerImpl::receiveAsync` calls back inline when a message is already buffered, `offer()` returns an already-completed future whenever the queue has room, and `SharedState::addListener` runs listeners inline — so each message adds stack frames instead of unwinding. An actively-receiving application makes this worse, not better, because `offer()` then hands messages straight to the parked receiver and the buffer never fills to apply the brake.

On testing: neither e2e test ever drains a sealed segment. `testConsumeAcrossSplitSegments` splits before producing, so the sealed parent is empty — the scenario the PR description leads with ("a split seals the parent without migrating its backlog") has no coverage. That is where I would put the next test, and it would likely surface the first finding directly.

lib/st/QueueConsumerImpl.cc
merlimat: Fixed with the outstanding-message refcount you suggested: every message the loop fans in counts until the application acks or nacks it, and end-of-topic now defers the close until that count reaches zero — the consumer stays in `segmentConsumers_` meanwhile, so late acks still route. The new `testDrainSealedSegmentBacklog` e2e drives exactly your failure scenario (1200-message backlog sealed behind a split, drained and acked through the sealed segment) and proves the acks stick by reattaching a second consumer on the same subscription and receiving nothing.

lib/st/QueueConsumerImpl.cc
merlimat: Fixed — the re-arm now hops through the IO executor (`postWork`), so each message unwinds the stack before the next receive is armed, on both inline-completion paths (prefetched message and already-complete `offer()` future). The drain e2e pushes 1200 messages — more than both the classic prefetch queue and the mux capacity — through a single segment loop.

lib/ConsumerImpl.cc
merlimat: Fixed — the flag is cleared in `handleCreateConsumer` alongside `incomingMessages_.clear()`, before flow permits are sent, so a receive in the redelivery window parks normally. The broker re-sends `CommandReachedEndOfTopic` once the re-created consumer's read position reaches the terminate marker again.

lib/ConsumerImpl.cc
merlimat: Both sync paths now agree with the async path: the untimed `receive()` fails fast with `ResultTopicTerminated` when the flag is set and the queue is empty, and the timed overload returns it both up front and when it wakes empty (instead of `ResultTimeout`). One limitation, noted in a comment: a receive already parked in `pop()` when the command arrives still waits — waking it would need an interruptible queue, which felt out of scope here. Covered by the extended `ConsumerTest.testReceiveAsyncAfterTopicTerminated` (both sync overloads asserted).

lib/st/QueueConsumerImpl.cc
merlimat: Added `subscribeSegmentWithRetry`: off the first-layout path (where no error can surface to a caller), a failed segment subscribe retries up to 10 times with 100→500 ms backoff — same constants as the producer's send retry — skipping segments that have left the DAG or drained, and giving up to the next DAG push after the bound. The first-layout path still fails `subscribe()` so the application sees the error.

lib/st/QueueConsumerImpl.cc
merlimat: Fixed — the scalable topic is passed as `topicOverride` when wrapping the message, so `topic()` reports the canonical topic the application subscribed to.

lib/st/QueueConsumerImpl.cc
merlimat: Went with failing the subscribe: a configured `deadLetterPolicy` now fails with `ResultOperationNotSupported` (subscribe has a real error channel, so loud beats a log line), and both API doc sites carry the "not implemented yet" caveat until the follow-up lands.

tests/st/StQueueConsumerE2ETest.cc
merlimat: Added `testDrainSealedSegmentBacklog`: produce first (1200 messages), then split, then consume — asserting every pre-split message arrives from the sealed parent (segment id 0), and that the acks stick (a second consumer on the same subscription receives nothing within 3 s; without the refcount fix it receives the whole backlog back). You were right that this test surfaces the ack-loss finding directly — it fails on the previous code.

lib/st/ReceiveQueue.cc
merlimat: Fixed this time — a parked receive is now a `{promise, timer}` entry, and both the delivery path and `close()` cancel the timer, so live timers track outstanding receives rather than receive rate × timeout. (You're right that the earlier resolution only addressed the QueueConsumerImpl half of the thread.)

lhotari (approved):
Thanks — all nine points are addressed, and several of the fixes are better than what I asked for.

**The drain path** (`QueueConsumerImpl.cc:262-290`, `308-333`) is now correct. The `outstanding_` refcount is incremented in `onMessageFannedIn` *before* the message is offered to the mux queue and decremented only when the application settles it, so an end-of-topic arriving with messages still in flight parks the segment in `terminatedSegments_` and keeps its consumer in `segmentConsumers_` for ack routing; the close runs from `onMessageSettled` once the count reaches zero. Settling after the ack is enqueued rather than before is the right ordering, and `takeDrainedSegmentLocked` handing the future back to be closed outside the lock is a nice touch.

**The receive loop** (`QueueConsumerImpl.cc:294-297`) now hops through `executor_->postWork`, so it is a loop rather than recursion. That was the crash-class one.

**The terminated-topic flag** is cleared on each new broker session (`ConsumerImpl.cc:344`), right alongside `incomingMessages_.clear()` — exactly where it needed to go, and the comment explains why termination does not cancel redelivery.

**Sync receive** now checks the flag in both `receiveHelper` overloads plus after a timeout wake-up (`ConsumerImpl.cc:1254`, `1288`, `1301`), and `ConsumerTest` asserts both sync paths agree with the async one.

Also fixed: the bounded subscribe retry with backoff and a still-wanted check (`subscribeSegmentWithRetry`), `topic()` now reporting the scalable topic, dead-lettering failing loudly with `ResultOperationNotSupported` instead of silently ignoring the policy, and the timed-receive timer now cancelled on delivery and on close.

**`testDrainSealedSegmentBacklog` is the test I was hoping for.** Producing before the split so the backlog is stranded in the sealed parent, 1200 messages to overrun both the prefetch and mux queues, asserting every message carries segment id 0, and then reattaching a second consumer to prove the acks stuck — that last assertion is what makes it a real regression test for the ack-loss bug rather than a happy-path walk.

Two non-blocking notes left in the threads rather than held against the PR: the per-call (rather than per-message) decrement in `onMessageSettled`, and the parked-`pop()` residual you documented in `receiveHelper`. Both are follow-up material at most.

I also checked whether closing a terminated segment on nack strands the redelivery, and I do not think it does: the receive loop has already returned by then, so an attached consumer would not deliver it either — the broker releasing it back to the subscription is the better outcome. Your inline comment gets this right.

One merge-readiness item that is not about your code: the **PR validation workflow hit a `startup_failure` on 60c763b** (run 33213225434, "This run likely failed because of a workflow file issue"), so only CodeQL and Analyze(cpp) actually ran on this head — the build and unit tests did not. It passed on the previous head, and nothing here touches workflow files, so it looks like infrastructure. Worth a re-run before merge so the classic `ConsumerTest` changes are actually exercised in CI.

Rebasing onto main would also pick up #604, which landed after this branch.

lib/ConsumerImpl.cc
lhotari: Confirmed at `ConsumerImpl.cc:344` — cleared in the consumer-created path right after `incomingMessages_.clear()` and before flow permits go out, so the redelivery window parks normally instead of reporting a stale terminal result. Thanks.

lib/ConsumerImpl.cc
lhotari: Both sync paths agree with the async one now — including the post-timeout re-check at line 1301, which is the part that is easy to miss.

On the residual you documented: agreed it is out of scope, and documenting beats papering over it. Worth noting for later that the shape most likely to hit it is the plain drain loop — park in `receive()` on an empty queue, then end-of-topic arrives — since nothing wakes `pop()`. In practice the flag is usually already set by the time the application re-enters `receive()`, so I am not holding the PR for it.

lib/st/QueueConsumerImpl.cc
lhotari: This is the fix I had in mind, and the details are right: the increment happens before the message is offered to the mux queue (so there is no window where a message is in flight but uncounted), and settling happens *after* the ack is enqueued, so a deferred close still flushes it. Handing the future out of `takeDrainedSegmentLocked` to be closed outside the lock is a nice touch.

One non-blocking note for a follow-up: `onMessageSettled` decrements once per ack/nack *call*, not once per message. An application that settles the same id twice — a double ack, or a nack then an ack — while another message on a terminated segment is still outstanding would drive the count to zero early and close the consumer, costing that last ack. The blast radius is a duplicate on the next attach rather than a lost message, and it needs API misuse to reach, so I would not hold the PR for it.

lib/st/QueueConsumerImpl.cc
lhotari: Confirmed — `executor_->postWork` at `QueueConsumerImpl.cc:294-297` breaks both inline-completion paths, so each message unwinds before the next receive is armed. The 1200-message drain test exercises it well past both the prefetch and mux capacities.

lib/st/QueueConsumerImpl.cc
lhotari: Good — and the two details that make it safe are both there: `isSegmentStillWantedLocked` so a retry cannot resurrect a segment that left the DAG or already drained, and a weak self-reference so a pending backoff timer does not pin the consumer. Keeping the first-layout path failing `subscribe()` outright is the right split.

lib/st/QueueConsumerImpl.cc
lhotari: Confirmed — `topic_` is passed as the override when wrapping the message, so `topic()` reports the topic the application actually subscribed to.

lib/st/QueueConsumerImpl.cc
lhotari: Failing the subscribe is the better of the two options I offered — a real error channel beats a log line nobody reads. Documenting the caveat at both API sites is the part that keeps it honest until the follow-up lands.

lib/st/ReceiveQueue.cc
lhotari: Confirmed — live timers now track outstanding receives rather than receive rate times timeout. I like that the timer is only armed after re-checking the entry is still parked, so a receive that completed in the race never starts one at all.

tests/st/StQueueConsumerE2ETest.cc
lhotari: This is exactly the test, and the reattach assertion is what makes it a real regression test rather than a happy-path walk — it fails loudly on the old code instead of silently passing. Sizing the backlog above both the prefetch and mux capacities also gets back-pressure and the executor hop under test for free.

lhotari:
Following up on the CI note in my review — I dug into why `PR validation` never ran on `60c763b`, and it is not something you did.

Two Docker actions in our workflows were pinned to SHAs that are **not on the ASF org-wide allowlist**:

- `docker/setup-buildx-action@4d04d5d9`
- `docker/setup-qemu-action@ce360397`

When a workflow references a non-allowlisted action, the run fails with a bare **"Startup failure"** — [per ASF infra](https://github.com/apache/infrastructure-actions/blob/main/allowlist-check/README.md), *"no logs, no notifications, and the PR may appear green because no checks ran."* That is exactly what happened here: `CodeQL` and `Analyze (cpp)` passed, so the PR looked green, while the build and the unit tests never executed. So the classic `ConsumerTest` changes in this PR are still unverified by CI.

#610 fixes it repo-wide, bumping both to the latest approved revisions from `apache/infrastructure-actions` (and updating the GitHub-owned actions, which the allowlist implicitly trusts, to their latest majors). Its own `PR validation` run is currently `in_progress` rather than failing at startup, which is the confirmation that the allowlist was the blocker.

Once #610 is merged, a rebase here should give this PR a real CI run. My approval stands — I would just want to see the build and tests actually go green on the rebased head before it lands.

PR #609, opened 2026-08-17
[fix][client] Apply connect and request timeouts to the OAuth2 HTTP requests

### LINKED ISSUE #608: [Bug] OAuth2 authentication makes HTTP requests with no timeout, hanging the client indefinitely
### Search before asking

### Version

pulsar-client-cpp v4.2.0 (also verified at master, commit 1d08b2b, 2026-07-21)
Reproduced through pulsar-client (Python) 3.13.0 from PyPI, which pins pulsar-cpp 4.2.0
via dependencies.yaml.
OS: Linux x86_64

### Minimal reproduce step

Requires no Pulsar broker and no real OAuth2 issuer. The script below starts a stub TCP
listener that accepts the connection and then never replies — the behaviour of a
black-holed or overloaded issuer.

  pip install pulsar-client==3.13.0
  python3 repro_oauth_timeout.py 100

Script attached. It configures the client with operation_timeout_seconds=5 and
connection_timeout_ms=5000, points issuer_url at the stub, and calls create_producer().

The OAuth2 flow fails within a bounded, ideally configurable, period — and at minimum
within the client's own configured timeouts.

### What did you see instead?

The call never returns. Observed output:

    [*] pulsar-client 3.13.0
    [*] operation_timeout_seconds=5, connection_timeout_ms=5000
    [*] Calling create_producer(); watchdog set to 100s

        [stub:issuer] accepted TCP connection, now stalling (will never respond)
        still blocked after   15.0s (no client-side timeout has fired)
        still blocked after   30.0s (no client-side timeout has fired)
        still blocked after   45.0s (no client-side timeout has fired)
        still blocked after   60.0s (no client-side timeout has fired)
        still blocked after   75.0s (no client-side timeout has fired)
        still blocked after   90.0s (no client-side timeout has fired)
        still blocked after  105.0s (no client-side timeout has fired)

    [RESULT] REPRODUCED. Still blocked in OAuth2 discovery after 105.0s.

Both configured client timeouts were exceeded by more than 20x with no sign of returning.
Neither operationTimeoutSeconds nor connectionTimeoutMs bounds the auth flow's HTTP calls.

### Anything else?

Root cause: CurlWrapper::Options::timeoutInSeconds defaults to 0 (lib/CurlWrapper.h:55) and
is applied unconditionally at lib/CurlWrapper.h:123:

    curl_easy_setopt(handle_, CURLOPT_TIMEOUT, options.timeoutInSeconds);

CURLOPT_TIMEOUT of 0 means no timeout in libcurl. CURLOPT_CONNECTTIMEOUT is never set
anywhere in the repository, leaving only libcurl's 300s default for the connect phase and
nothing at all once connected.

Both OAuth2 call sites leave the default in place:

  1. Issuer discovery — lib/auth/AuthOauth2.cc:277, in fetchTokenEndpoint(), passes a
     default-constructed options object:
         curl.get(wellKnownUrl, "Accept: application/json", {}, tlsContext);

  2. Token endpoint — lib/auth/AuthOauth2.cc:336-339, in fetchOauth2Token(), constructs
     options but sets only postFields.

This looks specific to the OAuth2 path. Other components set a timeout on the same wrapper:
  - lib/HTTPLookupService.cc:229  -> options.timeoutInSeconds = lookupTimeoutInSeconds_;
  - lib/auth/athenz/ZTSClient.cc:369 -> options.timeoutInSeconds = REQUEST_TIMEOUT;

Impact, as hit in production by a user running OAuth2 in Kubernetes: pods hang rather than
fail, so the problem stays invisible to health checks; and a transient issuer fault cannot
self-heal because recovery never starts.

Suggested fix: set a default timeout at both OAuth2 call sites following the existing
HTTPLookupService/ZTSClient pattern, and set CURLOPT_CONNECTTIMEOUT in CurlWrapper. Ideally
expose the value as an OAuth2 parameter.

[repro_oauth_timeout.py](https://github.com/user-attachments/files/31149391/repro_oauth_timeout.py)

david-streamlio:
Fixes #608

### Motivation

The OAuth2 authentication flows issue two HTTP requests — the OpenID discovery request to
`<issuer_url>/.well-known/openid-configuration` and the token request to the token endpoint —
and neither had any timeout configured.

`CurlWrapper::Options::timeoutInSeconds` defaults to `0`, which libcurl interprets as "no
limit" for `CURLOPT_TIMEOUT`, and `CURLOPT_CONNECTTIMEOUT` was never set at all. Both OAuth2
call sites left the default in place: `fetchTokenEndpoint()` passed `{}` and
`fetchOauth2Token()` only set `postFields`.

As a result, an issuer that accepts the TCP connection but never replies blocks
`Authentication::getAuthData()` — and therefore client creation and every subsequent
reconnection — forever, with no way for the application to recover.

### Modifications

- `CurlWrapper::Options` gains `connectTimeoutInSeconds`, applied as `CURLOPT_CONNECTTIMEOUT`
  when greater than 0. The existing `timeoutInSeconds{0}` default is unchanged, so the other
  callers of `CurlWrapper::get()` keep their current behaviour.
- Both timeouts are now set at the two OAuth2 call sites.
- The values are configurable through the OAuth2 `ParamMap`, alongside `issuer_url`,
  `audience` and `scope`:
  - `connect_timeout_seconds`, default `10`
  - `request_timeout_seconds`, default `30`

  A missing key, or a value that is not a non-negative integer, falls back to the default and
  logs a warning. `0` is accepted and falls back to libcurl's own defaults. Both keys work for
  `client_secret_post` and `tls_client_auth`, and are documented in
  `include/pulsar/Authentication.h`.
- Drive-by: `CURLOPT_TIMEOUT` and `CURLOPT_CONNECTTIMEOUT` are now passed as `long`.
  `curl_easy_setopt` is variadic and reads a `long` for these options, so passing an `int` was
  undefined behaviour.

### Verifying this change

- `AuthPluginTest.testOauth2UnresponsiveIssuer` points an OAuth2 flow at a local socket that
  is bound and listening but never accepted — the kernel completes the TCP handshake from the
  listen backlog, so the client connects and then waits for a response that never arrives. The
  test configures `request_timeout_seconds = 2` and asserts that `getAuthData()` returns
  `ResultAuthenticationError` within a bounded time instead of hanging. It completes in ~2s;
  with the timeout removed the same test hangs indefinitely.
- `AuthPluginTest.testOauth2TimeoutSettings` covers the defaults, explicit values, `0`, and
  invalid values falling back to the defaults.

Ran locally with `--gtest_filter=*Oauth2*`: 11 of 12 pass. The remaining failure,
`testOauth2Failure`, requires a broker on `localhost:6650` and fails identically on `main`.

### Documentation

- [x] `doc-required`
(The two new `connect_timeout_seconds` / `request_timeout_seconds` parameters should be added
to the OAuth2 section of the C++ client docs on the website. The Doxygen comments in
`include/pulsar/Authentication.h` are updated in this PR.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/pulsar/Authentication.h
+++ b/include/pulsar/Authentication.h
@@ -531,6 +531,10 @@ typedef std::shared_ptr<CachedToken> CachedTokenPtr;
  *   "tls_cert_file": "/path/to/cert.pem",
  *   "tls_key_file": "/path/to/key.pem"
  * ```
+ *
+ * Both forms accept the optional "connect_timeout_seconds" and "request_timeout_seconds" keys, see
+ * AuthOauth2::create(ParamMap&).
+ *
  *  If passed in as std::string, it should be in Json format.
  */
 class PULSAR_PUBLIC AuthOauth2 : public Authentication {
@@ -549,6 +553,11 @@ class PULSAR_PUBLIC AuthOauth2 : public Authentication {
      * `issuer_url`, `tls_cert_file`, and `tls_key_file`.
      * Optional keys: `client_id`, `audience`, `scope`. If `client_id` is omitted, the client
      * uses `pulsar-client`.
+     *
+     * Both methods accept the optional keys `connect_timeout_seconds` (default: 10) and
+     * `request_timeout_seconds` (default: 30), which bound the HTTP requests sent to the issuer so
+     * that an unresponsive issuer cannot block the client forever. 0 falls back to the underlying
+     * libcurl defaults: 300 seconds to connect and no limit for the whole request.
      *
      * @param parameters the key-value to create OAuth 2.0 client credentials
      * @see http://pulsar.apache.org/docs/en/security-oauth2/#client-credentials
--- a/lib/CurlWrapper.h
+++ b/lib/CurlWrapper.h
@@ -52,7 +52,10 @@ class CurlWrapper {
         std::string method;
         std::string postFields;
         std::string userAgent;
+        // The maximum time the whole request is allowed to take. 0 means no limit.
         int timeoutInSeconds{0};
+        // The maximum time the connection phase is allowed to take. 0 means libcurl's default (300s).
+        int connectTimeoutInSeconds{0};
         int maxLookupRedirects{-1};
         bool authAllowRedirect{false};
     };
@@ -120,7 +123,10 @@ inline CurlWrapper::Result CurlWrapper::get(const std::string& url, const std::s
     // Without this config, Curl_resolv_timeout might crash in multi-threads environment
     curl_easy_setopt(handle_, CURLOPT_NOSIGNAL, 1L);
 
-    curl_easy_setopt(handle_, CURLOPT_TIMEOUT, options.timeoutInSeconds);
+    curl_easy_setopt(handle_, CURLOPT_TIMEOUT, static_cast<long>(options.timeoutInSeconds));
+    if (options.connectTimeoutInSeconds > 0) {
+        curl_easy_setopt(handle_, CURLOPT_CONNECTTIMEOUT, static_cast<long>(options.connectTimeoutInSeconds));
+    }
     if (!options.userAgent.empty()) {
         curl_easy_setopt(handle_, CURLOPT_USERAGENT, options.userAgent.c_str());
     }
--- a/lib/auth/AuthOauth2.cc
+++ b/lib/auth/AuthOauth2.cc
@@ -51,6 +51,24 @@ OAuth2TokenEndpointAuthMethod parseTokenEndpointAuthMethod(const std::string& au
     return OAuth2TokenEndpointAuthMethod::Unknown;
 }
 
+int parseTimeoutSeconds(const ParamMap& params, const std::string& key, int defaultValue) {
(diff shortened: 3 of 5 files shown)


david-streamlio:
Closing this as a duplicate of #606, which covers the same fix and predates this PR.

Both add `connect_timeout_seconds` / `request_timeout_seconds` to the OAuth2 `ParamMap` with the same 10s / 30s defaults, add a connect-timeout option to `CurlWrapper`, and apply both timeouts at the discovery and token call sites for both OAuth2 flows.

Two small deltas from this PR that might be worth folding into #606:

1. `CURLOPT_TIMEOUT` and `CURLOPT_CONNECTTIMEOUT` should be passed as `long`. `curl_easy_setopt` is variadic and reads a `long` for these options, so passing an `int` is undefined behaviour.
2. Accepting `0` as "fall back to libcurl's own defaults" rather than rejecting it, so users can opt out of the application-level deadline without an out-of-band mechanism.

/cc @RobertIndie — #608 stays open until #606 lands.

PR #611, opened 2026-08-31
Scalable Topics: `pulsar::st` stream consumer — controller-assigned ordered consumption with cumulative acks

merlimat:
### Motivation

Next consumer in the `pulsar::st` (scalable-topics) SDK after the queue consumer (#605): the **stream consumer** — ordered consumption over the segment DAG.

The design headline, from the Java v5 reference (`ScalableStreamConsumer` / `ScalableConsumerClient` / broker `SubscriptionCoordinator`): **the client enforces no DAG ordering itself**. The broker's subscription coordinator withholds an active child segment from the consumer's assignment until *every* parent has drained for this subscription (per-subscription backlog reaching zero; merges fall out of the same rule). The client's job is therefore to register with the controller, subscribe **Exclusive** to exactly the assigned segments, and mux them — per-key order across splits and merges holds as long as the application keeps acknowledging. That makes acking-as-you-go a *liveness* requirement: a consumer that defers all acks to the end keeps the children withheld forever (documented on the API and encoded in the e2e).

### What's in this PR

- **Consumer assignment session** — the controller-registration protocol the queue consumer never needed: `Commands::newScalableTopicSubscribe`, `ClientConnection` registries for pushed `CommandScalableTopicAssignmentUpdate` (by consumer id) and one-shot `CommandScalableTopicSubscribeResponse` callbacks (by request id), both drained on connection close — the same idiom as the DAG-watch session. `ConsumerAssignmentSession` resolves the controller leader through a one-shot DAG-watch lookup (TLS-aware; behind a proxy or before leader election it falls back to the regular lookup path, where any broker forwards), registers + subscribes, applies epoch-gated assignments, replays the current assignment on listener registration, reconnects with backoff after the initial assignment and fails fast before it. `close()` drops only the local registration; the broker reaps it through its grace timer (Java parity).
- **`StreamConsumerImpl`** — one Exclusive classic consumer per assigned segment, receive loops fanning into the shared mux `ReceiveQueue`, and every delivered message stamped with a **position vector**: a snapshot of all segments' latest-delivered ids taken at delivery time, so one cumulative acknowledgment advances every cursor (fanned out as one `acknowledgeCumulativeAsync` per segment). `TopicTerminated` closes the segment immediately and late acks no-op — a deliberate divergence from the queue consumer's deferred close, which doesn't transfer because one cumulative ack settles an unbounded prefix. Newly assigned segments retry only rebalance collisions (`ConsumerBusy` / `ConsumerAssignError`, the previous Exclusive owner not having released yet) with bounded backoff, and fail fast on anything else. `AckPolicy::negativeAckRedeliveryDelay` is deliberately not wired — a stream consumer has no negative-ack path.
- **`receiveMulti`** — `ReceiveQueue::receiveMultiAsync` (greedy drain, then wait with the remaining deadline until full or timeout; a short or empty batch is a normal result), with broker-free unit tests. This also closes G1 from the API review for the stream side.
- **Seam fix** — the `readCompacted` guard only accepted the `persistent://` domain and would have rejected every `segment://` subscribe; relaxed only on the internal `allowSegmentTopic` path.
- **e2e** (runs under the existing st CI phase with no workflow changes) — three tests against a real broker:
  - ordered round-trip on one segment + one cumulative ack of the last message settles the stream (a reattached consumer receives nothing);
  - **the DAG-replay scenario**: produce → split → produce → attach Earliest; the initial assignment carries only the sealed parent, acking-as-we-go drains it, the broker's drain detection pushes the children, and every parent message arrives before any child message;
  - two initial segments drained via `receiveMulti` with no intermediate acks; acknowledging only the final message advances both cursors through its position vector.
- **Harness dedup** — the admin REST helpers the three e2e suites each carried now live in `tests/st/StE2EAdmin.h` (the follow-up flagged in #605).

### Deferred (fails loudly, not silently)

- **PIP-486 bucket-shared assignments** (consumers outnumbering segments): the subscribe fails with `ResultOperationNotSupported` when the initial assignment carries bucket ranges. The whole-segment Exclusive path covers every assignment the controller produces otherwise, and the deferral avoids porting the drain-release machinery blind.
- **Namespace-mode subscriptions**, same as the queue consumer.

### Testing

- `pulsar-st-tests`: **105/105** — 94 broker-free unit tests (incl. new session + `receiveMulti` suites) and 11 e2e (5 producer + 3 queue + 3 stream) against a `5.0.0-M1` standalone.
- clang-format-11 and clang-tidy (the Lint job's checks) clean on all new/changed translation units.

### DIFF
--- a/lib/st/StClientImpl.cc
+++ b/lib/st/StClientImpl.cc
@@ -23,6 +23,7 @@
 
 #include "QueueConsumerImpl.h"
 #include "StProducerImpl.h"
+#include "StreamConsumerImpl.h"
 
 namespace pulsar::st {
 
@@ -62,9 +63,18 @@ Future<detail::ProducerCore> ClientImpl::createProducerAsync(ProducerConfig conf
 // value (the sink the real implementation will move from), but as a stub it does not
 // consume the config yet — hence the value-param suppressions.
 
-// NOLINTNEXTLINE(performance-unnecessary-value-param)
-Future<detail::StreamConsumerCore> ClientImpl::subscribeStreamAsync(StreamConsumerConfig) {
-    return notImplementedYet<detail::StreamConsumerCore>("subscribeStream");
+Future<detail::StreamConsumerCore> ClientImpl::subscribeStreamAsync(StreamConsumerConfig config) {
+    auto impl = std::make_shared<StreamConsumerImpl>(classic_, std::move(config));
+    detail::Promise<detail::StreamConsumerCore> promise;
+    // Keep the impl alive until start() resolves; on success mint the public core over it.
+    impl->start().addListener([impl, promise](const Expected<void>& result) {
+        if (result) {
+            promise.setValue(detail::StreamConsumerCore{impl});
+        } else {
+            promise.setError(result.error());
+        }
+    });
+    return promise.getFuture();
 }
 
 Future<detail::QueueConsumerCore> ClientImpl::subscribeQueueAsync(QueueConsumerConfig config) {
--- a/lib/st/StClientImpl.cc
+++ b/lib/st/StClientImpl.cc
@@ -70,9 +70,12 @@ Future<detail::StreamConsumerCore> ClientImpl::subscribeStreamAsync(StreamConsum
     impl->start().addListener([impl, promise](const Expected<void>& result) {
         if (result) {
             promise.setValue(detail::StreamConsumerCore{impl});
-        } else {
-            promise.setError(result.error());
+            return;
         }
+        // No core will ever close it: release the controller registration and any
+        // segment consumer that did come up before failing the subscribe.
+        const Error& error = result.error();
+        impl->closeAsync().addListener([promise, error](const Expected<void>&) { promise.setError(error); });
     });
     return promise.getFuture();
 }
--- a/lib/st/StreamConsumerImpl.cc
+++ b/lib/st/StreamConsumerImpl.cc
@@ -1,0 +1,428 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "StreamConsumerImpl.h"
+
+#include <algorithm>
+#include <chrono>
+#include <cstdint>
+#include <optional>
+#include <random>
+#include <string>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "MessageIdImpl.h"
+#include "MessageImpl.h"
+#include "lib/LogUtils.h"
+
+DECLARE_LOG_OBJECT()
+
+namespace pulsar::st {
+
+namespace {
+
+pulsar::InitialPosition toClassicInitialPosition(SubscriptionInitialPosition position) {
+    return position == SubscriptionInitialPosition::Earliest ? pulsar::InitialPositionEarliest
+                                                             : pulsar::InitialPositionLatest;
+}
+
+// Close a segment consumer once its creation future resolves (a no-op if creation failed).
+void closeWhenReady(Future<pulsar::Consumer> future) {
+    future.addListener([](const Expected<pulsar::Consumer>& result) {
+        if (result) {
+            pulsar::Consumer consumer = *result;
+            consumer.closeAsync([](pulsar::Result) {});
+        }
+    });
+}
+
+// The consumer name is the controller's registration key, so one is always needed:
+// default to "v5-stream-" + 8 random hex chars when unset (Java parity).
+std::string defaultConsumerName() {
+    static const char kHex[] = "0123456789abcdef";
+    std::random_device rd;
+    std::string suffix(8, '0');
+    for (auto& c : suffix) {
+        c = kHex[rd() % 16];
+    }
+    return "v5-stream-" + suffix;
+}
+
+// Rebalance handoffs collide briefly by design: the previous Exclusive owner has not
+// released the segment yet (ConsumerBusy), or sticky ranges still overlap
+// (ConsumerAssignError). Only these are worth retrying; anything else is a real
+// error that retrying would only hide.
+bool isRebalanceCollision(Result result) {
+    return result == ResultConsumerBusy || result == ResultConsumerAssignError;
+}
+
+}  // namespace
+
+StreamConsumerImpl::StreamConsumerImpl(pulsar::ClientImplPtr classic, StreamConsumerConfig config)
+    : classic_(std::move(classic)),
+      config_(std::move(config)),
+      topic_(config_.topic),
+      subscription_(config_.subscriptionName),
+      consumerName_(config_.consumerName.value_or(defaultConsumerName())),
+      executor_(classic_->getIOExecutorProvider()->get()),
+      receiveQueue_(std::make_shared<ReceiveQueue>(executor_, kReceiveQueueCapacity)) {}
+
+Future<void> StreamConsumerImpl::start() {
+    if (config_.useNamespace) {
+        startPromise_.setError(Error{ResultOperationNotSupported,
+                                     "namespace-mode subscriptions are not implemented yet in the "
+                                     "scalable-topics client"});
+        return startPromise_.getFuture();
+    }
+    session_ = std::make_shared<ConsumerAssignmentSession>(
+        classic_, config_.topic, config_.subscriptionName, consumerName_,
+        pulsar::ScalableConsumerType_STREAM);
+    std::weak_ptr<StreamConsumerImpl> weak = weak_from_this();
+    session_->setListener([weak](const std::vector<AssignedSegment>& newSegments,
+                                 const std::vector<AssignedSegment>& oldSegments) {
+        if (auto self = weak.lock()) self->onAssignmentChange(newSegments, oldSegments);
+    });
+    // The listener drives the success path (the session applies the initial assignment
+    // through it before this future resolves); here only a failure is surfaced.
+    session_->start().addListener([weak](const Expected<std::vector<AssignedSegment>>& result) {
+        if (!result) {
+            if (auto self = weak.lock()) self->startPromise_.setError(result.error());
+        }
+    });
+    return startPromise_.getFuture();
+}
+
+void StreamConsumerImpl::onAssignmentChange(const std::vector<AssignedSegment>& newSegments,
+                                            const std::vector<AssignedSegment>& /*oldSegments*/) {
+    std::vector<Future<pulsar::Consumer>> retired;
+    std::vector<AssignedSegment> toAdd;
+    std::vector<std::uint64_t> bucketShared;
+    bool first = false;
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
+        first = !sawFirstAssignment_;
+        sawFirstAssignment_ = true;
+        currentAssignment_ = newSegments;
+
+        std::unordered_set<std::uint64_t> targetIds;
+        for (const auto& assigned : newSegments) targetIds.insert(assigned.segment.segmentId);
+        for (auto it = segmentConsumers_.begin(); it != segmentConsumers_.end();) {
+            if (targetIds.find(it->first) == targetIds.end()) {
+                // Released by a rebalance: close immediately. Unacked in-flight messages
+                // are redelivered to the segment's next owner from the shared cursor.
+                retired.push_back(std::move(it->second));
+                latestDelivered_.erase(static_cast<std::int64_t>(it->first));
+                it = segmentConsumers_.erase(it);
+            } else {
+                ++it;
+            }
+        }
+        for (const auto& assigned : newSegments) {
+            if (segmentConsumers_.find(assigned.segment.segmentId) != segmentConsumers_.end()) {
+                continue;
+            }
+            if (!assigned.ownedBucketRanges.empty()) {
+                bucketShared.push_back(assigned.segment.segmentId);
+            } else {
+                toAdd.push_back(assigned);
+            }
+        }
+    }
+
+    for (auto& future : retired) closeWhenReady(future);
+
+    if (!bucketShared.empty()) {
+        // PIP-486 bucket-sharing (consumers outnumbering segments) is not supported yet.
+        if (first) {
+            startPromise_.setError(
+                Error{ResultOperationNotSupported,
+                      "bucket-shared segment assignments (PIP-486) are not supported yet in the "
+                      "scalable-topics client; use at most one stream consumer per segment"});
+            return;
+        }
+        for (auto segmentId : bucketShared) {
+            LOG_ERROR("[" << topic_ << "] segment " << segmentId
+                          << " was assigned bucket-shared (PIP-486), which is not supported yet; "
+                             "skipping it");
+        }
+    }
+
+    if (first) {
+        if (toAdd.empty()) {
+            startPromise_.setSuccess();
+            return;
+        }
+        auto remaining = std::make_shared<std::atomic<int>>(static_cast<int>(toAdd.size()));
+        for (const auto& assigned : toAdd) {
+            getOrCreateSegmentConsumerAsync(assigned).addListener(
+                [self = shared_from_this(), remaining](const Expected<pulsar::Consumer>& result) {
+                    if (!result) {
+                        self->startPromise_.setError(result.error());  // first error wins (idempotent)
+                        return;
+                    }
+                    if (remaining->fetch_sub(1) == 1) self->startPromise_.setSuccess();
+                });
+        }
+    } else {
+        for (const auto& assigned : toAdd) subscribeSegmentWithRetry(assigned, /*attempt*/ 0);
+    }
+}
+
+pulsar::ConsumerConfiguration StreamConsumerImpl::buildSegmentConfiguration(
+    const AssignedSegment& assigned) const {
+    // Build a FRESH config every time (pulsar::ConsumerConfiguration's copy ctor shares its impl).
+    pulsar::ConsumerConfiguration conf;
+    conf.setConsumerType(pulsar::ConsumerExclusive);
+    conf.setSchema(config_.schema);
+    conf.setSubscriptionInitialPosition(toClassicInitialPosition(config_.initialPosition));
+    conf.setConsumerName(consumerName_ + "-seg-" + std::to_string(assigned.segment.segmentId));
+    if (config_.ackPolicy.groupTime) {
+        conf.setAckGroupingTimeMs(static_cast<long>(config_.ackPolicy.groupTime->count()));
+    }
+    // AckPolicy::negativeAckRedeliveryDelay is deliberately not wired: a stream
+    // consumer has no negative-ack path (documented on the config field).
+    if (config_.readCompacted) {
+        conf.setReadCompacted(*config_.readCompacted);
+    }
+    if (config_.replicateSubscriptionState) {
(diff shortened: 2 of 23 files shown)


lib/st/StreamConsumerImpl.cc
```diff
@@ -0,0 +1,427 @@
...
+        startPromise_.setError(Error{ResultOperationNotSupported,
+                                     "namespace-mode subscriptions are not implemented yet in the "
+                                     "scalable-topics client"});
+        return startPromise_.getFuture();
+    }
+    session_ =
+        std::make_shared<ConsumerAssignmentSession>(classic_, config_.topic, config_.subscriptionName,
+                                                    consumerName_, pulsar::ScalableConsumerType_STREAM);
+    std::weak_ptr<StreamConsumerImpl> weak = weak_from_this();
+    session_->setListener([weak](const std::vector<AssignedSegment>& newSegments,
```
lhotari: **[BUG] the listener is registered before `session_->start()`, so the empty replay completes `startPromise_` with success and every start failure is swallowed**

`ConsumerAssignmentSession::setListener` replays the current assignment with no emptiness guard, so registering the listener here — before `session_->start()` at :105 — fires it immediately with `({}, {})`:

https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L301-L316

[`tests/st/ConsumerAssignmentSessionTest.cc:58-77`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/tests/st/ConsumerAssignmentSessionTest.cc#L58-L77) pins exactly that ("Before any assignment the replay still fires"), so this is the session behaving as designed — it is the ordering on the consumer side that is wrong.

That replay lands in `onAssignmentChange`, which takes `first = !sawFirstAssignment_` at [`StreamConsumerImpl.cc:121`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L121), sets the flag, and — with an empty `toAdd` — calls `startPromise_.setSuccess()` at [`StreamConsumerImpl.cc:170`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L170). `detail::SharedState::complete` is first-writer-wins ([`include/pulsar/st/detail/SharedState.h:42-46`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/include/pulsar/st/detail/SharedState.h#L42-L46)), so every later `setError` on that promise returns `false` and is dropped.

**Failure scenario 1 — a subscribe that cannot fail.** Point a `StreamConsumer` at a broker that does not advertise scalable-topics support. `subscribeOn` fails the attempt with `ResultUnsupportedVersionError` ([`ConsumerAssignmentSession.cc:146-149`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L146-L149)), which reaches [`StreamConsumerImpl.cc:107`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L107) — where `startPromise_.setError` is a no-op because the promise is already successful. `ClientImpl::subscribeStreamAsync` ([`lib/st/StClientImpl.cc:70-77`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StClientImpl.cc#L70-L77)) therefore hands the application a `StreamConsumerCore`, and an untimed `receiveAsync()` on it never completes, because nothing ever closes the receive queue. The same holds for `ResultTopicNotFound` and for a connect failure.

**Failure scenario 2 — the PIP-486 deferral never fires.** The real initial assignment always arrives with `first == false`, so the branch at [`StreamConsumerImpl.cc:152-160`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L152-L160) that fails the subscribe with `ResultOperationNotSupported` for bucket-shared segments is unreachable; control falls to the `LOG_ERROR`-and-skip at [`:161-165`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L161-L165) and the consumer silently consumes nothing. The description lists this under *Deferred (fails loudly, not silently)*.

**Failure scenario 3.** For the same reason the initial-subscribe error propagation at [`:173-183`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L173-L183) is dead: the real initial assignment goes down the `subscribeSegmentWithRetry` branch at [`:185`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L185), which only logs a non-collision failure.

The Java reference orders this the other way round and guards the replay. `ScalableStreamConsumer.createAsyncImpl` calls `subscribeInitialWithRetry(initialAssignment)` on the assignment returned by registration and only then `session.setListener(consumer)`; `ScalableConsumerClient.setListener` replays under `if (current != null)`. Passing the initial assignment in explicitly and registering the listener after it has been applied would restore both, and would also make `sawFirstAssignment_` unnecessary.

No test currently covers this: all three e2e cases are happy-path and skip without `PULSAR_ST_E2E=1`, and there is no `StreamConsumerImpl` unit test. A broker-free case asserting that `start()` fails when the session never registers would pin it.

lib/st/StreamConsumerImpl.cc
```diff
@@ -0,0 +1,427 @@
...
+}
+
+void StreamConsumerImpl::onAssignmentChange(const std::vector<AssignedSegment>& newSegments,
+                                            const std::vector<AssignedSegment>& /*oldSegments*/) {
+    std::vector<Future<pulsar::Consumer>> retired;
+    std::vector<AssignedSegment> toAdd;
+    std::vector<std::uint64_t> bucketShared;
+    bool first = false;
+    {
+        std::lock_guard<std::mutex> lock(mutex_);
```
lhotari, line 163: **[BUG] `onAssignmentChange` never checks `closed_`, so an assignment applied after `closeAsync()` creates segment consumers that nothing ever closes**

`closeAsync` ([`StreamConsumerImpl.cc:389-406`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L389-L406)) sets `closed_`, closes the session and the queue, then takes a **one-time** snapshot of `segmentConsumers_` under this same mutex and closes what it found. `onAssignmentChange` does not check `closed_` — unlike `startReceiveLoop` at :298 and the retry timer at :291 — so an assignment that wins the race afterwards repopulates the map through `getOrCreateSegmentConsumerAsync` and creates fresh Exclusive consumers that are in no close set. A second `closeAsync()` short-circuits on `closed_.exchange(true)`, so it cannot collect them either.

Two ways in. First, a pushed `CommandScalableTopicAssignmentUpdate`: `handleSessionEvent` checks `closed_` at [`lib/st/ConsumerAssignmentSession.cc:198`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L198) and then calls `handleAssignmentReceived`, which does not — an ordinary check-then-act window, and `close()` is called from an application thread while these callbacks run on an IO thread, with no strand serialising the two.

Second, a subscribe response landing after `close()`. The initial attempt's listener has no `closed_` check at all, while `reconnect()`'s listener at [`lib/st/ConsumerAssignmentSession.cc:277`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L277) has one:

https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L82-L98

`close()` ([`lib/st/ConsumerAssignmentSession.cc:318-336`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L318-L336)) drops the consumer-session registration but never the pending subscribe request, so that callback is still live in `ClientConnection::pendingScalableSubscribeRequests_` and fires normally.

**Failure scenario.** `close()` runs while a `CommandScalableTopicAssignmentUpdate` is being dispatched; the update applies, `getOrCreateSegmentConsumerAsync` subscribes Exclusive to `segment://…`, and `startReceiveLoop` returns immediately on `closed_` — so nothing consumes, but the broker-side consumer stays attached for as long as the closed impl is alive. The next owner of that segment gets `ConsumerBusy` from `subscribeSegmentWithRetry` and, after ten attempts, gives up.

Re-checking `closed_` at entry to `onAssignmentChange`, and again when a segment subscribe completes (closing the consumer if the impl closed meanwhile), would cover both paths; adding the `closed_` guard to the initial attempt listener would make it symmetric with `reconnect()`.

lib/st/StreamConsumerImpl.cc
```diff
@@ -0,0 +1,427 @@
...
+                // move ctor), so this is a shared-impl copy, not a deep copy.
+                pulsar::Consumer c = *consumer;
+                self->startReceiveLoop(c, segmentId);
+                promise.setValue(c);
+            } else {
+                // Evict the failed subscribe so a later attempt (retry or assignment
+                // push) can re-create it.
+                {
+                    std::lock_guard<std::mutex> lock(self->mutex_);
+                    self->segmentConsumers_.erase(segmentId);
```
lhotari: **[BUG] a stale segment-subscribe completion erases `segmentConsumers_[segmentId]` by id, so after a remove-then-re-add rebalance the live entry is dropped and that segment is never acked or closed**

This erase — and the `TopicTerminated` one at [`:310`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L310) — removes by segment id without checking that the entry still belongs to the attempt that is failing. There is no per-segment generation token, so a completion from a retired attempt mutates the state of the current one.

**Failure scenario.** Segment X is assigned; `getOrCreateSegmentConsumerAsync` inserts future A at [`:225`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L225) and the subscribe goes in flight. A rebalance drops X: `onAssignmentChange` moves A out to `retired` and erases the entry ([`:127-137`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L127-L137)). A later rebalance re-adds X: no entry exists, so `subscribeSegmentWithRetry` inserts future B and subscribes again. Attempt A now fails (very likely `ConsumerBusy`, since B is racing it for the same Exclusive slot) and its callback erases X — taking B's entry with it. B then succeeds: `startReceiveLoop` runs and its messages reach the application, but `segmentConsumers_` has no entry for X, so `acknowledgeCumulative` hits the `if (!future) continue` at :372 and X's cursor never advances, and `closeAsync` ([`:398-403`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/StreamConsumerImpl.cc#L398-L403)) never closes it.

That second consequence matters more than a leaked consumer here: per this PR's own design the broker withholds a child segment until every parent has drained for the subscription, so a segment whose cursor can no longer advance stalls the DAG for that subscription rather than just leaking a handle.

The Java reference fences this with `segmentReceiveEpoch` — bumped when a release drain starts, captured at receive-loop arm time, and compared at `ScalableStreamConsumer.java:773-775` so a completion from a superseded generation is discarded. The C++ port has no equivalent. Even without porting the whole epoch machinery, erasing only when the mapped future is still the one this attempt inserted would close the window at both call sites.

The precondition is a rebalance that retires and re-adds the same segment while a subscribe is in flight, so this is narrower than the other two — but it is silent when it happens, which is what makes it worth a guard.

lib/st/StClientImpl.cc
```diff
@@ -62,9 +63,18 @@ Future<detail::ProducerCore> ClientImpl::createProducerAsync(ProducerConfig conf
...
-    return notImplementedYet<detail::StreamConsumerCore>("subscribeStream");
+Future<detail::StreamConsumerCore> ClientImpl::subscribeStreamAsync(StreamConsumerConfig config) {
+    auto impl = std::make_shared<StreamConsumerImpl>(classic_, std::move(config));
+    detail::Promise<detail::StreamConsumerCore> promise;
+    // Keep the impl alive until start() resolves; on success mint the public core over it.
+    impl->start().addListener([impl, promise](const Expected<void>& result) {
+        if (result) {
+            promise.setValue(detail::StreamConsumerCore{impl});
+        } else {
+            promise.setError(result.error());
```
lhotari: **[QUALITY] nothing closes the session when `start()` fails, so the controller registration is left on the pooled connection; `removeScalableSubscribeRequest` is added and never called**

On this failure path the impl is simply dropped. Neither `StreamConsumerImpl` nor `ConsumerAssignmentSession` has a destructor, and `ConsumerAssignmentSession::close()` at [`lib/st/ConsumerAssignmentSession.cc:334`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L334) is the only thing that ever calls `removeScalableConsumerSession` — so the registration installed at [`lib/st/ConsumerAssignmentSession.cc:155`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L155) stays in the pooled connection's `scalableConsumerSessions_` until that connection closes.

It is registered before the subscribe is sent, and none of the response paths unwind it: a broker error ([`lib/st/ConsumerAssignmentSession.cc:173`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L173)), a response carrying neither assignment nor error ([`lib/st/ConsumerAssignmentSession.cc:179`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L179)), or `addScalableSubscribeRequest` returning false ([`lib/st/ConsumerAssignmentSession.cc:186`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/st/ConsumerAssignmentSession.cc#L186)) all fail the promise and return. A `TopicNotFound` response is the concrete case — it arrives after registration, on a connection that stays open and shared with every other producer and consumer in the client. Each such attempt leaves one more dead entry behind.

Today the first finding masks this (the failure is swallowed, the application gets a consumer, and closing it cleans up); once `start()` can actually fail, this becomes the normal path. The Java reference handles it explicitly — `ScalableStreamConsumer.createAsyncImpl` ends with `.exceptionallyCompose(ex -> consumer.closeAsync()…)` so a failed start closes the consumer, which closes the session.

Related: `ClientConnection::removeScalableSubscribeRequest` ([`lib/ClientConnection.cc:1478`](https://github.com/apache/pulsar-client-cpp/blob/5e9dfdf19fd8b5f4c45ec571e6fbf9162747b452/lib/ClientConnection.cc#L1478)) is added by this PR and called from nowhere in the tree — I think it is the cleanup that was meant to run on these paths.

lhotari (commented):
Nice piece of work — the position-vector design, the `TopicTerminated` divergence from the queue consumer's deferred close, and the `readCompacted`/`segment://` seam fix all read as faithful ports of the Java v5 reference, and I could follow the reasoning from the code alone. `ReceiveQueue::receiveMultiAsync` and its four broker-free tests look right to me.

The one thing I would like fixed before merge is the start-up ordering. `StreamConsumerImpl::start()` registers the assignment listener *before* calling `session_->start()`, and `ConsumerAssignmentSession::setListener()` replays the current assignment unconditionally — including when there is no assignment yet. So the listener fires once with `({}, {})`, that empty replay consumes `sawFirstAssignment_`, and `startPromise_` is completed with success before the controller registration has even been attempted. Because `detail::SharedState::complete` is first-writer-wins, every later `setError` is a silent no-op: `subscribeStreamAsync` cannot fail, and the PIP-486 `ResultOperationNotSupported` deferral the description advertises as "fails loudly, not silently" is unreachable. The Java reference does the opposite — `ScalableStreamConsumer.createAsyncImpl` subscribes the initial assignment first and calls `session.setListener(consumer)` only after it resolves, and `ScalableConsumerClient.setListener` guards the replay with `if (current != null)`.

The other three are smaller and share a theme: an async completion that arrives late is not fenced against the state it is landing in — neither against `close()` nor against a rebalance that has since retired and re-added the same segment.

On the tests: the three e2e cases are all happy-path (`ASSERT_TRUE(consumerResult)`) and skip without `PULSAR_ST_E2E=1`, and `ConsumerAssignmentSessionTest` covers close-before-start and the replay contract but nothing about `StreamConsumerImpl`. Nothing currently pins the failure contract of `subscribeStreamAsync`, which is why the first point can pass CI. A broker-free test that drives `StreamConsumerImpl::start()` against a session that never registers, asserting the returned future fails, would pin it.

lib/st/StreamConsumerImpl.cc
merlimat: Good catch — confirmed. The listener was registered before `session_->start()`, and `setListener` replayed unconditionally, so the `({}, {})` replay completed `startPromise_` successfully before the controller subscribe had even run; every later `setError` (initial-subscribe failure, PIP-486 gate) was a no-op, and the real initial assignment then arrived with `first == false`.

Fixed in a15d0ed by restructuring to the Java order: `start()` now waits for the session's own `start()` result, subscribes the initial assignment from it (`applyInitialAssignment`, which owns the PIP-486 gate and the counted subscribes that complete `startPromise_`), and only then registers the change listener. `setListener` no longer replays before the first assignment (`currentEpoch_ >= 0`), and `sawFirstAssignment_` is gone. A new broker-free test (`StStreamConsumerTest.testSubscribeFailsWhenControllerIsUnreachable`) pins the regression — it handed back a "live" consumer before the fix.

lib/st/StreamConsumerImpl.cc
merlimat: Confirmed and fixed in a15d0ed: `onAssignmentChange` now returns early under `mutex_` once `closed_` is set, and `getOrCreateSegmentConsumerAsync` refuses (`AlreadyClosed`) under the same lock — so an entry is either created before `closeAsync()`'s snapshot (and closed by it) or never created. `applyInitialAssignment` has the same fence, and the session's `close()` now also withdraws an in-flight subscribe request (details in the `StClientImpl` thread).

lib/st/StreamConsumerImpl.cc
merlimat: Right — a retire-then-re-add of the same segment between a subscribe's start and its failed completion (or a late `TopicTerminated` from the previous consumer) would have evicted the live entry and stalled the cursor. Fixed in a15d0ed with a per-entry generation: `segmentConsumers_` now maps to `SegmentEntry{future, generation}`, `startReceiveLoop` carries the generation, and both eviction paths go through `dropSegmentLocked(segmentId, generation)`, which only removes the entry when the generation matches.

lib/st/StClientImpl.cc
merlimat: Fixed in a15d0ed: on a failed `start()`, `subscribeStreamAsync` now runs `impl->closeAsync()` (session close plus any segment consumer that did come up) and fails the subscribe once that completes. `ConsumerAssignmentSession::close()` also tracks and removes the pending subscribe request (`removeScalableSubscribeRequest` was indeed never called) and fails a still-pending `start()` with `AlreadyClosed`, so nothing hangs on the withdrawn request.

The producer/queue `create*` / `subscribe*` failure paths have the same drop-without-close shape; I'll handle those in a follow-up rather than widen this PR.

